PR #9, opened 2016-11-17
Remove watch_* operations in favor of "list_*(watch=True)"

### LINKED ISSUE #8: Consider removing watch_* operations
watch_* operations are deprecated in kubernetes in favor of "watch" query flag. Consider removing them from the swagger.json in the preprocessing step.

mbohlool:
Kubernetes supports two type of watch operations. One is the old deprecated one (watch_resource...) and on is the new one (list_resource...(watch=True)). k8sutil.Watch class support only the new one. It is not hard to support old deprecated ones too, but they are redundant and removing them make client less crowded. This PR removes old deprecated watch operations after it make sure there exists a new watch operation for them.

fixes #8


scripts/preprocess_spec.py
```diff
@@ -16,24 +16,63 @@ def _to_camel_case(s):
...
             for t in operation['tags']:
                 operation_id = operation_id.replace(_to_camel_case(t), '')
             operation['operationId'] = operation_id
 
-        iterate_through_operations(spec, strip_tags_from_operation_id)
+        iterate_through_operations(spec, strip_tags_from_operation_id,
+                                   make_operation_id_map)
+
+        def remove_watch_operations(op, parent):
+            id = op['operationId']
```
marun: id is a builtin, so this variable assignment is effectively shadowing it.  Consider differentiating via a trailing underscore (e.g. ``id_``) or using a more verbose name (e.g. ``op_id``).

scripts/preprocess_spec.py
```diff
@@ -16,24 +16,63 @@ def _to_camel_case(s):
     return ''.join(_title(y) for y in s.split("_"))
 
 
-def iterate_through_operations(spec, func):
+def iterate_through_operations(spec, *funcs):
```
marun: (No action required) 

'iterate' usually implies an iterator, which this is not.  Consider renaming to something like 'apply_funcs_to_spec'.

scripts/preprocess_spec.py
```diff
@@ -16,24 +16,63 @@ def _to_camel_case(s):
...
     for k, v in spec['paths'].iteritems():
         for op in _ops:
             if op in v:
```
marun: (No action required) 

Consider avoiding unnecessary indentation:

```python
for op in _ops:
  if not op in v:
    continue
  ...

File: scripts/preprocess_spec.py [hunk 4]
```diff
@@ -16,24 +16,63 @@ def _to_camel_case(s):
...
             operation_id = operation['operationId']
             for t in operation['tags']:
                 operation_id = operation_id.replace(_to_camel_case(t), '')
             operation['operationId'] = operation_id
 
-        iterate_through_operations(spec, strip_tags_from_operation_id)
+        iterate_through_operations(spec, strip_tags_from_operation_id,
+                                   make_operation_id_map)
+
+        def remove_watch_operations(op, parent):
```
marun: (No action required)

Consider avoiding the use of complex inline functions (longer than a few lines).  They are effectively impossible to unit test because they can't be referenced outside of their scope.

Speaking of testing, consider adding tests for this script.

scripts/preprocess_spec.py
```diff
@@ -16,24 +16,63 @@ def _to_camel_case(s):
...
+            id = op['operationId']
+            if not id.startswith("watch"):
```
marun: (No action required)

Consider defining ``watch`` as a constant.

scripts/preprocess_spec.py
```diff
@@ -16,24 +16,63 @@ def _to_camel_case(s):
...
+                for func in funcs:
+                    if func(v[op], v):
+                        del v[op]
+
+
+def _has_property(prop_list, property):
+    for prop in prop_list:
+        if prop["name"] == property:
+            return True
+    return False
```
marun: (No action required)

Consider being consistent in returning True/False or True/None (as with ``remove_watch_operations``).

mbohlool:
@marun Thanks for the review. I've applied your comments and also restructured the code to make it more readable. Please take another look.

scripts/preprocess_spec.py
```diff
@@ -30,24 +39,67 @@ def _to_camel_case(s):
     return ''.join(_title(y) for y in s.split("_"))
 
 
-def iterate_through_operations(spec, func):
+def apply_funcs_to_spec(spec, func, *params):
```
marun: nit: Given that this function only applies a single func, consider renaming to ``appy_func_to_spec``

scripts/preprocess_spec.py
```diff
@@ -30,24 +39,67 @@ def _to_camel_case(s):
...
+    """Apply fun to each operation in the spec.
```
marun: nit: fun -> func

scripts/preprocess_spec.py
```diff
@@ -30,24 +39,67 @@ def _to_camel_case(s):
...
+
+    First two parameters of func will be operation and it's parent, and the
```
marun: (No action required)

Consider defining the signature rather than explaining it (e.g. ``func(operation, parent)``).

scripts/preprocess_spec.py
```diff
@@ -30,24 +39,67 @@ def _to_camel_case(s):
...
-                func(v[op])
+            if op not in v:
+                continue
+            if func(v[op], v, *params):
+                del v[op]
+
+
+def _has_property(prop_list, property_name):
+    for prop in prop_list:
+        if prop["name"] == property_name:
```
marun: (No action required)

Consider using the boolean evaluation directly:

```python
return prop["name"] == property_name
```

scripts/preprocess_spec.py
```diff
@@ -30,24 +39,67 @@ def _to_camel_case(s):
...
+            return True
+
+
+def remove_watch_operations(op, parent, operation_ids):
```
marun: (No action required)

If/when this method is targeted by a unit test, it may make sense to return messages (e.g. ``Cannot find...``) as exceptions and allow the calling function more flexibility in how those cases are handled (i.e. script would output to stdout, test would check that the expected exception was returned for a given input).

scripts/preprocess_spec.py
```diff
@@ -30,24 +39,67 @@ def _to_camel_case(s):
...
+    if not _has_property(params, WATCH_QUERY_PARAM_NAME):
+        print "%s has no watch query param" % list_id
+        return
+    return True
+
+
+def strip_tags_from_operation_id(operation, _):
+    operation_id = operation['operationId']
+    for t in operation['tags']:
+        operation_id = operation_id.replace(_to_camel_case(t), '')
```
marun: (No action required)

``to_camel_case`` is only referenced here, consider inlining.

scripts/preprocess_spec.py
```diff
@@ -30,24 +39,67 @@ def _to_camel_case(s):
...
-        def strip_tags_from_operation_id(operation):
-            operation_id = operation['operationId']
-            for t in operation['tags']:
-                operation_id = operation_id.replace(_to_camel_case(t), '')
-            operation['operationId'] = operation_id
+        apply_funcs_to_spec(spec, strip_tags_from_operation_id)
 
-        iterate_through_operations(spec, strip_tags_from_operation_id)
+        operation_ids = {}
+        apply_funcs_to_spec(spec, make_operation_id_map, operation_ids)
```
marun: (No action required)

``make_operation_id_map`` is only referenced here, consider inlining.  Given that this is a one-liner, it may be appropriate to use lambda.

marun:
Is it necessary to include the generated commit in this PR (i.e. does the second commit fix things breakage in the generated code)?  If not, consider breaking it the generated commit out into a separate PR that just gets merged.

mbohlool:
In k8s, we always include the generated code commit in the same PR that changes the generating code. That will make sure the state of master after submitting each PR is valid (verify-* scripts will fail if we don't do that and we can't actually submit the PR). Here we do not have any verify script yet, but I still like to keep the master always in sync in regard to generating and generated code. I understand this is a suggestion not required. Can you elaborate more on how this is better? Thanks.

scripts/preprocess_spec.py
mbohlool: I missed that. Done.

scripts/preprocess_spec.py
mbohlool: Done

marun:
Ok, makes sense.  The code change results in a generated code change, I wasn't clear on that.

scripts/preprocess_spec.py
marun: Yes, use the syntax in the comment.

scripts/preprocess_spec.py
mbohlool: done.

scripts/preprocess_spec.py
mbohlool: Would like to keep this one :)

scripts/preprocess_spec.py
mbohlool: done.

mbohlool:
@marun I've made sure that changes to client here is related only to the changes in codegen in this PR by updating master generated code in #27

scripts/preprocess_spec.py
mbohlool: Done.

mbohlool:
@marun I think I applied all of your comments, please take another look and LGTM the PR if you are satisfied with the changes. Thanks.

marun:
LGTM

PR #51, opened 2016-11-30
Make the README example code more portable

SEJeff:
Also, the `$HOME` variable can be modified while `os.path.expanduser()` does the right thing.

This works on windows as well whereas the previous code did not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ import os
 from kubernetes import client, config
 
 # Configs can be set in Configuration class directly or using helper utility
-config.load_kube_config(os.environ["HOME"] + '/.kube/config')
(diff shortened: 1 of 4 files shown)


SEJeff:
I signed it!

SEJeff:
If this seems sensible, I can change the actual code as well.

mbohlool:
It does make sense. Please change example codes too and let me know when it is ready for review. Thanks for your contribution.

SEJeff:
@mbohlool all you good sir. I figured I could start small contributing to some of the k8s ecosystem projects. Maybe I'll be able to help you grow this project to graduate out of the incubator.

sebgoa:
lgtm

PR #54, opened 2016-12-01
Adding codecov.io report and badge

mbohlool:
Adding code coverage report and badge only for Non-Generated packages.


SEJeff:
It might sound weird, but coverage of your tests is good as well. In large testing codebases, you'll find dead code *in* your tests. Coverage is how you find it and make it die!~

mbohlool:
Good idea, added.

PR #64, opened 2016-12-14
Add default kube config location

mbohlool:
`kubenetes.config.load_config` needs the location of config file to load it. We can set the location to the default location of ~/.kube/config.

### DIFF
--- a/kubernetes/config/kube_config.py
+++ b/kubernetes/config/kube_config.py
@@ -24,6 +24,7 @@ from oauth2client.client import GoogleCredentials
 
 from .config_exception import ConfigException
 
+KUBE_CONFIG_DEFAULT_LOCATION = '~/.kube/config'
 _temp_files = {}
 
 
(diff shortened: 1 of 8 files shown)


kubernetes/config/kube_config.py
```diff
@@ -244,13 +245,17 @@ def get_with_name(self, name):
             'Expected object with name %s in %s list' % (name, self.name))
 
 
-def list_kube_config_contexts(config_file):
+def list_kube_config_contexts(config_file=None):
+
+    if not config_file:
```
mml: I would use `if config_file is not None` here.

kubernetes/config/kube_config.py
```diff
@@ -259,6 +264,9 @@ def load_kube_config(config_file, context=None):
     from config file will be used.
     """
 
+    if not config_file:
```
mml: I would use `if config_file is not None` here.

kubernetes/config/kube_config.py
mbohlool: you mean `if config_file is None`, right?

kubernetes/config/kube_config.py
mbohlool: ditto

kubernetes/config/kube_config.py
mbohlool: You mean `if config_file is None` right?

PR #69, opened 2016-12-15
Bugfix: Python client does not resolve relative paths in kubeconfig

### LINKED ISSUE #68: Python client does not resolve relative paths in kubeconfig
When using relative paths in `~/.kube/config` like this:
```yaml
- cluster:
    certificate-authority: certs/cfm.ca.crt
```

The following [error message](https://gist.github.com/webwurst/5b3aece21c45a634eb0db6e14b2aa3c5#file-z-error-txt) is thrown:
```
kubernetes.client.rest.ApiException: (0)
Reason: SSLError
[Errno 2] No such file or directory
```

Additionally it would be nice when in case of a `No such file or directory` error, the tried path would be shown. 😃

mbohlool:
Resolving relative paths based on cube config file path.

Fixes #68

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,11 @@
-# current-version
+# v1.0.0-beta1
 
-- Add context switch to kube config loader
-- Test coverage improvments
+- Add context switch to kube config loader #46 
+- Add default kube config location #64
+- Bugfix: Python client does not resolve relative paths in kubeconfig #68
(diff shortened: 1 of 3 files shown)


mml:
Replace 'cube' with 'kube' in the commit messages.  Otherwise, LG.

mbohlool:
Done. Thanks.

PR #71, opened 2016-12-16
Support connecting to multiple clusters

### LINKED ISSUE #7: generated Configuration() singleton prevents client from connecting to multiple clusters
Should follow up with swagger-codegen developers on the reasoning around this. We should fix the generator to have one configuration class per client.

mbohlool:
fixes #7


kubernetes/client/configuration.py
```diff
@@ -235,3 +224,11 @@ def to_debug_report(self):
                "Version of the API: v1.5.0-snapshot\n"\
                "SDK Package Version: 1.0.0-snapshot".\
                format(env=sys.platform, pyversion=sys.version)
+
+
+configuration = ConfigurationObject()
+
+
+def Configuration():
```
caesarxuchao: Who is the caller of this function? If a client is created via `new_client_from_config`, I think `configuration` will contain default values rather than values from a real kube config file, is that the callers of `Configuration()` expect?

examples/multiple_clusters.py
```diff
@@ -0,0 +1,51 @@
...
+
+def main():
+
+    contexts, active_context = config.list_kube_config_contexts()
+    if not contexts:
+        print("Cannot find any context in kube-config file.")
+        return
+    contexts = [context['name'] for context in contexts]
+    active_index = contexts.index(active_context['name'])
+    cluster1, first_index = pick(contexts, title="Pick the first context",
```
mbohlool: Yes, it lets user pick two clusters and then show list of pods in them. I am going to extend this example in future, maybe letting user copy a resource from one cluster to the other.

caesarxuchao:
@mbohlool and I discussed offline and it LGTM. Thanks.

PR #88, opened 2017-01-10
Run Tests with python 3.5

dims:
Update tox.ini and .travis.yml to support python 3.5


dims:
@mbohlool are you ok with adding python 3.5 support? we need to support this for pulling this library into openstack projects and CI. please let me know.

mbohlool:
@dims Sure. look like it is passing for 3.5. Thanks.

PR #89, opened 2017-01-10
Extract DevelopmentStatus as a constant to scripts/constants.py

mbohlool:
Having all release related constant in one place make the release process easier and much less error-prone. 

cc: @dims


dims:
LGTM 👍

PR #94, opened 2017-01-11
Add functional test

dims:
Port a bunch of tests from python-k8sclient repository. There is a script that starts a real instance of kubernetes and the tests are executed against it.

Note, if you just have k8s accessible in your localhost:8080, the tests still run. so you don't really need to run "py27-functional" which runs the kube-init.sh

Also, 2 API calls fail currently, i've added a TODO so we can dig into them and fix them

### DIFF
--- a/kubernetes/e2e_test/test_k8sclient.py
+++ b/kubernetes/e2e_test/test_k8sclient.py
@@ -1,0 +1,204 @@
+# -*- coding: utf-8 -*-
+
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+"""
+test_k8sclient
+----------------------------------
+
+Tests for `k8sclient` module. Deploy Kubernetes using:
+http://kubernetes.io/docs/getting-started-guides/docker/
+
+and then run this test
+"""
+
+import unittest
+import urllib3
+import uuid
+
+from kubernetes.client import api_client
+from kubernetes.client.apis import core_v1_api
+
+
+def _is_k8s_running():
+    try:
+        urllib3.PoolManager().request('GET', '127.0.0.1:8080')
+        return True
+    except urllib3.exceptions.HTTPError:
+        return False
+
+
+class TestK8sclient(unittest.TestCase):
+    @unittest.skipUnless(
+        _is_k8s_running(), "Kubernetes is not available")
+    def test_list_endpoints(self):
+        client = api_client.ApiClient('http://127.0.0.1:8080/')
+        api = core_v1_api.CoreV1Api(client)
+
+        endpoints = api.list_endpoints_for_all_namespaces()
+        self.assertTrue(len(endpoints.items) > 0)
+
+    @unittest.skipUnless(
+        _is_k8s_running(), "Kubernetes is not available")
+    def test_pod_apis(self):
+        client = api_client.ApiClient('http://127.0.0.1:8080/')
+        api = core_v1_api.CoreV1Api(client)
+
+        name = 'test-' + str(uuid.uuid4())
+
+        pod_manifest = {'apiVersion': 'v1',
+                        'kind': 'Pod',
+                        'metadata': {'color': 'blue', 'name': name},
+                        'spec': {'containers': [{'image': 'dockerfile/redis',
+                                                 'name': 'redis'}]}}
+
+        resp = api.create_namespaced_pod(body=pod_manifest,
+                                         namespace='default')
+        self.assertEqual(name, resp.metadata.name)
+        self.assertTrue(resp.status.phase)
+
+        resp = api.read_namespaced_pod(name=name,
+                                       namespace='default')
+        self.assertEqual(name, resp.metadata.name)
+        self.assertTrue(resp.status.phase)
+
+        number_of_pods = len(api.list_pod_for_all_namespaces().items)
+        self.assertTrue(number_of_pods > 0)
+
+        resp = api.delete_namespaced_pod(name=name, body={},
+                                         namespace='default')
+
+    @unittest.skipUnless(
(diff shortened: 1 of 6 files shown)


dims:
@mbohlool ah looks like the TODO's i have here are stemming from the same bug already reported! https://github.com/kubernetes-incubator/client-python/issues/93

kubernetes/client/api_client.py
```diff
@@ -193,7 +193,7 @@ def sanitize_for_serialization(self, obj):
                     for sub_obj in obj]
         elif isinstance(obj, tuple):
             return tuple(self.sanitize_for_serialization(sub_obj)
-                for sub_obj in obj)
+                         for sub_obj in obj)
```
mbohlool: scripts/update-pep8.sh does not update this file. If this is the right style we should have, we need to be able to automate it there.

kubernetes/test/test_k8sclient.py
```diff
@@ -0,0 +1,204 @@
+# -*- coding: utf-8 -*-
```
mbohlool: the test folder will be removed and re-generated by scripts/update-client.sh. You need to either change the script there to keep this file or move this file to another folder. maybe put it in kubernetes/e2e_test folder or something like that? Also would be nice if we can have an script in scripts/ folder to deploy a minikube and run this against it.

scripts/kube-init.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
```
mbohlool: add copyright header here.

scripts/kube-init.sh
```diff
@@ -0,0 +1,112 @@
...
+    curl -sSL https://get.docker.io | sudo bash
+fi
+docker --version
+
+# Get the latest stable version of kubernetes
+export K8S_VERSION=$(curl -sS https://storage.googleapis.com/kubernetes-release/release/stable.txt)
+echo "K8S_VERSION : ${K8S_VERSION}"
+
+echo "Starting docker service"
+sudo systemctl enable docker.service
```
mbohlool: Is this supported on osx? if not, we should check at the start of script and fail for osx with a relevant message. it may not be hard to just support osx too.

scripts/kube-init.sh
```diff
@@ -0,0 +1,112 @@
...
+sudo systemctl start docker.service --ignore-dependencies
+echo "Checking docker service"
+sudo docker ps
+
+# Run the docker containers for kubernetes
+echo "Starting Kubernetes containers"
+sudo docker run \
```
mbohlool: most probably sudo is not necessary on osx.

scripts/kube-init.sh
```diff
@@ -0,0 +1,112 @@
...
+    /hyperkube kubelet \
+        --containerized \
+        --hostname-override="127.0.0.1" \
+        --address="0.0.0.0" \
+        --api-servers=http://localhost:8080 \
+        --config=/etc/kubernetes/manifests \
+        --allow-privileged=true --v=2
+
+
+echo "Download Kubernetes CLI"
```
mbohlool: we can first check if kubectl is available. if you want to always download an stable version, would suggest to download it in a temp folder and remove it at the end.

scripts/kube-init.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
+
```
mbohlool: How/where do we use this file? If it is a manual run, it would be nice if we have an example here.

tox.ini
```diff
@@ -1,10 +1,34 @@
...
+deps = -r{toxinidir}/test-requirements.txt
+commands =
    nosetests \
       []
+
+[testenv:py27-functional]
+commands = {toxinidir}/scripts/kube-init.sh nosetests []
+
+[testenv:coverage]
+commands = nosetests --with-coverage --cover-package=kubernetes.config,kubernetes.watch --cover-tests
```
mbohlool: I don't see code coverage comment from codecov-io. If we do the code coverage right, we should get such a comment. The comment look like this: https://github.com/kubernetes-incubator/client-python/pull/89#issuecomment-271661688

mbohlool:
@dims I had some comment. Let me know if they make sense. I was looking into have some kind of e2e test from the beginning. This is great. Thanks.

kubernetes/client/api_client.py
dims: Reverted

kubernetes/test/test_k8sclient.py
dims: Moved the file to e2e_test

scripts/kube-init.sh
dims: Done

scripts/kube-init.sh
dims: it's used when you run "tox -e py27-functional" so will run on travis

scripts/kube-init.sh
dims: This is mainly meant for travis. we can look into supporting other platforms. As mentioned, you really don't need this script, as long as you have localhost:8080 proxied to a working k8s instance for the tests to run.

scripts/kube-init.sh
dims: This is working on travis. can tweak for OSX later

scripts/kube-init.sh
dims: Ack. I can do this in a followup PR as this is mainly for the travis environment right now where we know kubectl is not available

tox.ini
dims: Ack. i'll check on this

scripts/kube-init.sh
mbohlool: Ignore all osx comments then now that I understand this is meant for travis.

scripts/kube-init.sh
mbohlool: this is for travis, no need to do so.

kubernetes/e2e_test/test_k8sclient.py
```diff
@@ -0,0 +1,204 @@
+# -*- coding: utf-8 -*-
```
mbohlool: would you please add copyright header to kubernetes/e2e_test/__init__.py file too. annoying but I think it is required on all files.

dims: Done (added header to kubernetes/e2e_test/init.py

dims:
ok @mbohlool - this is ready as well

mbohlool:
@dims LGTM. please fix the conflict. thanks.

dims:
@mbohlool Done!

PR #95, opened 2017-01-11
Initial cut at documentation

dims:
reuses existing markdowns and adds api documentation

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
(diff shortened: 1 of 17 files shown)


doc/source/conf.py
```diff
@@ -0,0 +1,82 @@
...
+htmlhelp_basename = '%sdoc' % project
+
+# Grouping the document tree into LaTeX files. List of tuples
+# (source start file, target name, title, author, documentclass
+# [howto/manual]).
+latex_documents = [
+    ('index',
+     '%s.tex' % project,
+     u'%s Documentation' % project,
+     u'OpenStack Foundation', 'manual'),
```
mbohlool: Should this be Kubernetes?

tox.ini
```diff
@@ -1,6 +1,9 @@
 [tox]
 envlist = py27, py34, py35
 
+[testenv:docs]
+commands = python setup.py build_sphinx
```
mbohlool: so we generate documents when we run test? if that is the case, it does not make sense. can we add a script in scripts folder to generate docs. If we want to add the generation here, we need to make sure it fails if there is a difference.

doc/source/conf.py
dims: Done

tox.ini
dims: This is here because if any PR checks in stuff that breaks docs generation it should be found quickly. Yes, i can add a follow up PR to do an addition script in the scripts folder later after this merges

mbohlool: by breaking docs, you mean not be able to generate it right? but what if somebody changes docs but forget to check them in? we won't catch that here, right?

dims:
@mbohlool this one is ready

PR #96, opened 2017-01-12
Test if docs can be built

dims:
Add a travis job for docs. This will make sure we can catch if a PR is likely to break docs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,13 @@ sudo: true
 services:
   - docker
 env:
-  - TOXENV=py35,codecov
-  - TOXENV=py34,codecov
-  - TOXENV=py27,codecov
-  - TOXENV=py27-functional,codecov
-  - TOXENV=py35-functional,codecov
-  - TOXENV=update-pep8,codecov
+  - TOXENV=py35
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,6 +9,7 @@ env:
   - TOXENV=py27,codecov
   - TOXENV=py27-functional,codecov
   - TOXENV=py35-functional,codecov
+  - TOXENV=docs,codecov
```
mbohlool: do we need codecov for all of these? I think it will fail for doc or functional tests. Also we have a coverage env, maybe that is enough?

dims: yeah, i'll fix it in another PR @mbohlool

mbohlool:
you have a conflict. I guess now that you are fixing the conflict you can remove all unnecessary codecov commands. Up to you though. this look good to me after fixing the conflict.

dims:
Ack @mbohlool fixing it while i fix the conflicts in this PR

dims:
@mbohlool - rebased. Let's get this in please.

mbohlool:
Great. Thanks.

PR #97, opened 2017-01-12
tox target to enforce pep8

dims:
This will help nudge future PR authors to fix style related issues

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,6 @@
 [tox]
 envlist = py27, py34, py35
 
-[testenv:docs]
-commands = python setup.py build_sphinx
-
 [testenv]
 passenv = TOXENV CI TRAVIS TRAVIS_*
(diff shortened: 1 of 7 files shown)


tox.ini
```diff
@@ -13,6 +10,14 @@ commands =
    python -V
    nosetests []
 
+[testenv:docs]
+commands =
+   python setup.py build_sphinx
+
+[testenv:update-pep8]
+commands =
+   {toxinidir}/scripts/update-pep8.sh
```
mbohlool: On thing that is not clean for me is, do we fail if update-pep8.sh change any file on travis? If we don't we should add verify-pep8.sh the same way we have verify in k8s and call that one.

dims: yes, if you see the update-pep8.sh, we run "git status" and fail if any files need updates

PR #98, opened 2017-01-14
Fix for patch_* to work

### LINKED ISSUE #93: What needs to be done to get patch_namespaced_deployment to work?
I want to update the image for one of my containers in a deployment and it seems I can't provide an UnversionedPatch. Using a valid JSON patch string, or the strategic-merge string I use in kubectl ends up with variations of 
```HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"json: cannot unmarshal string into Go value of type jsonpatch.Patch","code":500}```

Am I doing something wrong or is this not implemented yet? The link to UnversionedPatch returns a 404 https://github.com/kubernetes-incubator/client-python/blob/master/kubernetes/docs/UnversionedPatch.md If it's not implemented I'd be happy to work on it.

dims:
Hack to get patch_namespaced_config_map and patch_namespaced_service
etc to work by essentially switching over the content-type from:
application/json-patch+json
to:
application/strategic-merge-patch+json

Fixes #93

### DIFF
--- a/kubernetes/client/rest.py
+++ b/kubernetes/client/rest.py
@@ -146,7 +146,18 @@ class RESTClientObject(object):
             if method in ['POST', 'PUT', 'PATCH', 'OPTIONS', 'DELETE']:
                 if query_params:
                     url += '?' + urlencode(query_params)
-                if re.search('json', headers['Content-Type'], re.IGNORECASE):
+                if headers['Content-Type'] == 'application/json-patch+json':
+                    headers[
+                        'Content-Type'] = 'application/strategic-merge-patch+json'
+                    request_body = None
+                    if body:
+                        request_body = json.dumps(body)
+                    r = self.pool_manager.request(method, url,
+                                                  body=request_body,
+                                                  preload_content=_preload_content,
+                                                  timeout=timeout,
+                                                  headers=headers)
+                elif re.search('json', headers['Content-Type'], re.IGNORECASE):
                     request_body = None
                     if body:
                         request_body = json.dumps(body)
--- a/kubernetes/e2e_test/test_client.py
+++ b/kubernetes/e2e_test/test_client.py
@@ -105,17 +105,15 @@ class TestClient(unittest.TestCase):
         self.assertEqual('frontend', resp.metadata.name)
         self.assertTrue(resp.status)
 
-        # TODO(dims) : Fails with "json: cannot unmarshal object into
-        # Go value of type jsonpatch.Patch"
-        # service_manifest['spec']['ports'] = [{'name': 'new',
-        #                                       'port': 8080,
-        #                                       'protocol': 'TCP',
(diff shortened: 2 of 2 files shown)


mbohlool:
That would definitely fix #93. From the description, it look like we just change the content type (and if that is the case, I would rather do it in preprocessing step on the spec instead of in the code), but we also deal with the fact that for those type of APIs, we get an empty model and the model does not have anything in it, thus make it impossible to pass data. I would suggest we create a special model for these cases (or even better, edit their models) that accepts json and do the right thing to pass that json to the api call instead of hacking the client, what do you think?

dims:
yes, this was a straight port of hack from what we had in python-k8sclient. yes, we can do slightly better as you suggest.

mbohlool:
Can we at least have an issue and a TODO attached to that code pointing to that issue? (also this branch has conflicts need to be fixed)

dims:
@mbohlool i'll iterate on this PR and ping you back when ready. thanks.

mbohlool:
Oh, cool, I was working on side for a fix for this and intstr both but look like you are working on it. You need to edit client-update.sh to keep your changes. Let me know if I can be of any help.

dims:
@mbohlool : ouch. sorry. i've updated the client-update.sh as well (simple git checkout seems to be enough).

mbohlool:
That will work, however if you actually had changed that file, it will undo your changes too. Don't worry about that though, I have the changed script and can send a PR for that. Keep the git checkout for now.

mbohlool:
and let me know when this is ready for review.

dims:
@mbohlool this is ready for review now. thanks

mbohlool:
I thought the idea of adding a model for the patch is to not hack the rest client, but I still see the hack in. My suggestion is to add a custom json serializing method (name to be decided) to the model and in the rest client serializing (sanitize_for_serialization), check for that method and use it if exists, to serialize the model. This way you shouldn't need any hack.

dims:
@mbohlool : ok i think i misunderstood. If you want to take a whack at adding a custom json serializing method, that would be cool. thanks. please feel free to reuse any of the code/test in this PR. (or even modifying this PR).

Thanks,
Dims

mbohlool:
Sure, I can do that. Lets just merge the first commit of this PR (the one with the hack), can you please remove the rest of the commits? (probably push them to another branch on your repo so we can refer to them later).

mbohlool:
I don't think I can modify this PR as it is coming form your repo. Please keep the first commit and rebase and we can merge this. Thanks.

dims:
Done. thanks @mbohlool

mbohlool:
@dims The PR needs rebase.

dims:
@mbohlool oops done.

PR #103, opened 2017-01-19
Bugfix: kubernetes/client/__init__.py was not updated resulting in not exported models

### LINKED ISSUE #101: `V1HostPathVolumeSource` is not imported into `kubernetes.client`
It does make it to the `kubernetes.client.models` [via this line](https://github.com/kubernetes-incubator/client-python/blob/cd2342dc2e220eff42040f82d4f53f3b5d108cd6/kubernetes/client/models/__init__.py#L98), but then it gets lost in [`kubernetes/client/__init__.py`](https://github.com/kubernetes-incubator/client-python/blob/cd2342dc2e220eff42040f82d4f53f3b5d108cd6/kubernetes/client/__init__.py).

I would submit a PR, but it looks like the code is auto-generated, so it should be fixed somewhere else.

Either way, awesome work @mbohlool, many thanks for that!

mbohlool:
fixes #101


dims:
LGTM

seeker89:
Looks good (for https://github.com/kubernetes-incubator/client-python/issues/101). Thanks !

PR #108, opened 2017-01-25
Add example to create a deployment from yaml file

### LINKED ISSUE #63: Models <-> Json/Yaml
In addition to fully wrapping the k8s client and replacing kubectl, I've found it very useful to be able to generate raw JSON/YAML for objects created using the swagger generated schema models.

This makes a few other deployment methods easier, such as Spread or Helm.

Thoughts on whether it's worth adding to this project, and where the cleanest place would be? Seems like it should likely be some utility functions in the model layer, rather than patching the client (unless there's a built-in way of doing this with the swagger-codegen client).

mbohlool:
fixes #63

### DIFF
--- a/examples/create_deployment.py
+++ b/examples/create_deployment.py
@@ -1,0 +1,37 @@
+# Copyright 2016 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


dims:
Looks like there's a pep8 issue. we should add a e2e test for this as well when possible

mbohlool:
Thanks @dims. I will add an e2e test later.

PR #110, opened 2017-01-25
More e2e tests

dims:
Add more tests :)


mbohlool:
More tests. Fantastic. Thanks @dims

Issue #48, opened 2016-11-27
Make a release script with a consistent behavior on different enviroment


mbohlool:
**Problem:** We have two identical package for alpha2 in pypi with different names. 

**diagnosis:** That is because we build and release the same thing on different environments (linux vs osx, different version of python).

**Solution:** we need to have a script to make sure the environment we are making release on, is consistent. Probably running release inside a docker container is the best solution.

PR #113, opened 2017-01-25
Add release process

mbohlool:
ref #48


dims:
LGTM as a first step. thanks for writing this up @mbohlool

mbohlool:
Yes, it definitely can be improved. But it is better than having nothing.

PR #120, opened 2017-02-01
Implementation for /exec using websocket

### LINKED ISSUE #58: Exec calls
```json
{
  "kind":"Status",
  "apiVersion":"v1",
  "metadata":{},
  "status":"Failure",
}
```

dims:
inspired by the POC from @chekolyn

* Adds a new requirement on websocket-client
* Add a new class WSClient that uses WebSocketApp from the websocket-client.
* Make sure we pass Authorization header
* Make sure we honor the SSL settings in configuration
* Some of the code will get overwritten when we generate fresh classes from swagger definition. To remind us added a e2e test so we don't lose the changes

Reference #58 (by mbohlool: Removed Fix. This PR is definitely a huge step toward the right direction, but I don't call 58 done yet.)

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -21,6 +21,7 @@ Copyright 2016 SmartBear Software
 from __future__ import absolute_import
 
 from . import models
+from . import ws_client
 from .rest import RESTClientObject
 from .rest import ApiException
 
@@ -343,6 +344,15 @@ class ApiClient(object):
         """
         Makes the HTTP request using RESTClient.
         """
+        # FIXME(dims) : We need a better way to figure out which
+        # calls end up using web sockets
+        if url.endswith('/exec') and method == "GET":
+            return ws_client.GET(self.config,
+                                 url,
+                                 query_params=query_params,
+                                 _request_timeout=_request_timeout,
+                                 headers=headers)
+
         if method == "GET":
             return self.rest_client.GET(url,
                                         query_params=query_params,
--- a/kubernetes/client/ws_client.py
+++ b/kubernetes/client/ws_client.py
@@ -1,0 +1,114 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from .rest import ApiException
+
+import certifi
+import collections
+import websocket
+import six
+import ssl
+from six.moves.urllib.parse import urlencode
+from six.moves.urllib.parse import quote_plus
+
+
+class WSClient:
+    def __init__(self, configuration, url, headers):
+        self.messages = []
+        self.errors = []
+        websocket.enableTrace(False)
+        header = None
+
+        # We just need to pass the Authorization, ignore all the other
+        # http headers we get from the generated code
+        if 'Authorization' in headers:
+            header = "Authorization: %s" % headers['Authorization']
+
+        self.ws = websocket.WebSocketApp(url,
+                                         on_message=self.on_message,
+                                         on_error=self.on_error,
+                                         on_close=self.on_close,
+                                         header=[header] if header else None)
+        self.ws.on_open = self.on_open
+
+        if url.startswith('wss://') and configuration.verify_ssl:
+            ssl_opts = {
+                'cert_reqs': ssl.CERT_REQUIRED,
+                'keyfile': configuration.key_file,
+                'certfile': configuration.cert_file,
+                'ca_certs': configuration.ssl_ca_cert or certifi.where(),
+            }
+            if configuration.assert_hostname is not None:
+                ssl_opts['check_hostname'] = configuration.assert_hostname
+        else:
+            ssl_opts = {'cert_reqs': ssl.CERT_NONE}
+
+        self.ws.run_forever(sslopt=ssl_opts)
+
+    def on_message(self, ws, message):
+        if message[0] == '\x01':
+            message = message[1:]
+        if message:
+            if six.PY3 and isinstance(message, six.binary_type):
+                message = message.decode('utf-8')
+            self.messages.append(message)
+
+    def on_error(self, ws, error):
+        self.errors.append(error)
+
+    def on_close(self, ws):
+        pass
+
+    def on_open(self, ws):
+        pass
+
+
+WSResponse = collections.namedtuple('WSResponse', ['data'])
+
+
+def GET(configuration, url, query_params, _request_timeout, headers):
+    # switch protocols from http to websocket
+    url = url.replace('http://', 'ws://')
+    url = url.replace('https://', 'wss://')
+
+    # patch extra /
+    url = url.replace('//api', '/api')
+
+    # Extract the command from the list of tuples
+    commands = None
+    for key, value in query_params:
+        if key == 'command':
+            commands = value
+            break
+
+    # drop command from query_params as we will be processing it separately
+    query_params = [(key, value) for key, value in query_params if
+                    key != 'command']
+
+    # if we still have query params then encode them
+    if query_params:
+        url += '?' + urlencode(query_params)
+
+    # tack on the actual command to execute at the end
+    if isinstance(commands, list):
+        for command in commands:
+            url += "&command=%s&" % quote_plus(command)
+    else:
+        url += '&command=' + quote_plus(commands)
+
+    client = WSClient(configuration, url, headers)
+    if client.errors:
+        raise ApiException(
+            status=0,
+            reason='\n'.join([str(error) for error in client.errors])
+        )
+    return WSResponse('%s' % ''.join(client.messages))
--- a/kubernetes/e2e_test/test_client.py
+++ b/kubernetes/e2e_test/test_client.py
@@ -12,6 +12,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import time
 import unittest
 import uuid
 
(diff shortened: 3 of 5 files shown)


kubernetes/e2e_test/test_client.py
```diff
@@ -27,22 +28,57 @@ def test_pod_apis(self):
...
+        # print('EXEC response : %s' % resp)
+
+        exec_command = ['/bin/sh',
+                        '-c',
+                        'for i in $(seq 1 3); do date; sleep 1; done']
+        resp = api.connect_get_namespaced_pod_exec(name, 'default',
+                                                   command=exec_command,
+                                                   stderr=False, stdin=False,
+                                                   stdout=True, tty=False)
+        print('EXEC response : %s' % resp)
```
mbohlool: Can we do better instead of printing out the output? If we know what output we expect, we can check that or even part of the output.

kubernetes/client/api_client.py
```diff
@@ -343,6 +344,13 @@ def request(self, method, url, query_params=None, headers=None,
         """
         Makes the HTTP request using RESTClient.
         """
+        if url.endswith('/exec') and method == "GET":
+            return ws_client.GET(self.config,
```
mbohlool: I maybe not familiar enough with exec command, but does it let interactive connection with the host? I mean can a script send something, then read the result, then decide what next command to send, etc.? In that case, should we return some kind of client allows that instead of command result?

kubernetes/client/ws_client.py
```diff
@@ -0,0 +1,93 @@
...
+
+import websocket
+import six
+import ssl
+from six.moves.urllib.parse import urlencode
+from six.moves.urllib.parse import quote_plus
+
+
+class WSClient:
+    def __init__(self, configuration, url):
```
mbohlool: what does configuration used for?

kubernetes/client/ws_client.py
```diff
@@ -0,0 +1,93 @@
...
+        self.messages = []
+        self.errors = []
+        websocket.enableTrace(False)
+        self.ws = websocket.WebSocketApp(url,
+                                         on_message=self.on_message,
+                                         on_error=self.on_error,
+                                         on_close=self.on_close,
+                                         header=[])
```
mbohlool: should we send authorization token?

kubernetes/client/ws_client.py
```diff
@@ -0,0 +1,93 @@
...
+        self.ws.on_open = self.on_open
+        sslopt = None
+        if url.startswith('wss://'):
+            sslopt = {"cert_reqs": ssl.CERT_NONE}
```
mbohlool: Should we use ssl cert (and client cert and key) from configuration?

kubernetes/client/ws_client.py
```diff
@@ -0,0 +1,93 @@
...
+from .rest import ApiException
+
+import websocket
+import six
+import ssl
+from six.moves.urllib.parse import urlencode
+from six.moves.urllib.parse import quote_plus
+
+
+class WSClient:
```
mbohlool: I would suggest testing this (even manually) on GCE, GCP and minikube to make sure we cover a good range of authentication methods.

kubernetes/e2e_test/test_client.py
```diff
@@ -27,22 +28,57 @@ def test_pod_apis(self):
...
+                                           namespace='default')
+            self.assertEqual(name, resp.metadata.name)
+            self.assertTrue(resp.status.phase)
+            print(">>>> pod status : %r" % resp.status.phase)
+            if resp.status.phase != 'Pending':
+                break
+            else:
+                time.sleep(1)
+
+        # exec_command = 'uptime'
```
mbohlool: remove the commented out codes.

mbohlool:
Sorry, I missed [WIP] in the title. Please ignore my comments until you are done with this.

kubernetes/e2e_test/test_client.py
dims: Done

kubernetes/e2e_test/test_client.py
dims: ack. let's make sure we get 3 lines back

kubernetes/client/api_client.py
dims: Nope. the "/exec" is one shot only.

kubernetes/client/ws_client.py
dims: @mbohlool i do not have any kind of access to GCE or GCP. so i'll rely on your help with testing these

kubernetes/client/ws_client.py
dims: will add code for getting the ssl configuration options shortly

kubernetes/client/ws_client.py
dims: yep for sure :)

kubernetes/client/ws_client.py
dims: right. yes.

dims:
Tested by adding a TLS proxy in front of the K8S API

```
sudo make-ssl-cert /usr/share/ssl-cert/ssleay.cnf example.pem
hitch --frontend=[*]:8443 --backend=[localhost]:8080 example.pem
```

mbohlool:
I think I had a comment about interactiveness of this solution but I cannot find it. The solution does not let the script to interactive with the server. Like open connection, run command, get output/error, run another command, etc... Is it possible to support that?

kubernetes/client/apis/core_v1_api.py
```diff
@@ -1007,7 +1007,7 @@ def connect_get_namespaced_pod_exec_with_http_info(self, name, namespace, **kwar
                                         auth_settings=auth_settings,
                                         callback=params.get('callback'),
                                         _return_http_data_only=params.get('_return_http_data_only'),
-                                        _preload_content=params.get('_preload_content', True),
```
mbohlool: What is the purpose of this change?

chekolyn:
@dims This is awesome work!

kubernetes/client/apis/core_v1_api.py
dims: the _preload_content tries to coerce the response into a json. would be great to find another spot to put this.

dims:
@chekolyn kudos to you for finding a way :)

dims:
@mbohlool : the API definition does not allow for interactivity

kubernetes/client/apis/core_v1_api.py
mbohlool: api_client.py, if you move your changes out of `request` method into `__call_api` itself, you have access to preload_content flag. You can even branch before calling `request` to make sure the preload_content flag does not have any effect. e.g. if exec, do stuff and return, else do normal stuff of calling request and preload_content stuff. We should not change generated files as much as possible.

mbohlool:
> @mbohlool : the API definition does not allow for interactivity

you can return an object instead of an string, right? the object can have methods to work interactively with the open connection.

kubernetes/client/apis/core_v1_api.py
dims: i figured out a way to avoid this change, please see latest update

dims:
@mbohlool i am happy with what we have here in this PR now. i am probably missing something about interactivity idea that you have brought up a few times. I hope we can merge this and then iterate on it. I have added a FIXME for the "/exec" check in one spot, so we can deal with that later. Hope this is good enough.

kubernetes/client/ws_client.py
```diff
@@ -0,0 +1,114 @@
...
+        self.errors.append(error)
+
+    def on_close(self, ws):
+        pass
+
+    def on_open(self, ws):
+        pass
+
+
+WSResponse = collections.namedtuple('WSResponse', ['data'])
```
mbohlool: I like the idea of WSResponse. make it easier to add interactivity later (which I am fine with).

kubernetes/e2e_test/test_client.py
```diff
@@ -27,22 +28,49 @@ def test_pod_apis(self):
...
+
+        exec_command = ['/bin/sh',
+                        '-c',
+                        'for i in $(seq 1 3); do date; sleep 1; done']
+        resp = api.connect_get_namespaced_pod_exec(name, 'default',
+                                                   command=exec_command,
+                                                   stderr=False, stdin=False,
+                                                   stdout=True, tty=False)
+        print('EXEC response : %s' % resp)
+        self.assertEqual(3, len(resp.splitlines()))
```
mbohlool: resp is a WSResponse, right? should we check it's `data` field?

mbohlool:
@dims OK. I am fine with this change (except one comment above) and we can merge and iterate on it later. It is in a pretty good shape.

kubernetes/e2e_test/test_client.py
dims: @mbohlool the WSResponse is already unwrapped and the resp is just the string. If it were WSResponse the assert would fail :)

mbohlool: That was my problem too, why the assert didn't fail. I am just trying to run it locally (just curious to see it), then I will merge. Thanks.

kubernetes/client/ws_client.py
mbohlool: Does WSResponse needs an 'status'? and 'reason' in case of an error?

kubernetes/e2e_test/test_client.py
mbohlool: I think I got the place that WSResponse is unwrapped, now I have more questions about that, e.g. why we don't have 'status' in WSResponse? rest.py checks for an status and raise an exception if status is not in OK range.

dims:
@mbohlool : there is no rest.py in the call stack. api_client.py calls ws_client.py directly so we don't need the "status" field.

[https://www.evernote.com/shard/s412/sh/3a91433b-7125-4393-8506-2794b4565d2e/7ac292e184053c6a/res/ef38c1d4-214a-454b-b869-b2816c18087e/skitch.png](url)

mbohlool:
@dims Where we unwrap WSResponse then?

dims:
@mbohlool - we deserialize here - https://github.com/kubernetes-incubator/client-python/blob/master/kubernetes/client/api_client.py#L160

PR #121, opened 2017-02-03
Run e2e tests against against https url

dims:
Use hitch tls proxy to listen on 8443 port and forward the
traffic to 8080. Add a configuration option to ignore any
issues with host names as we don't generate the cert on the
fly.


dims:
@mbohlool : there are two projects with the same name. what i use is this (https://hitch-tls.org/) which is a TLS proxy to redirect from 8443 to 8080. The url you pasted is a test/ci framework :)

mbohlool:
@dims I understand. Thanks. Where did you get the repo name ppa:0k5....? can you add a comment there to the reference page?

PR #123, opened 2017-02-09
Switch to minikube/localkube

### LINKED ISSUE #122: Use minikube to run e2e tests

dims:
* Drop the old Kubernetes on docker containers method as it
  did not have a SSL enabled port
* Use localkube component from minikube instead. This enables
  us to drop the hitch TLS proxy as well.
* E2E tests should be easy to run locally and pick up configuration
  from ~/.kube/config
* Consolidate the urls in one spot (base.py), also consolidate the
  SkipTest in one spot.
* For local testing, just run minikube and run the py27 or py34 tox
  target, that should run all the tests including the e2e tests.
* Fix the connect_post_namespaced_pod_exec and add a e2e test for it

Fixes #122


mbohlool:
Nicely done. Thanks @dims

PR #125, opened 2017-02-15
Improvements on ws_client

### LINKED ISSUE #58: Exec calls

mbohlool:
The ws_client should be able to return an object so the caller can interact with server. This will happen if _preload_content is not True. In that case, the client will return an object that can interact with stdin/stdout/stderr of the remote container.

fixes #58

### DIFF
--- a/examples/exec.py
+++ b/examples/exec.py
@@ -1,0 +1,94 @@
+import time
+
+from kubernetes import config
+from kubernetes.client import configuration
+from kubernetes.client.apis import core_v1_api
+from kubernetes.client.rest import ApiException
+
+config.load_kube_config()
+configuration.assert_hostname = False
+api = core_v1_api.CoreV1Api()
+name = 'busybox-test'
+
+resp = None
+try:
+    resp = api.read_namespaced_pod(name=name,
+                                   namespace='default')
+except ApiException as e:
+    if e.status != 404:
+        print("Unknown error: %s" % e)
+        exit(1)
+
+if not resp:
+    print("Pod %s does not exits. Creating it..." % name)
+    pod_manifest = {
+        'apiVersion': 'v1',
+        'kind': 'Pod',
+        'metadata': {
+            'name': name
+        },
+        'spec': {
+            'containers': [{
+                'image': 'busybox',
+                'name': 'sleep',
+                "args": [
+                    "/bin/sh",
+                    "-c",
+                    "while true;do date;sleep 5; done"
+                ]
+            }]
+        }
+    }
+    resp = api.create_namespaced_pod(body=pod_manifest,
+                                     namespace='default')
+    while True:
+        resp = api.read_namespaced_pod(name=name,
+                                       namespace='default')
+        if resp.status.phase != 'Pending':
+            break
+        time.sleep(1)
+    print("Done.")
+
+
+# calling exec and wait for response.
+exec_command = [
+    '/bin/sh',
+    '-c',
+    'echo This message goes to stderr >&2; echo This message goes to stdout']
+resp = api.connect_get_namespaced_pod_exec(name, 'default',
+                                           command=exec_command,
+                                           stderr=True, stdin=False,
+                                           stdout=True, tty=False)
+print("Response: " + resp)
+
+# Calling exec interactively.
+exec_command = ['/bin/sh']
+resp = api.connect_get_namespaced_pod_exec(name, 'default',
+                                           command=exec_command,
+                                           stderr=True, stdin=True,
+                                           stdout=True, tty=False,
+
+                                           _preload_content=False)
+commands = [
+    "echo test1",
+    "echo \"This message goes to stderr\" >&2",
+]
+while resp.is_open():
+    resp.update(timeout=1)
+    if resp.peek_stdout():
+        print("STDOUT: %s" % resp.read_stdout())
+    if resp.peek_stderr():
+        print("STDERR: %s" % resp.read_stderr())
+    if commands:
+        c = commands.pop(0)
+        print("Running command... %s\n" % c)
+        resp.write_stdin(c + "\n")
+    else:
+        break
+
+resp.write_stdin("date\n")
+sdate = resp.readline_stdout(timeout=3)
+print("Server date command returns: %s" % sdate)
+resp.write_stdin("whoami\n")
+user = resp.readline_stdout(timeout=3)
+print("Server user is: %s" % user)
--- a/kubernetes/client/ws_client.py
+++ b/kubernetes/client/ws_client.py
@@ -12,33 +12,40 @@
 
 from .rest import ApiException
 
+import select
 import certifi
+import time
 import collections
-import websocket
+from websocket import WebSocket, ABNF, enableTrace
 import six
 import ssl
 from six.moves.urllib.parse import urlencode
 from six.moves.urllib.parse import quote_plus
 
+STDIN_CHANNEL = 0
+STDOUT_CHANNEL = 1
+STDERR_CHANNEL = 2
+
 
 class WSClient:
     def __init__(self, configuration, url, headers):
-        self.messages = []
-        self.errors = []
-        websocket.enableTrace(False)
-        header = None
+        """A websocket client with support for channels.
+
+            Exec command uses different channels for different streams. for
+        example, 0 is stdin, 1 is stdout and 2 is stderr. Some other API calls
+        like port forwarding can forward different pods' streams to different
+        channels.
+        """
+        enableTrace(False)
+        header = []
+        self._connected = False
+        self._channels = {}
+        self._all = ""
 
         # We just need to pass the Authorization, ignore all the other
         # http headers we get from the generated code
-        if 'Authorization' in headers:
-            header = "Authorization: %s" % headers['Authorization']
-
-        self.ws = websocket.WebSocketApp(url,
-                                         on_message=self.on_message,
-                                         on_error=self.on_error,
-                                         on_close=self.on_close,
-                                         header=[header] if header else None)
-        self.ws.on_open = self.on_open
+        if headers and 'authorization' in headers:
+            header.append("authorization: %s" % headers['authorization'])
 
         if url.startswith('wss://') and configuration.verify_ssl:
             ssl_opts = {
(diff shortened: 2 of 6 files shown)


kubernetes/client/ws_client.py
```diff
@@ -52,30 +49,87 @@ def __init__(self, configuration, url, headers):
...
+        self._stdout = ""
+        self._stderr = ""
+        return out
+
+    def is_open(self):
+        return self._connected
+
+    # TODO: This method does not seem to work.
+    def write_stdin(self, data):
+        self.sock.send(data)
```
mrmcmuffinz: Is the connection still open to even send the data? How does the underlining api url call look like when you try to tack on more `exec` commands to an already successful call?

examples/exec.py
```diff
@@ -0,0 +1,94 @@
...
+                                           _preload_content=False)
+while resp.is_open():
+    resp.update(timeout=1)
+    if resp.peek_stdout():
+        print("STDOUT: %s" % resp.read_stdout())
+    if resp.peek_stderr():
+        print("STDERR: %s" % resp.read_stderr())
+
+exit(1)
+# This part does not work yet. resp.write_stdin does not work.
```
mrmcmuffinz: Could it be that you need to set the websocket to lingering, so that the connection stays open on both ends till the action you are trying to execute completes and forwards back the result?

mbohlool: In the example bellow, I only run sh without closing the connection. _preload_content is False so the call will stays open. Then on line 94 inside the loop I send commands one by one. This example code will not end by itself because the connection stays open.

kubernetes/client/ws_client.py
mbohlool: The connection stays open until explicitly closed by either side. My example code does not close the connection before sending data.

mbohlool:
@mrmcmuffinz @dims this is ready for review.

mrmcmuffinz:
If you don't mind I can review it tomorrow, I'm on the east coast and it's close to midnight here. I want to get some sleep.

mbohlool:
@mrmcmuffinz no rush. thanks.

examples/exec.py
```diff
@@ -0,0 +1,89 @@
+import time
+
+from kubernetes import config
+from kubernetes.client import configuration
+from kubernetes.client.apis import core_v1_api
+from kubernetes.client.rest import ApiException
+
+config.load_kube_config(
+    context="gke_cloud-kubernetes-dev_us-central1-f_mehdy-cluster")
```
dims: have a more generic name here?

dims:
@mbohlool : looks good, i haven't tried it yet. Though that reminds me that we should add a e2e test especially for the new code path where we pull data on demand.

examples/exec.py
mbohlool: Done.

mbohlool:
@dims. Will add an e2e test.

kubernetes/client/ws_client.py
```diff
@@ -23,22 +25,17 @@
...
+        enableTrace(False)
+        header = []
+        self._connected = False
+        self._stdout = ""
+        self._stderr = ""
+        self._all = ""
 
         # We just need to pass the Authorization, ignore all the other
         # http headers we get from the generated code
         if 'Authorization' in headers:
```
mrmcmuffinz: Do we need to be concerned that headers is None or that there is no `Authorization` in it?

kubernetes/client/ws_client.py
```diff
@@ -52,30 +49,87 @@ def __init__(self, configuration, url, headers):
...
+            self.update(timeout=None)
+        ret = self._stdout
+        self._stdout = ""
+        return ret
+
+    def peek_stderr(self):
+        self.update()
+        return self._stderr
+
+    def read_stderr(self):
```
mrmcmuffinz: Where was this in the example?

kubernetes/client/ws_client.py
```diff
@@ -52,30 +49,87 @@ def __init__(self, configuration, url, headers):
...
+        if not self._stderr:
+            self.update(timeout=None)
+        ret = self._stderr
+        self._stderr = ""
+        return ret
+
+    def read_all(self):
```
mrmcmuffinz: Where was this in the example?

kubernetes/client/ws_client.py
```diff
@@ -12,9 +12,11 @@
 
 from .rest import ApiException
 
+import select
 import certifi
+import time
 import collections
-import websocket
+from websocket import WebSocket, ABNF, enableTrace
```
mrmcmuffinz: Just a suggestion however can we organize these, `import` first then `from`?

kubernetes/client/ws_client.py
```diff
@@ -52,30 +49,87 @@ def __init__(self, configuration, url, headers):
...
+        out = self._all
+        self._all = ""
+        self._stdout = ""
+        self._stderr = ""
+        return out
+
+    def is_open(self):
+        return self._connected
+
+    # TODO: This method does not seem to work.
```
mrmcmuffinz: If this does not work, do it make sense to keep it in the code or remove it?

kubernetes/client/ws_client.py
```diff
@@ -52,30 +49,87 @@ def __init__(self, configuration, url, headers):
...
+            self._connected = False
+            return
+        r, _, _ = select.select(
+            (self.sock.sock, ), (), (), timeout)
+        if r:
+            op_code, frame = self.sock.recv_data_frame(True)
+            if op_code == ABNF.OPCODE_CLOSE:
+                self._connected = False
+                return
+            elif op_code == ABNF.OPCODE_BINARY or op_code == ABNF.OPCODE_TEXT:
```
mrmcmuffinz: Is there any other `op_code` that needs to be considered other than the ones you caught here and above?

kubernetes/client/ws_client.py
```diff
@@ -52,30 +49,87 @@ def __init__(self, configuration, url, headers):
...
+            while self.is_open():
+                self.update(timeout=None)
 
 
 WSResponse = collections.namedtuple('WSResponse', ['data'])
 
 
-def GET(configuration, url, query_params, _request_timeout, headers):
+def GET(configuration, url, query_params, _request_timeout, _preload_content,
+        headers):
```
mrmcmuffinz: Documentation?

kubernetes/client/ws_client.py
```diff
@@ -52,30 +49,87 @@ def __init__(self, configuration, url, headers):
...
+                    data = data[1:]
+                if data:
+                    self._all += data
+                    if data[0] == '\x02':
+                        self._stderr += data[1:]
+                    else:
+                        self._stdout += data
+
+    def run_forever(self, timeout=None):
+        if timeout:
```
mrmcmuffinz: Can we get an explanation on what run_forever means and are there any gotchas(what are the assumptions you are making)?

kubernetes/client/ws_client.py
mbohlool: I don't think so.

kubernetes/client/ws_client.py
mbohlool: old comment. removed.

kubernetes/client/ws_client.py
mbohlool: it used later in this file in GET method for non-interactive case.

kubernetes/client/ws_client.py
mbohlool: end of the example, I've used this method.

kubernetes/client/ws_client.py
mbohlool: if check to see if there is authorization in it. if there is none, we don't need to pass any header. Fixed the case that headers is None.

examples/exec.py
```diff
@@ -0,0 +1,89 @@
...
+# calling exec and wait for response.
+exec_command = [
+    '/bin/sh',
+    '-c',
+    'echo This message goes to stderr >&2; echo This message goes to stdout']
+resp = api.connect_get_namespaced_pod_exec(name, 'default',
+                                           command=exec_command,
+                                           stderr=True, stdin=False,
+                                           stdout=True, tty=False)
+print "Response: " + resp
```
mbohlool: There are many cases that we can cover. We can add more examples later (and feel free to add more).

kubernetes/client/ws_client.py
mbohlool: we have pep8 and sort_import running automatically and checking these. We cannot organize them the way we want unless we want to remove those checks.

examples/exec.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+# Calling exec interactively.
+exec_command = ['/bin/sh']
+resp = api.connect_get_namespaced_pod_exec(name, 'default',
+                                           command=exec_command,
+                                           stderr=True, stdin=True,
```
mbohlool: 1. stdin is part of auto-generated arguments that we never touch.
2. stdin is working now :)

kubernetes/client/ws_client.py
```diff
@@ -23,22 +25,17 @@
 
 class WSClient:
     def __init__(self, configuration, url, headers):
```
mbohlool: I will add some documentation for methods in this class. __init__ specifically does not need docs in my opinion as this class is only created inside this file.

kubernetes/client/ws_client.py
mbohlool: will add documentation.

kubernetes/client/ws_client.py
mbohlool: GET is internal, but above methods need docs that will be added.

mbohlool:
@mrmcmuffinz, @dims addressed comments and also make the implementation more general to support any channel number (0,1,2 for std in,out,err). We may want to reuse this again in port forwarding api calls (they use websocket too). PTAL.

examples/exec.py
mrmcmuffinz: Awesome!

kubernetes/client/ws_client.py
mrmcmuffinz: I see.

kubernetes/client/ws_client.py
mrmcmuffinz: I don't know much about sockets however I see that there are more codes https://github.com/websocket-client/websocket-client/blob/master/websocket/_abnf.py#L102 where I thought we may have to cover our bases but I would like to leave that up to someone else who is more experienced in the area.

mrmcmuffinz:
General comment, was your intention to expose the channel constants for the end user like myself or just for the purpose of the websocket client?

kubernetes/client/ws_client.py
mbohlool: I've modeled this from WebSocketApp here: https://github.com/websocket-client/websocket-client/blob/master/websocket/_app.py#L205. Ping and Pong are optional (and we can support it later). Cont can be ignored and it is only useful if you want to get notified on an incomplete frame. however, we may improve this later by doing something on "Cont" (like running update again).

mbohlool:
@mrmcmuffinz Channel numbering is a kubernetes thing as far as I can tell and I think it is not bad to let user know about it. But the helper function like read_stdout can hide that if user don't want to care about them. The whole channels expose can be useful later for port forwarding api implementation.

kubernetes/client/ws_client.py
```diff
@@ -12,33 +12,40 @@
...
-        if 'Authorization' in headers:
-            header = "Authorization: %s" % headers['Authorization']
-
-        self.ws = websocket.WebSocketApp(url,
-                                         on_message=self.on_message,
-                                         on_error=self.on_error,
-                                         on_close=self.on_close,
-                                         header=[header] if header else None)
-        self.ws.on_open = self.on_open
+        if headers and 'Authorization' in headers:
```
chekolyn: I'm running into a small issue when running the exec.py example using header authentication. 
Looks like is a small case condition; it appears that the header key is "authorization" instead of "Authorization" and then we pass an empty header to the WS call.
```
(kubernetes-local) PAOC02S79NJG8WP:examples sa$ python exec.py
{'Content-Type': 'application/json', 'authorization': 'Bearer CLSMM1PCEbKNOgy2fz5runETCrf-yzV1ThPXAiJ4aLw', 'Accept': '*/*', 'User-Agent': 'Swagger-Codegen/1.0.0-snapshot/python'}
  File "exec.py", line 61, in <module>
    stdout=True, tty=False)
  File "/Users/sagui10/git/kubernetes/client-python/kubernetes/client/apis/core_v1_api.py", line 907, in connect_get_namespaced_pod_exec
    (data) = self.connect_get_namespaced_pod_exec_with_http_info(name, namespace, **kwargs)
  File "/Users/sagui10/git/kubernetes/client-python/kubernetes/client/apis/core_v1_api.py", line 1012, in connect_get_namespaced_pod_exec_with_http_info
  File "/Users/sagui10/git/kubernetes/client-python/kubernetes/client/api_client.py", line 329, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/Users/sagui10/git/kubernetes/client-python/kubernetes/client/api_client.py", line 153, in __call_api
    _request_timeout=_request_timeout)
  File "/Users/sagui10/git/kubernetes/client-python/kubernetes/client/api_client.py", line 355, in request
    headers=headers)
  File "/Users/sagui10/git/kubernetes/client-python/kubernetes/client/ws_client.py", line 239, in websocket_call
    raise ApiException(status=0, reason=str(e))
kubernetes.client.rest.ApiException: (0)
Reason: Handshake status 403
```

> I added this to debug and make it work:
> 
>         # We just need to pass the Authorization, ignore all the other
>         # http headers we get from the generated code
>         if headers and 'Authorization' in headers:
>             header.append("Authorization: %s" % headers['Authorization'])
>         elif headers and 'authorization' in headers:
>             header.append("Authorization: %s" % headers['authorization'])
> 
>         print headers

mrmcmuffinz: @chekolyn brings up a good point, however instead of doing the `elif headers and 'authorization' in headers:` we just need to do a case insensitive search instead. What do you think @mbohlool ?

mbohlool: This should be always authorization with all lower case. I wonder why it worked for me before, I am investigating it. But I am agains making the check case insensitive as it will result in being inconsistent in different part of the code while we know this should be always lowecase.

mbohlool: I've fixed this to be lower case authorization to be consistent with the rest of the client, also created #128 to further investigate this.

kubernetes/client/ws_client.py
```diff
@@ -12,33 +12,40 @@
...
-        if 'Authorization' in headers:
-            header = "Authorization: %s" % headers['Authorization']
-
-        self.ws = websocket.WebSocketApp(url,
-                                         on_message=self.on_message,
-                                         on_error=self.on_error,
-                                         on_close=self.on_close,
-                                         header=[header] if header else None)
-        self.ws.on_open = self.on_open
+        if headers and 'authorization' in headers:
```
dims: Should we support both `Authorization` and `authorization`?

dims:
One question inline, LGTM otherwise. thanks @mbohlool

kubernetes/client/ws_client.py
mbohlool: I think we should be consistent. We used lowercase `authorization` everywhere else in our client. Based on the discussion in #128, I think we should have capital `Authorization` everywhere and case-insensitive checks. I will let another PR fixing #128 to do it consistently for whole client and keep this here to be consistent meanwhile.

mbohlool:
thanks @dims @mrmcmuffinz @chekolyn

mrmcmuffinz:
Hey @mbohlool to be clear I can not set in my pip requirements.txt `v1.0.0b2` for kubernetes client-python right?

mrmcmuffinz:
Btw, thanks for being patient and following through on the comments. Awesome work!

chekolyn:
Second that, awesome work!
@mbohlool with your consent I would like to port your WSClient implementation to [pykube](https://github.com/kelproject/pykube). Both projects have the same Apache License, would that be OK?

mbohlool:
@chekolyn That would be fine. But I think it is better for pykube to add a dependency on client-python. It could simplify their config and transport layers too (by delegating them to client-python) and get any patch/update for free.

mbohlool:
@mrmcmuffinz you're welcome. v1.0.0b2 has a dependency problem that @dims is fixing. Was your question related to that?

PR #132, opened 2017-02-22
Sync requirements and add license info

### LINKED ISSUE #131: No module named websocket in b2
Hello,

I install the client using pip and since b2 I've this error "No module named websocket".
Installing it manually works fine.

I've not much experience in python packaging etc... but it seems this dependency is missing from setup.py!

dims:
Follow guidelines in:
https://packaging.python.org/requirements/

* Make sure the same list of packages are in setup.cfg and
  requirements.txt
* install_requires reqs in setup.cfg should be "Abstract" with
  no version information
* Add License information for each package

Fixes #131


mrmcmuffinz:
@dims Do we have to bump up the release version with these changes?

dims:
@mrmcmuffinz : i'd vote to release another beta to make sure this works fine for folks who want to try it (cc @mbohlool)

mbohlool:
@dims, this looks good. Can you add an entry in CHANGELOG in the same PR? I want to start requiring that for each PR to make sure we always have updated change log with correct references. I can do another release after this.

mbohlool:
@dims, I didn't realize the release is broken, I will go ahead and add the changelog myself to do another release.

mbohlool:
I've released 1.0.0b3. I think we should have an automated check for this to make sure it won't happen in future. Will add an issue for it.

PR #142, opened 2017-02-24
Remove channel id from exec response in blocking mode

### LINKED ISSUE #140: exec response returning with extra data
**Scenario**
I'm trying to check if a specific files exists in my pod and assert it on my end using a direct string to string comparison.

**Code Snippet**
```python
cmd = "[ -f /etc/hosts ] && echo 'True' || echo 'False'"
response = kubernetes_client.connect_post_namespaced_pod_exec(pod, kubernetes_namespace, stderr=True, stdin=True, stdout=True, command=cmd, tty=False)
is_file = response.strip()
print is_file
print [ord(c) for c in is_file]
print len(is_file)
```

**Output**
```
True
[1, 1, 84, 114, 117, 101]
6
```

**Problem**
Before in kubernetes v1.0.0.b1, I did not have those extra characters in the beginning of my response but with the latest v1.0.0b3 I do. Anyone know why this is the case?

mbohlool:
Fixes #140

### DIFF
--- a/kubernetes/client/ws_client.py
+++ b/kubernetes/client/ws_client.py
@@ -168,11 +168,13 @@ class WSClient:
                 data = frame.data
                 if six.PY3:
                     data = data.decode("utf-8")
-                self._all += data
                 if len(data) > 1:
                     channel = ord(data[0])
                     data = data[1:]
                     if data:
+                        # keeping all messages in the order they received for
+                        # non-blocking call.
+                        self._all += data
                         if channel not in self._channels:
                             self._channels[channel] = data
                         else:


dims:
LGTM 👍

mrmcmuffinz:
Do we have to bump up the release version now?

mbohlool:
So the plan for stable release is for next Wednesday. I think this change was small and I will wait till Sunday to see if we find anything else. I will do a release on Sunday or early Monday to be prepared for Wednesday stable release. Please keep testing :)

PR #146, opened 2017-02-27
Add close method to websocket client

### LINKED ISSUE #145: Interactive Exec on WebSocket api Close() requirement
Had a discussion with @mbohlool on slack about externalizing a close() method for websocket client. Scenario describe below:

**Scenario:**

1. I have opened up an interactive exec.
2. Executed a few commands.
3. Completed my job.

**Proposal:**

1. We need to externalize some method to close the socket in the event we are done using it and don't want the socket lingering around. Yes we could use an `exit` as part of our commands to run but does that really clean up the socket we opened up on our end?

mbohlool:
fixes #145


kubernetes/client/ws_client.py
```diff
@@ -191,6 +191,14 @@ def run_forever(self, timeout=None):
             while self.is_open():
                 self.update(timeout=None)
 
+    def close(self, **kwargs):
+        """
+        close websocket connection.
+        """
+        self._connected = False
```
mrmcmuffinz: Should we check that the socket is open before closing it?

kubernetes/client/ws_client.py
```diff
@@ -191,6 +191,14 @@ def run_forever(self, timeout=None):
             while self.is_open():
                 self.update(timeout=None)
 
+    def close(self, **kwargs):
```
mrmcmuffinz: What are some of the variable arguments you envision being used here? Just trying to get a feel for this change...

mrmcmuffinz: Also, just out of curiosity, should we return True/False if we were able to close successfully the socket?

mbohlool: Look at WebScoket close method for more information: https://github.com/websocket-client/websocket-client/blob/master/websocket/_core.py#L372
that method does not return any status.

kubernetes/client/ws_client.py
mbohlool: close method is safe to call even if the socket is closed.

kubernetes/client/ws_client.py
mrmcmuffinz: Got it, thanks.

kubernetes/client/ws_client.py
mrmcmuffinz: Got it, thanks.

PR #147, opened 2017-02-28
Added OpenSSL version requirement to README.

WilliamDenniss:


WilliamDenniss:
I signed it!

README.md
```diff
@@ -6,6 +6,15 @@
 
 Python client for the [kubernetes](http://kubernetes.io/) API.
 
+## Requirements
+
+A version of python with OpenSSL version 1.0.0 or later.
```
mbohlool: I am not sure if this is the minimum requirement. Worth mentioning this is on mac. What we know is, the client does not work with current OPENSSL_VERSION default on osx.

WilliamDenniss: Ah ok, should I make this a macOS-specific instruction then? If so, I'd mention using brew as well.

mbohlool: Yes, that's more appropriate I think. Thanks.

mbohlool:
@WilliamDenniss Hey William, are you going to update this PR? I can send another one to fix it if you don't have time.

PR #158, opened 2017-03-14
Update README.md with ssl hostname troubleshooting [ci skip]

mbohlool:


mbohlool:
@WilliamDenniss would you please review this? I think we can make error messages more descriptive for both of these troubleshootings, but having them in readme file should be ok for now.

README.md
```diff
@@ -109,3 +109,9 @@ python -c "import ssl; print ssl.OPENSSL_VERSION"
 ```
 
 You'll need a version with OpenSSL version 1.0.0 or later.
+
+### Hostname doesn't match
+
+If you get an ssl.CertificateError complaining about hostname match, your installed packages does not meet version [requirements](requirements.txt). 
+Specificly check ipaddress and urllib3 package versions to make sure they met requiurments in [requirements.txt](requirements.txt) file.
```
WilliamDenniss: Typos: requiurments  -> requirements.  Specificly -> Specifically.

Also, I'd add backquotes to the packages, like `ipaddress`.

mbohlool:
Thanks.

Issue #169, opened 2017-04-04
Update preprocessing script to shorten model names


mbohlool:
Starting kubernetes 1.6, the OpenAPI spec has full package name of the models. These long names are really inconviniant in python client but it has been added because there were conflicts in short names. We should modify preprocessing script to shorten these names as much as possible (possibly keep model names for most of the existing models). This, however, could change some of the names and hence it is a backward incompatible change. We need to increase major version of the client for that.

mbohlool:
More clarification:
kubernetes moved to full model names instead of short names. For example `v1.pod` is now `io.k8s.kubernetes.pkg.api.v1.Pod`. However this name is not convenient in python. But we cannot just shorten names because there would be duplicates. I've added a code to shorten model names as much as possible. So as long as there is no other model name ending with `v1.pod`, for example, we shorten `io.k8s.kubernetes.pkg.api.v1.Pod` to `v1.pod`.

PR #170, opened 2017-04-04
Update generated client

### LINKED ISSUE #167: Update generated client
I would like to request that you update the generated client for API v1.6.0

mbohlool:
Updating the client after kubernetes 1.6 release. The update process was not trivial as there were breaking changes in OpenAPI spec of kubernetes 1.6 release.

fixes #167 #169


scripts/constants.py
```diff
@@ -15,15 +15,15 @@
 import sys
 
 # Kubernetes branch to get the OpenAPI spec from.
-KUBERNETES_BRANCH = "release-1.5"
+KUBERNETES_BRANCH = "b0b7a323cc5a4a2019b2e9520c21c7830b7f708e"
```
dims: Isn't there a release-1.6 branch now?

dims:
Quick question inline, LGTM 👍

scripts/constants.py
mbohlool: I don't have a strong opinion about this yet (so I am changing it back to release-1.6) but need need to discuss this. A branch may change (patches) and we should somehow keep track of which spec we created client from (until we have a consistent versioning in spec itself).

mbohlool:
Answered. Thanks.

PR #180, opened 2017-04-18
add support for attach api call using websockets.

skuda:
Hi,

I needed to use /attach api call, this method exposes the same problem as /exec so I just reused the current solution using websockets.


skuda:
Check CLA

mbohlool:
LGTM. Thanks for the contribution.

mbohlool:
nit: can you squash all commits into one please?

skuda:
Done, sorry, someway 4 spaces where added to an empty line and I fixed it using the webinterface.

PR #183, opened 2017-04-19
BUGFIX: token file should not be decoded

### LINKED ISSUE #182: Contents of service account token mounted in POD is being base64 encoded
A project I have is using the client-python and using the mounted tokens wasn't working. After some investigation I figured out that the token file contents is being encoded in base64. From my understanding this content should not be encoded after read, the token should be ready to use.

The problem is in the [KubeConfigLoader#_load_user_token](https://github.com/kubernetes-incubator/client-python/blob/master/kubernetes/config/kube_config.py#L164-L170) which uses the class `FileOrData#as_data` and encode the contents of the file after read.

I have created a small project that illustrates the problem. I create a [ConfigMap](https://github.com/rafaelcaricio/kube-pods-lister/blob/0.1.0/deployment.yaml#L32-L57) which uses the service account token from the filesystem in a custom `kubeconfig`. Then I use the function [`new_client_from_config`](https://github.com/rafaelcaricio/kube-pods-lister/blob/0.1.0/pods_lister.py#L35) in my app passing this custom `kubeconfig`.

A work around I used for now is to split the `Bearer ` part of the header and decode the token and concat it back.
```python
authotization = config.api_key.get('authorization')
k, v = authorization.split(' ')
v = base64.decodebytes(v.encode())
request.headers['Authorization'] = "{} {}".format(k, v.decode())
```

Maybe I am missing the reason why the contents of the token file is being encoded in base64?

mbohlool:
Fixes #182

Also patch this to Release 1.0.


rafaelcaricio:
@mbohlool Thank you!

kubernetes/config/kube_config.py
```diff
@@ -62,11 +62,12 @@ class FileOrData(object):
      raw data and data field is base64 string."""
```
ravisantoshgudimetla: NIT: I may be wrong but shouldn't we update comment to reflect that raw data can be base64 encoded now?

kubernetes/config/kube_config_test.py
```diff
@@ -120,6 +126,13 @@ def test_data_given_file(self):
         t = FileOrData(obj=obj, file_key_name=TEST_FILE_KEY)
         self.assertEqual(TEST_DATA_BASE64, t.as_data())
 
+    def test_data_given_file_no_base64(self):
+        obj = {
+            TEST_FILE_KEY: self._create_temp_file(content=TEST_DATA)}
+        t = FileOrData(obj=obj, file_key_name=TEST_FILE_KEY,
+                       base64_file_content=False)
+        self.assertEqual(TEST_DATA, t.as_data())
+
```
ravisantoshgudimetla: So, I am not able to find test case for base64_file_content=True. I might have missed something here as I am new to the codebase. One more thing, what's the difference between this test-case and the one above, I think, since base64_file_content=False by default, the -ve scenario may be same as the old one. Please correct me, if I am wrong.

mbohlool: `base64_file_content` default value is `True`. So the test case above is testing `True` case and this one is testing the `False` case.

kubernetes/config/kube_config.py
```diff
@@ -62,11 +62,12 @@ class FileOrData(object):
...
 
     def __init__(self, obj, file_key_name, data_key_name=None,
-                 file_base_path=""):
+                 file_base_path="", base64_file_content=True):
```
mbohlool: @ravisantoshgudimetla this is where I set the default value to `True`.

ravisantoshgudimetla: I misread this part. Thanks for correcting.

kubernetes/config/kube_config_test.py
ravisantoshgudimetla: Apologies, I wrote my other comment in haste. In fact, I was thinking the default value is False.

kubernetes/config/kube_config.py
mbohlool: Added a comment explaining the flag. Thanks.

PR #186, opened 2017-04-20
Port Kubernetes cherry-pick script

mbohlool:
Kubernetes main repo has a cherry pick script that is useful for client-python, now that we have more than one release branch. I've generalized the script a bit and port it for our use.


zmerlynn:
Actually .. can you squash and get rid of those weird master commits? I just noticed the commit count.

(You should be rebasing the PR, it looks like you're trying to merge a merge commit here.)

mbohlool:
@zmerlynn Those merge commits created by github UI directly. When I merge pull request they should disappear (I think). I will check that before merge.

mbohlool:
@zmerlynn I assumed that those merge commit will go away but look like at least one of them stayed when I merged. I should somehow disable UI for that. Thanks for pointing it out.

PR #189, opened 2017-04-24
Automated cherry pick of #183

mbohlool:
Cherry pick of #183 on release-1.0.

#183: BUGFIX: token file should not be decoded


mbohlool:
original PR has LGTM. the diff is exactly the same.

mbohlool:
This is a backward incompatible change (people may fixed their token file to make it work). Reverting PR.

PR #202, opened 2017-04-26
Exclude urllib3 version 1.21 from depedencies as it breaks our code

mbohlool:
ref: https://github.com/kubernetes-incubator/client-python/issues/197

Look like the problem is going away after 1.21 because of this patch: https://github.com/shazow/urllib3/pull/1157

so we should be fine with versions > 1.21


mbohlool:
actually this should be a patch after being merged to master. closing this.

mbohlool:
@dims please review #203 and ignore this one. I will patch the PR to 1.0 branch when that is merged.

Issue #187, opened 2017-04-21
patch_namespaced_deployment gets 500 as response when using json-patch


aosterkamp:
When using a [RFC6902](https://tools.ietf.org/html/rfc6902)  formatted json-patch you get a 500 response.
The hack for the issue #93 introduced this bug. 
Are there any plans to get rid of the hack?

To repoduce the problem the following can be done:

create the deployment with kubectl 
```bash
kubectl run nginx --image=nginx:1.11.12
```

try to patch it
```python
from kubernetes import client, config
config.load_kube_config()
client.configuration.debug = True
v1 = client.ExtensionsV1beta1Api()
patch = [{"op": "replace", "value": "nginx:1.11.13", "path": "/spec/template/spec/containers/0/image"}]
v1.patch_namespaced_deployment(name='nginx', namespace='default', body=patch)
```

log output
```
DEBUG Starting new HTTPS connection (1): xxx
send: b'PATCH /apis/extensions/v1beta1/namespaces/default/deployments/nginx HTTP/1.1\r\nHost: xxx\r\nAccept-Encoding: identity\r\nContent-Length: 95\r\nAccept: application/json\r\nContent-Type: application/strategic-merge-patch+json\r\nUser-Agent: Swagger-Codegen/1.0.0-alpha/python\r\nauthorization: Bearer xxx\r\n\r\n'
send: b'[{"op": "replace", "value": "nginx:1.11.13", "path": "/spec/template/spec/containers/0/image"}]'
reply: 'HTTP/1.1 500 Internal Server Error\r\n'
DEBUG https://xxx:443 "PATCH /apis/extensions/v1beta1/namespaces/default/deployments/nginx HTTP/1.1" 500 114
header: Content-Type header: Date header: Content-Length Traceback (most recent call last):
  File "api_test.py", line 25, in <module>
    main()
  File "api_test.py", line 21, in main
    print(v1.patch_namespaced_deployment(name='nginx', namespace='default', body=patch_test))
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/apis/extensions_v1beta1_api.py", line 5400, in patch_namespaced_deployment
    (data) = self.patch_namespaced_deployment_with_http_info(name, namespace, body, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/apis/extensions_v1beta1_api.py", line 5496, in patch_namespaced_deployment_with_http_info
    collection_formats=collection_formats)
[...]

PR #205, opened 2017-04-28
fix for #187 failure when using RFC 6902 json-patch operation

bsherman:
My solution to the issue assumes a body containing a JSON array to be a RFC 6902 'json-patch' operation because this RFC requires an array of operations.  The existing code already assumes the use of 'strategic-merge-patch' (Kubernetes preferred) instead of the RFC 7386 'merge-patch', so I left that behavior as is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v2.0.0b2
+- Bugfix: support RFC6902 'json-patch' operations #187
+
 # v1.0.1
 - Bugfix: urllib3 1.21 fails tests, Excluding version 1.21 from dependencies #197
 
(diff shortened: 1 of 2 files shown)


bsherman:
CLA completed.

dims:
LGTM 👍

mbohlool:
LGTM. Thanks. nit: Can you add an entry in CHANGELOG.md?

bsherman:
I'm happy to add the changelog entry. It seems this is a bugfix which would also be good in the 1.0 branch for kubernetes 1.5 users. Would you prefer a separate PR for that?

mbohlool:
There is no need for a separate PR for that. We can cherry-pick this PR into 1.0 release after it got merged.

bsherman:
Great, thank you!

mbohlool:
I just realize this PR is sent against release-2.0. I will move it to master. We should always merge these into master first.

PR #208, opened 2017-05-03
Update update-client.sh

mbohlool:
Generator script post-processing if failing if development status has `/` in it. this is true for stable releases.


mbohlool:
Thanks.

PR #215, opened 2017-05-09
Move config and watch into a base folder to prepare for multi-repo client

mbohlool:
We need to separate our base and generated code so other users can create their own client and use our base utilities. This is a required restructuring explained here:

https://github.com/kubernetes-client/community/blob/master/design-docs/clients-library-structure.md

`config` and `watch` package has been moved into a `base` package as well as `api_client` `rest` and `configuration` classes. Symlinks created to make everything work as before.


mbohlool:
@dims this is ready to review.

dims:
Looks like just things moving around :) /lgtm

mbohlool:
Thanks

PR #222, opened 2017-05-19
Add kubernetes-client/python-base as submodule to this repo

mbohlool:
We need to restructure python client according to https://github.com/kubernetes-client/community/blob/master/design-docs/clients-library-structure.md

for that purpose, base folder is moved to the other repo.


mbohlool:
I see `python setup.py install` is failing on master, so it is not related to this PR. I will look at it. Would you file an issue for it?

PR #230, opened 2017-05-22
Fix TPRs security issue

### LINKED ISSUE #227: Authentication issue for new list third party resource endpoint
Hi @mbohlool I am trying the new tpr resources endpoints and it seems to work fine if I am outside a cluster, but when I run it in a cluster it appears that my call to `list_third_party_resource` is getting a 401 as there is no auth passed with the request. This is in contrast to the `replace_third_party_resource` call which seems to work fine.

I did notice that here: https://github.com/kubernetes-incubator/client-python/blob/master/kubernetes/client/apis/third_party_resources_api.py#L487, there is no 'Bearer' in the auth settings like there appears to be in the replace call I do elsewhere.

Is there some configuration on the client generation that is missing?

Thanks.

mbohlool:
The TPR spec file defined security tags for TPR operations incorrectly. All of our operations inherit security from the root one, thus this file should not specify any security tag.

fixes #227


pvdvreede:
👍  Thanks for the quick turnaround! Just used your branch from this PR and it seems to have fixed my issue. :)

PR #238, opened 2017-06-02
Read setup.py dependencies from requeriments.txt

### LINKED ISSUE #136: Add check to make sure dependencies in requirements.txt and setup.py are in sync
Adding a dependency to one but not the other creates problems like #131. We should add a check script for this. It is a plus if the check scripts checks for all requirements in https://packaging.python.org/requirements/

pokoli:
Fixes #136

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,9 @@ DEVELOPMENT_STATUS = "3 - Alpha"
 with open('requirements.txt') as f:
     REQUIRES = f.readlines()
 
+with open('test-requirements.txt') as f:
+    TESTS_REQUIRES = f.readlines()
+
 setup(
     name=PACKAGE_NAME,
     version=CLIENT_VERSION,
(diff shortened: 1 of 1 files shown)


pokoli:
I also added test-requirments as test_requires on setup.py, which will make the test requeriments installed when running `python setup.py test` for me is the expected behaviour, but let me know if you have some remarks with it.

mbohlool:
Thanks for the PR. It looks good. The new `REQUIRES` list, however, is a little different after this PR (It has all versions in it as well as commented license). Do we know that is fine? I think there were a guideline somewhere to not include versions in `REQUIRES` field.

pokoli:
@mbohlool for me the important part is that requirements and setup.py must contain the same package dependencies, so installing with requeriments and with setup.py produces the exact same behaviour. Having said that, if you do not want the versions in the setup.py I think we should not include on requirements.txt, but maybe this out of the scope of this PR. (I can create a new one if you prefer).

mbohlool:
More information here: https://packaging.python.org/requirements/

I think we can go ahead with this PR.

PR #239, opened 2017-06-03
Update create_deployment notebook

djkonro:


examples/notebooks/create_deployment.ipynb
```diff
@@ -12,48 +23,288 @@
...
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {
     "collapsed": true
    },
    "outputs": [],
    "source": [
-    "client.Configuration().host=\"http://localhost:8080\""
+    "config.load_kube_config()\n",
+    "extensions = client.ExtensionsV1beta1Api()"
```
pokoli: I also changed this line in:

https://github.com/kubernetes-incubator/client-python/pull/242/files

I think it's better to load incluster config as the examples are intended to be run from inside a kubernetes cluster.

sebgoa: @pokoli do you mind if we merge this PR and close yours. This addresses more issues in the notebook. Once merged we can discuss more.

The general issue with these notebooks is to figure out a way to quickly test them.

I am tempted to add two steps in the notebooks with a try loop that loads in cluster and loads out of cluster.

sebgoa: @mbohlool I am +1 on this, we need to improve things but it is a first step

pokoli: Makes sense to merge my PR into this. When I first created my PR I wasn't aware of this. So please go ahead.

examples/notebooks/create_deployment.ipynb
```diff
@@ -12,48 +23,288 @@
...
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {
     "collapsed": true
    },
    "outputs": [],
    "source": [
-    "extensions = client.ExtensionsV1beta1Api()"
+    "deployment = client.models.extensions_v1beta1_deployment.ExtensionsV1beta1Deployment()"
```
sebgoa: @djkonro actually you should change this to `deployment = client.ExtensionsV1beta1Deployment()`

examples/notebooks/create_deployment.ipynb
```diff
@@ -12,48 +23,288 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "#extensions.delete_namespaced_deployment(name='nginx-deployment', namespace='default', body=client.V1DeleteOptions(),grace_period_seconds = 56, propagation_policy =  \"Background\")"
```
mbohlool: Search issues for this. I think somebody had the same problem recently and we discussed it there. Briefly, I think there is a bug upstream (in k8s itself) and you need to delete childs one by one.

sebgoa: I don't think it is a bug. The kubectl delete deployment has a reapper that cascades deletion to pods, but the delete deployment API call itself does not do that.

@djkonro you can list pods with the label of the deployment and delete them based on that.

mbohlool: ref: https://github.com/kubernetes-incubator/client-python/issues/162

Look like with 2.0.0, you should be able to do cascade delete?

djkonro: @mbohlool, I have tried various options but it does not seem to delete dependents, even when I set `propagationPolicy` to `Background` or `Foreground`, like in the function call below.
`extensions.delete_namespaced_deployment(name='nginx-deployment', namespace='default', body=client.V1DeleteOptions(),grace_period_seconds = 56, propagation_policy =  \"Background\")`

djkonro: Even the `grace_period_second` does not seem to take effect.

djkonro: @sebgoa The pods are a part of a ReplicaSet, so they automatically get recreated when I delete them. I usually just tend to deleted the ReplicaSet in order to delete the pods, but I am looking for a better solution where I can delete them directly from the deployment.

djkonro: Thanks to @vascofg, updating my version of Kubernetes solved my cascading deletion issue.

djkonro:
@sebgoa, I have added RollingUpdate and RollBack to the deployment notebook, as seen in https://kubernetes.io/docs/concepts/workloads/controllers/deployment/. Please let me know what you think.

sebgoa:
@djkonro there is no rolling update or rollback, I think you forgot to push the commit.

djkonro:
@sebgoa  Thanks for noticing the omission, sorry I did not realize the commit did not contain rolling update or rollback.

sebgoa:
/lgtm

@mbohlool

PR #242, opened 2017-06-05
Use load_incluster_config in jupyther notebook examples

### LINKED ISSUE #219: Fix generated examples
Generated examples refer the wrong way to load configs and setup client and this will result in confusions such as #217. Fix the generated examples at the source.

pokoli:
Closes #219


sebgoa:
I'd rather merge #239 and close this one. #239 address is a complete notebook.

We need to revisit all notebooks with a try/except maybe to load in both cases.

Issue #262, opened 2017-06-20
SSL Verification error when trying to stream logs from pod


jonastl:
The following snippet when run in a cluster pod generates the error: `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)`

```
from kubernetes import client, config

try:
    config.load_incluster_config()
except:
    config.load_kube_config()

v1 = client.CoreV1Api()

# SSL verification fails
c = v1.connect_get_namespaced_pod_attach(
    name='kube-addon-manager-minikube', namespace='kube-system',
    container='kube-addon-manager', stdin=False, tty=False,
    stdout=True, stderr=False, _preload_content=False
)
```

Installed packages:
```
$ pip3 freeze
certifi==2017.4.17
httplib2==0.10.3
ipaddress==1.0.18
kubernetes==2.0.0
oauth2client==4.1.1
pyasn1==0.2.3
pyasn1-modules==0.0.9
python-dateutil==2.6.0
PyYAML==3.12
rsa==3.4.2
six==1.10.0
urllib3==1.21.1
websocket-client==0.42.1
```

jonastl:
In fact, the error happens when run outside a pod as well, so it's a generic problem.

jonastl:
Closing this since a fix has been merged incorporated into v3.0.0

PR #263, opened 2017-06-20
Fix invalid CA cert used by websocket lib (#262)

jonastl:
The following change in upstream websocket-client changed the ssl
options interface by renaming the configuration key from "ca_certs" to
"ca_cert": https://goo.gl/cxSrXt

Since version 0.41.0 of the websocket library it was no longer possible
to stream the pod logs via the kubernetes python library. Reason being
that the websocket client library (project dependency) was falling back
on the platform default SSL bundle, which does not include the required
kubernetes ca cert. Consequently python's ssl library raised a
CERTIFICATE_VERIFY_FAILED error when trying to talk to the API.

This commit adds handling for both cases, meaning now both the older
configuration key-name and the new are supported, ensuring the proper
cert propagates to the websocket client. Thus avoiding requiring a
lower bound version bump on the required websocket client library
version. The associated unit test also verifies that the cert reaches
python's native SSL code as intended, to guard against future
regressions of this sort.

### DIFF
--- a/kubernetes/client/ws_client_test.py
+++ b/kubernetes/client/ws_client_test.py
@@ -12,9 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ssl
 import unittest
+import websocket
 
-from .ws_client import get_websocket_url
+from .ws_client import get_websocket_url, websocket_call
+from kubernetes.client import configuration
+from unittest.mock import patch
 
 
 class WSClientTest(unittest.TestCase):
@@ -31,6 +35,24 @@ class WSClientTest(unittest.TestCase):
                 ('https://api.domain.com/', 'wss://api.domain.com/'),
                 ]:
             self.assertEqual(get_websocket_url(url), ws_url)
+
+    @patch.object(websocket._http, '_can_use_sni')
+    @patch.object(websocket._http, '_open_socket')
+    @patch.object(ssl.SSLContext, 'load_verify_locations')
+    def test_use_of_correct_ca_cert(self, load_verify_locations, *ignored):
+        # arrange
+        expected = '/expected-ca.pem'
+        configuration.ssl_ca_cert = expected
+        load_verify_locations.side_effect = Exception("need not go further")
+
+        # act
+        with self.assertRaises(Exception):
+            websocket_call(configuration, 'https://localhost',
+                           {}, None, True, {})
+
+        # assert
+        actual = load_verify_locations.call_args[1]['cafile']
+        self.assertEqual(expected, actual)
 
 
 if __name__ == '__main__':
--- a/kubernetes/client/ws_client.py
+++ b/kubernetes/client/ws_client.py
@@ -16,7 +16,9 @@ import select
 import certifi
 import time
 import collections
+from distutils.version import StrictVersion
 from websocket import WebSocket, ABNF, enableTrace
+from websocket import __version__ as websocket_version
 import six
 import ssl
 from six.moves.urllib.parse import urlencode, quote_plus, urlparse, urlunparse
@@ -51,8 +53,11 @@ class WSClient:
                 'cert_reqs': ssl.CERT_REQUIRED,
                 'keyfile': configuration.key_file,
                 'certfile': configuration.cert_file,
-                'ca_certs': configuration.ssl_ca_cert or certifi.where(),
+                'ca_cert': configuration.ssl_ca_cert or certifi.where(),
             }
+            if StrictVersion(websocket_version) < StrictVersion('0.41.0'):
+                ssl_opts['ca_certs'] = ssl_opts.pop('ca_cert')
+
             if configuration.assert_hostname is not None:
                 ssl_opts['check_hostname'] = configuration.assert_hostname
         else:


kubernetes/client/ws_client_test.py
```diff
@@ -32,6 +36,24 @@ def test_websocket_client(self):
...
+    @patch.object(websocket._http, '_open_socket')
+    @patch.object(ssl.SSLContext, 'load_verify_locations')
+    def test_use_of_correct_ca_cert(self, load_verify_locations, *ignored):
+        # arrange
+        expected = '/expected-ca.pem'
+        configuration.ssl_ca_cert = expected
+        load_verify_locations.side_effect = Exception("need not go further")
+
+        # act
+        with self.assertRaises(Exception):
```
pokoli: Why you need to raise and error here?

If you do not return an exception as side_effect, but return something that the code will understand. You can remove the assertRaises

jonastl:
cncf cla ping

kubernetes/client/ws_client_test.py
jonastl: It's just a trick to unwind the stack (force an early return deeper in the call graph). 

The SSL bundle loading in python happens deep within the websocket library, and I really _didn't_ want to test that code's implementation, but could think of no other way to assert that the SSL configuration made in the kube client gets propagated properly to python-ssl.

That propagation is at the heart of this bug, and testing just the interface between the kube client and websocket library would not provide any increased certainty, and would not reveal this bug since the bug is an _interface mismatch_ between this project and its immediate dependency. I.e. a breaking change in the dependent project.

Since the python SSL library is well documented and understood, testing what data eventually propagates to _it_ should increase confidence that the intention underpinning dependent functions invocations from this project yield the expected effect.

The other option would have been to simply bump the requirements spec to websocket-client >= 0.41.0 and just change the configuration key (one character) in the code, but I felt uncertain of what implications bumping lower bounds for the package requirement would entail. Some teams support a wide range of package versions for various reasons, while other projects pin versions hard (to exact revisions). Since this project uses version ranges I assumed there was probably a good reason for that, a reason I however am unaware of. Thus sticking with existing requirements and compensating for change in interface by conditional branching.

Does that answer the question sufficiently, and provide enough context to see the reasoning?

pokoli: Yes, I understand the problem. But what i dont understand is why you return an exception as side_efect. 

Have you tried using None as side efect? 

What if the code raises another exception due to some missconfiguration? This test will silently fail...

jonastl: If the code raises another exception, then it won't have reached the ssl context configuration step, and consequently the assertion will then fail. As is intended. I.e. The proper call to the ssl library wasn't made.

If however the code execution raises no exception until the step we're interested in, the call arguments are recorded in the mock, and we then raise an explicit exception to unwind back to the original call site in the unit test function. That's what I tried to convey with the exception message "go no further"; meaning don't execute a single additional statement in the library, since we have the info needed to determine expexted behavior at that point.

Not raising an exception would cause execution to continue in the websocket library, causing it to enter an event loop expecting socket related events, and hang the test.

In short, returning None would not cause a stack unwind (escape back to the original call site)

mbohlool:
> So instead of checking versions there is another option to limit the version of websocker-client in https://github.com/kubernetes-incubator/client-python/blob/master/requirements.txt#L9

I was thinking the same. Not sure how restrictive it would be to require a new version. Generally I am leaning toward updating the version instead of supporting more than one version but here I didn't look into the issue deep enough to have an opinion. Will do.

kubernetes/client/ws_client_test.py
pokoli: The method that causes the websocket library to enter the event loop should be probably mocked also. IIUC, then it won't be needed to raise an exception.

jonastl:
I'm all for _pinning_ explicit version dependencies for private / company internal projects as it's easier to resolve _transient dependency hell_ for special cases. For widely shared libraries however, I felt that might cause a problem since as far as I know, python doesn't support different versions of packages in the same address space.

So if a user has built some app with dependencies of which one of them has a dependency on a specific version of A (websocket-client), and then adds the kubernetes package, which if we were to pin the websocket dependency to a different version (A') , problem would ensue for setuptools.

jonastl:
In this case, adding a >= 0.41.0 version requirement would prevent users from utilizing any library not yet capable of handling version 0.41.0 and higher.

This however might not be a problem in practice, since public libs seldom fix the upper bound, and instead try to encourage upgrades of dependencies to latest.

jonastl:
To progress this I'm fine with bumping the lower bound to 0.41.0 and remove the branch, making this a 1 character change. Are you guys fine with that?

The remaining question then is the unit test. The test makes a behavioural verification past a transitive dependency, in effect testing the implementation of _another_ project. That feels wrong, but as explained above, the reason for doing this is because of the nature of the bug.

To me it seems the remaining question then is one of policy; should unit tests verify collaborator behaviours or not. 

If the consensus is to not test past collaborator interfaces, then we have an observed error kind we know the reason for, know how to verify, but have _chosen not to test_ because of strict adherence to the mentioned policy. Otherwise if strict policy adherence isn't absolute for all situations, then some kind of testing like what's proposed would seem needed, and the question then remaining is if there's a more elegant way to test this than what's proposed in the PR.

So..
1. Are people good with requiring a version >= 0.41.0 of the websocket library?
2. Should anything be done about the unit test, and if so what improvement suggestions are there?

kubernetes/client/ws_client_test.py
jonastl: Works until the websocket library elects to change their _private_ implementation and branch to such behavior somewhere else. Point being that stubbing implementation details in external code is always brittle and the best one can do is minimize the number of instructions executed in such a "black box". The chosen option was the least risky I could find, as it minimizes the execution in (by this project) uncontrolled code, to the absolute minimal extent required to assert the side-effect it requires.

mbohlool:
I think with version check we can support more users so I am in favor of keeping it. Only one comment regarding version check. I will let you guys decide on the other issue @pokoli raised.

jonastl:
The error has now been detected and fixed upstream in version 0.43.0 and later.
I'm closing this PR as the way to tackle the issue has changed in the light of the upstream evolution. A different PR (#270) has been opened instead, which just blacklists the bad WS client versions.

PR #265, opened 2017-06-21
Parametrize resource path with version part of apiVersion

### LINKED ISSUE #256: Bug in ThirdPartyResourcesApi (hardcoded resource path)
This project is awesome and I was delighted to see #216 get merged.

There is a bug when constructing the resource paths, however: For instance, in the class ThirdPartyResourcesApi, on line 126 (+/-):

resource_path = '/apis/{fqdn}/v1/namespaces/{namespace}/{resource}'.replace('{format}', 'json')

This does not work for third-party resources which are not of version 'v1' exactly. So, for example, the apiVersion "monitoring.coreos.com/v1alpha1" fails with a 404.

I suggest taking in the version as a formal parameter in the relevant methods and injecting it in the resource path.

ebbeelsborg:
Fixes #256 .

Also swaps the order of parameters "name" and "namespace" where relevant to match the standard API.

Note: This change will, intentionally, break consumers of the API that were relying on the version to always be "v1" without specifying it. Such consumers will now have to specify "v1" explicitly.


mbohlool:
OpenAPI spec has unrelated changes. I suggest you run update client script once to pull any changes upstream, then commit these changes, then add manual changes in a separate commit and then the third commit updating client again to apply effects on manual changes to the generated client.

ebbeelsborg:
@codevulture , Sure, here's an example: https://raw.githubusercontent.com/kubernetes-incubator/client-python/master/kubernetes/client/apis/core_v1_api.py (search for patch_namespaced_pod e.g.). Pardon the raw link, GitHub won't show the usual one right now. I realize that the ordering for declaring TPR resources is not <name, namespace>, but rather <namespace, name>, but for many of the standard APIs (Core V1, Extensions, Batch etc.) it seems to be the case that the ordering is <name, namespace>. And the [Go APIs](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/pod.go) are a bit different as well. I don't feel strongly about it all, so if you wish, I can revert that part.

ebbeelsborg:
@mbohlool , Thanks. I tried to do what you suggested.

@codevulture , this second reincarnation is without swapping the name and namespace parameter ordering. If it is desired, I can put it back in, just let me know.

mbohlool:
@ebbeelsborg thanks. Almost there (commit-wise). Can you now use `git rebase -i HEAD~7` and squash relevant commits together to make only three commits in my previous comment.

codevulture:
@ebbeelsborg : Thanks for your work, though it is a matter of choice. i wanted to understand the reason of change which you explained well.

ebbeelsborg:
@mbohlool , @codevulture , let me know if I need to do something more/different.

mbohlool:
Thanks @ebbeelsborg, it looks good. So to be clear, a `name` parameter was missing in the TPR spec and you added it, right? I liked reversing order of `namespace` and `name` but don't have strong opinion either way, let me know if you want me to merge it unchanged or you want to change the order before merge.

ebbeelsborg:
@mbohlool , Correct. A `name` parameter was missing for the `update` operation. I would like the PR merged as-is. Thanks.

Regarding the order of parameters `name` and `namespace`, I prefer to have `name` first. It would be a breaking change, but it seems that there aren't too many users of the TPR API yet so probably that's okay (and better now than later). I'll make a new PR for that, okay?

(I haven't tested it, but I suspect that the `delete` operation will need a similar make-over.)

mbohlool:
Sure, will merge this one. Thanks.

PR #267, opened 2017-06-23
Swap order of parameters 'name' and 'namespace' for TPRs

ebbeelsborg:
Change as agreed upon, see https://github.com/kubernetes-incubator/client-python/pull/265#issuecomment-310298050.


mbohlool:
LGTM. Thanks.

PR #268, opened 2017-06-23
Add example about node label management

flavio:
Add example covering node label management as requested [here](https://github.com/kubernetes-incubator/client-python/issues/240#issuecomment-310497983).

### DIFF
--- a/examples/manage_node_labels.py
+++ b/examples/manage_node_labels.py
@@ -1,0 +1,46 @@
+# Copyright 2016 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


examples/add-node-label.py
```diff
@@ -0,0 +1,20 @@
+from pprint import pprint
```
mbohlool: Add copyright header please.

examples/add-node-label.py
```diff
@@ -0,0 +1,20 @@
...
+from kubernetes import client, config
+
+config.load_kube_config()
+client.configuration.debug = True
```
mbohlool: We don't need this in the example.

examples/remove-node-label.py
```diff
@@ -0,0 +1,19 @@
+from pprint import pprint
```
mbohlool: We can add this to the previous example. right? in between we can also read the node and check to see if the label is there.

flavio:
I applied all the chances your proposed.

mbohlool:
I think it is good to merge. Please fix pep8 check, more info here: https://travis-ci.org/kubernetes-incubator/client-python/jobs/249944871

Thanks.

PR #270, opened 2017-06-26
Restrict supported websocket client versions #262

jonastl:
There are a couple versions which won't work with this library,
specifically 0.41.1 - 0.42.0

Those versions prevent the kube CA certificate from being propagated to
python's SSL Context, which results in a CERTIFICATE_VERIFY_FAILED error
being raised when a user tries to stream logs from a Pod using the
kubernetes python library.

This change explicitly blacklists the problematic range of bad websocket
client library versions, so that users don't experience odd failures if
they happen to have the affected versions installed when they install
this project's library.

The restriction should provide the least amount of forced dependecy
updates, affecting only those that would have been affected by the bug.

Bug appeared in 0.41.1: https://goo.gl/cxSrXt
Was fixed in 0.43.0: https://goo.gl/EmJcF8

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,4 @@ urllib3>=1.19.1,!=1.21  # MIT
 pyyaml>=3.12  # MIT
 oauth2client>=4.0.0  # Apache-2.0
 ipaddress>=1.0.17  # PSF
-websocket-client>=0.32.0 # LGPLv2+
+websocket-client>=0.32.0,!=0.41.*,!=0.42.* # LGPLv2+


mbohlool:
My e2e test failed locally. The websocket version was 0.44.0 and when I downgraded it to 0.40.0 it worked.

jonastl:
Interesting, I only ran the python unit tests, not the full e2e test.
It seems there may be larger issues than anticipated upgrading this library.

Perhaps the best route for now is to just upper bound the requirement as you guys discussed in the earlier ticket, to <=0.40.0.

I'll close this one as it requires more investigation that I can afford atm. At my company we've just frozen the dependency to 040.0, but that option doesn't scale for a widely used library, as has been previously discussed.

mbohlool:
I am not sure why it fails for me locally (osx). It passes on Travis. I will try to find the issue next week. We may still be able to merge this.

jonastl:
Ok, I'll reopen in case your investigation of the failure turns out to only be an instance of  _doesn't work on my machine_ (false negative)

PR #281, opened 2017-07-05
 Add notebook on how to watch changes to an object 

djkonro:


djkonro:
Work in process (WIP)

sebgoa:
Sorry for the delay, you should add more information about what a watch is. Which resources can you watch ? etc.

Also the main issue here is that you only receive a fix amount of events and then exit. We need to figure out how to keep the stream open so that new events get received. It should not exit when the counter is 0.

djkonro:
@sebgoa I have updated the notebook.

sebgoa:
ok approved, a bit minimalist but ok.

/lgtm

PR #282, opened 2017-07-05
Use py.test instead of nosetests for better problem reporting

dims:
nosetests output is not very useful for diagnosing issues when
running in travis. Let us switch to py.test

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,5 +1,6 @@
 coverage>=4.0.3
 nose>=1.3.7
+pytest
 pluggy>=0.3.1
 py>=1.4.31
 randomize>=0.13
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -22,12 +22,12 @@ commands =
 [testenv:py27-functional]
 commands =
    python -V
-   {toxinidir}/scripts/kube-init.sh nosetests -v []
```
mbohlool: Why only these two e2e tests? what about py36-functional and non-e2e tests? can we use py.test for those too?

dims: Switched them all except for coverage. Let's wait on that one to see if it runs into trouble

test-requirements.txt
```diff
@@ -1,5 +1,6 @@
 coverage>=4.0.3
 nose>=1.3.7
```
mbohlool: Do we need this anymore?

dims: yes, testenv:coverage still uses both coverage and nose.

mbohlool: OK. this looks good, can you rebase your branch so I can merge it? (the branch is out of date).

PR #283, opened 2017-07-05
Support both TPR and CRD with custom objects spec

### LINKED ISSUE #251: Migrate ThirdPartyResources to CustomResourceDefinitions
Custom Resource Definitions (CRDs) are successor of TPRs. We should add support for those in next version of kubernetes.

- Feature issue: https://github.com/kubernetes/features/issues/95 
- Design proposal: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/thirdpartyresources.md
- CRD documentation: https://github.com/kubernetes/kubernetes.github.io/blob/release-1.7/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions.md
- TPR deprecation notice and migration path: https://github.com/kubernetes/kubernetes.github.io/blob/release-1.7/docs/tasks/access-kubernetes-api/extend-api-third-party-resource.md 
- issue about accessing the apiextensions-apiserver client: https://github.com/kubernetes/kubernetes/issues/46702
- client-go example for CRDs: https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apiextensions-apiserver/examples/client-go

### LINKED ISSUE #280: ThirdPartyResourcesApi.delete_third_party_resource()
I've been trying to use the ThirdPartyResourcesApi class to manage instances of our third party resources, and it all seems to work nicely except for the delete_third_party_resource() function.  Here, the code (kubernetes/client/apis/third_party_resources_api.py) has it defined as: `def delete_third_party_resource(self, body, **kwargs)`, with "body" described as a "V1DeleteOptions" type.

It is not clear how to specify the identify of the resource to be deleted.  I would expect parameters like you have with get or update - i.e., name, namespace, fqdn, version, and resource.

Is this a problem?  Or is the identify of the object to be deleted to be descrbed in the 'body' parameter somehow?  I've experimented with it some, but without success.  Also, reading the code, it looks like it is expecting those parameters to be defined, though they are not.

I installed the library from the source with setup.py, using the commit made on June 25 (7650c9a)

mbohlool:
Fixes #251 

should also fix #280


mbohlool:
cc @djmckinley

mbohlool:
cc @enisoc

scripts/custom_objects_spec.json
```diff
@@ -0,0 +1,554 @@
...
+        "type": "string",
+        "description": "If 'true', then the output is pretty printed.",
+        "name": "pretty",
+        "in": "query"
+      },
+      {
+        "name": "group",
+        "in": "path",
+        "required": true,
+        "description": "The Custom object's group name",
```
enisoc: Some of the usages of "custom object" in this spec make more sense to me as "custom resource". We intentionally added the "D" in "CRD" to distinguish the custom resource from the thing that installs it (the custom resource definition).

The way I think of it is: the CRD installs a custom resource, which is the location in the API where you can store custom objects.

scripts/custom_objects_spec.json
```diff
@@ -0,0 +1,554 @@
...
+      },
+      {
+        "name": "version",
+        "in": "path",
+        "required": true,
+        "description": "The Custom object's version",
+        "type": "string"
+      },
+      {
+        "name": "kind",
```
mbohlool: Yes, this is the one goes into URL. Do you have any suggestion for its name that make sense in both CRD and TPR world?

scripts/custom_objects_spec.json
```diff
@@ -0,0 +1,554 @@
...
+    "get": {
+      "operationId": "listClusterCustomObject",
+      "description": "list or watch cluster scoped custom objects",
+      "tags": [
+        "custom_objects"
+      ],
+      "consumes": [
+        "*/*"
+      ],
+      "produces": [
```
mbohlool: Make sense. Will remove yaml and proto.

mbohlool:
@enisoc updated with your comments. Can you take another look?

djmckinley:
This looks good as far as fixing #280.

I have a couple of questions about the "list" functions, though.

The previous third_party_resources_api.py module appeared to list third party resources across all namespaces.  Now, with the custom_objects_api.py module, the list functions for namespaced CRs seem to support listing namespaced custom objects only in a single namespace.  Is there a reason there is not a "list_namespaced_custom_objects_for_all_namespaces" function like there are for the various core resources?  (For example, does "list_cluster_custom_objects" perhaps do that if the object kind it identifies is a namespaced object?)

Second, is there a reason that the list functions do not have the optional parameters, "label_selector" and "resource_generation" that are present on the list functions for core objects?

Both the ability to list objects in a single namespace, or across all namespaces, and the ability to filter by label and resource_generation would be very useful for my use case.

mbohlool:
@djmckinley both good points. On your first point, `list_cluster_custom_objects` should do the trick, if you can give it a try and see if that is the case, it would be awesome. The naming prefered CRDs as they are the way to go and TPRs will be removed in next version of kubernetes (1.8).

On your second point, you are right, I will add those optional parameters to all list operations.

enisoc:
LGTM other than the additional list options mentioned above.

I verified that you can list namespaced custom objects across all namespaces by hitting the non-namespaced resource path:

```console
$ kubectl get --raw /apis/stable.example.com/v1/crontabs/ | json_pp
{
   "kind" : "CronTabList",
   "metadata" : {
      "resourceVersion" : "1805",
      "selfLink" : "/apis/stable.example.com/v1/crontabs/"
   }, 
   "items" : [
      {
         "spec" : {
            "image" : "old-image",
            "cronSpec" : "* * * * /5"
         }, 
         "apiVersion" : "stable.example.com/v1",
         "kind" : "CronTab",
         "metadata" : {
            "deletionGracePeriodSeconds" : null,
            "creationTimestamp" : "2017-07-06T18:52:03Z",
            "namespace" : "default",
            "uid" : "3385cb16-627c-11e7-9d68-a0481c9805b1",
            "deletionTimestamp" : null,
            "resourceVersion" : "1363",
            "clusterName" : "",
            "name" : "my-new-cron-object",
            "selfLink" : "/apis/stable.example.com/v1/namespaces/default/crontabs/my-new-cron-object"
         }
      },
      {
         "spec" : {
            "cronSpec" : "* * * * /5",
            "image" : "my-awesome-cron-image"
         },
         "apiVersion" : "stable.example.com/v1",
         "kind" : "CronTab",
         "metadata" : {
            "name" : "crontab-1",
            "clusterName" : "",
            "selfLink" : "/apis/stable.example.com/v1/namespaces/ns2/crontabs/crontab-1",
            "creationTimestamp" : "2017-07-06T18:56:00Z",
            "resourceVersion" : "1796",
            "deletionTimestamp" : null,
            "uid" : "c05f8fd5-627c-11e7-9d68-a0481c9805b1",
            "namespace" : "ns2",
            "deletionGracePeriodSeconds" : null
         }
      }
   ],
   "apiVersion" : "stable.example.com/v1"
}
```

mbohlool:
Thanks @enisoc and @djmckinley. I need to close this PR and send the PR agains kubernetes-client/gen repo (we moved to use that for generation).

mbohlool:
Re-opening this PR (to keep history of this change) with just updating the client after gen repo PR is merged. 
@djmckinley can you try your example again? it should work now.

ebbeelsborg:
@mbohlool Thanks for doing this; it's an important issue. There is one thing I can't get to work though.

**Main question**: How do I manage CRDs with client-python 3.0?

**Background**
I can manage the CRD *objects* with the new CustomObjectsApi, which works nicely. I can also still manage TPR definitions via the ExtensionsV1beta1Api, but that API seems to not work directly for CRD *definitions* (YAML with "Kind: CustomResourceDefinition").

Here is an example cluster response:

`HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"CustomResourceDefinition in version \"v1beta1\" cannot be handled as a ThirdPartyResource: no kind \"CustomResourceDefinition\" is registered for version \"apiextensions.k8s.io/v1beta1\"","reason":"BadRequest","code":400}
`

I'm calling the method "create_third_party_resource" YAML as its body on a client.ExtensionsV1beta1Api instance.

I'm on k8s 1.7.4.

**Concretely**:
Is the new ApiExtensionsV1beta1Api missing from the library, or am I supposed to transform (strip a few fields such as scope etc.) the CRD def. to a TPR def. before calling "create_third_party_resource"?

([Changeset](https://github.com/kubernetes-incubator/client-python/commit/917ec63d510154303f1508b042f9b76d4e981d93))

Thanks!

PR #288, opened 2017-07-07
Switch to minikube instead of localkube for e2e tests

mbohlool:
WIP

### DIFF
--- a/scripts/kube-init.sh
+++ b/scripts/kube-init.sh
@@ -70,6 +70,8 @@ done
 
 # kubectl commands are now able to interact with minikube cluster
 
+sed -i'' "s/127.0.0.1:8080/127.0.0.1:8443/" ~/.kube/config
+
 echo "Dump Kubernetes Objects..."
 cat ~/.kube/config
 kubectl get componentstatuses
--- a/scripts/kube-init.sh
+++ b/scripts/kube-init.sh
@@ -70,7 +70,7 @@ done
 
 # kubectl commands are now able to interact with minikube cluster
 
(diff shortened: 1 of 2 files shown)


dims:
Looks like you are pretty close `1 failed` :)

mbohlool:
Regardless of this works or not, I am worried about assert_hostname
failure. That means there is still a configuration that we get that error.

scripts/kube-init.sh
```diff
@@ -39,64 +39,41 @@ else
...
+
+curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 && chmod +x minikube
+
+export MINIKUBE_WANTUPDATENOTIFICATION=false
+export MINIKUBE_WANTREPORTERRORPROMPT=false
+export MINIKUBE_HOME=$HOME
+export CHANGE_MINIKUBE_NONE_USER=true
+mkdir $HOME/.kube || true
+touch $HOME/.kube/config
```
pokoli: Can we use a temp path for holding kube config? 

I remember losing all my .kube/config when running the e2e test on my local computer. 

As the config is exported latter, i do not see any drawback to use a non default hostpath.

mbohlool: This should not overwrite your existing config, but it is a good suggestion to be on the safe side.

PR #301, opened 2017-07-21
Add notebook on how to create and use a Secret

djkonro:


sebgoa:
I like this one a lot , but the exec call did not work for me.

We should check that it does work, it might be a client version issue

djkonro:
@sebgoa I updated it by setting `stderr=True` and `stdout=True` and it now works fine.

sebgoa:
ok to merge

/lgtm

PR #302, opened 2017-07-21
Add requests requirement for gce config refresher

mbohlool:
A requirement for PR in base repo:


mbohlool:
I am going to Merge this as it is only a requirement addition.

PR #310, opened 2017-07-28
fix the type of service_port.target_port

sctlee:
Hi there:
I fix a type error in service_port model. The type of target_port should be int, not str.


sctlee:
I have signed a CLA.

mbohlool:
@sctlee thanks for your contribution. Unfortunately the problem you are trying to fix is not that simple. Search for IntOrStr in the issues and you will find useful information. The file you are trying to change is auto-generated and cannot be changed manually. If you went through the issues of IntOrStr (the actual type of target port here) and had any suggestion/fix, I would love to hear that.

PR #318, opened 2017-08-10
Introducing __version__ variable

### LINKED ISSUE #317: No __version__ information
Sometimes it could be useful to get the version information programmatically, just like Django is doing it for example. There you can get the version string via `django.__version__`.

This is especially helpful, if your program requires `kubernetes` and is fine with either version, but needs to distinguish between the different APIs `kubernetes` offers.

brejoc:
With the gathered distribution information we are trying to fetch the version number for `kubernetes.__version__`.
Fixes #317

### DIFF
--- a/kubernetes/__init__.py
+++ b/kubernetes/__init__.py
@@ -14,7 +14,7 @@
 
 __project__ = 'kubernetes'
 # The version is auto-updated. Please do not edit.
-__version__ = None
+__version__ = "3.0.0-snapshot"
 
 import kubernetes.client
 import kubernetes.config
--- a/kubernetes/__init__.py
+++ b/kubernetes/__init__.py
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+__project__ = 'kubernetes'
+# The version is auto-updated. Please do not edit.
+__version__ = None
+
 import kubernetes.client
 import kubernetes.config
 import kubernetes.watch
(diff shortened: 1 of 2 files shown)


brejoc:
@mbohlool Could you please take a look here? I've signed the CLA, but the process seems to be broken. Linux Foundation helpdesk is just answering with canned answers. Is there a way to do this manually? Thanks!

mbohlool:
I don't think there is a way around it. More information along with individual's contact information here: https://github.com/kubernetes/community/blob/master/CLA.md

mbohlool:
You can also ask your question in this issue: https://github.com/kubernetes/kubernetes/issues/27796

brejoc:
Bumping again for the CLA check.

mbohlool:
Thanks for the patch. I think this will work most of the time but we can hardcode the version as it only changes when we call `scripts/update_client.sh`. This line in the script hard-code the version in `setup.py` file:

```bash
sed -i'' "s/^CLIENT_VERSION = .*/CLIENT_VERSION = \\\"${CLIENT_VERSION}\\\"/" "${SCRIPT_ROOT}/../setup.py"
```

it can do the same here too if you only add this line:

``` python
__version__ = None
```

and add this line to the `update_client.sh` script:

```bash
sed -i'' "s/^__version__ = .*/__version__ = \\\"${CLIENT_VERSION}\\\"/" "${CLIENT_ROOT}/__init__.py"
```

what do you think about this solution?

brejoc:
@mbohlool Since you've already got that in place, this is the way to go. I assume `update-client.sh` is invoked before every package generation?

mbohlool:
Yes. The process is explained in the release page.
I have just one comment. Otherwise it looks good.

mbohlool:
Would be nice if you can squash all commits into one then run update-client.sh script commit its changes separately. This will make sure everything works. If you saw changes to spec versions, that is expected. Commit those too.

brejoc:
Okay, I hope this was right. There where a lot of changed files coming from `update-client.sh`.

mbohlool:
You probably need to do a rebase, then the changes will be much less as most of them are already in master.

brejoc:
@mbohlool This looks much better!

mbohlool:
Thanks for bear with me :) This looks good.

PR #324, opened 2017-08-19
feat: add examples for asynchronous usage (callback, asycio)

tomplus:
I added examples how to use a callback and how the library can be integrated with `asyncio`. Should I add this examples to readme too ?

Related to https://github.com/kubernetes-incubator/client-python/issues/323

### DIFF
--- a/examples/async.py
+++ b/examples/async.py
@@ -1,5 +1,6 @@
 import asyncio
 import functools
+
 from kubernetes import client, config
 
 
@@ -8,12 +9,14 @@ def create_async_callback():
     def set_result(loop, future, data):
         loop.call_soon_threadsafe(future.set_result, data)
     future = asyncio.Future()
-    return functools.partial(set_result, asyncio.get_event_loop(), future), future
+    return functools.partial(
+        set_result, asyncio.get_event_loop(), future), future
 
 
 async def main():
 
-    # Configs can be set in Configuration class directly or using helper utility
+    # Configs can be set in Configuration class directly or using helper
+    # utility
     config.load_kube_config()
 
     v1 = client.CoreV1Api()
@@ -23,7 +26,11 @@ async def main():
     v1.list_pod_for_all_namespaces(watch=False, callback=callback)
     ret = await future
     for i in ret.items:
-        print("%s\t%s\t%s" % (i.status.pod_ip, i.metadata.namespace, i.metadata.name))
+        print(
+            "%s\t%s\t%s" %
+            (i.status.pod_ip,
+             i.metadata.namespace,
+             i.metadata.name))
 
 
 if __name__ == "__main__":
--- a/examples/async.py
+++ b/examples/async.py
@@ -1,0 +1,31 @@
+import asyncio
+import functools
+from kubernetes import client, config
+
+
+# return a callback and assigned future where data will be stored
+def create_async_callback():
+    def set_result(loop, future, data):
+        loop.call_soon_threadsafe(future.set_result, data)
+    future = asyncio.Future()
+    return functools.partial(set_result, asyncio.get_event_loop(), future), future
+
+
+async def main():
+
+    # Configs can be set in Configuration class directly or using helper utility
+    config.load_kube_config()
+
+    v1 = client.CoreV1Api()
+    print("Listing pods with their IPs:")
+
+    callback, future = create_async_callback()
+    v1.list_pod_for_all_namespaces(watch=False, callback=callback)
+    ret = await future
+    for i in ret.items:
+        print("%s\t%s\t%s" % (i.status.pod_ip, i.metadata.namespace, i.metadata.name))
+
+
+if __name__ == "__main__":
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(main())
--- a/examples/callbacks.py
+++ b/examples/callbacks.py
@@ -1,11 +1,16 @@
 import time
+
 from kubernetes import client, config
 
 
 def recv_namespace(data):
     print("Listing pods with their IPs:")
     for i in data.items:
-        print("%s\t%s\t%s" % (i.status.pod_ip, i.metadata.namespace, i.metadata.name))
+        print(
+            "%s\t%s\t%s" %
+            (i.status.pod_ip,
+             i.metadata.namespace,
+             i.metadata.name))
 
 
 # Configs can be set in Configuration class directly or using helper utility
--- a/examples/callbacks.py
+++ b/examples/callbacks.py
@@ -1,0 +1,20 @@
+import time
+from kubernetes import client, config
+
+
+def recv_namespace(data):
+    print("Listing pods with their IPs:")
+    for i in data.items:
+        print("%s\t%s\t%s" % (i.status.pod_ip, i.metadata.namespace, i.metadata.name))
+
+
+# Configs can be set in Configuration class directly or using helper utility
+config.load_kube_config()
+
+v1 = client.CoreV1Api()
+thread = v1.list_pod_for_all_namespaces(watch=False, callback=recv_namespace)
+
+# do something in the main thread
+for i in range(0, 10):
+    print('-- do something --')
+    time.sleep(0.5)


sebgoa:
Could you add an example with a `Watch=True` so the callbacks runs only when an event happens ?

thanks for working on this.

tomplus:
I think callbacks don't work well with `watch=True`. API returns new lines on every change but a callback is called once when a server sends some response (at least headers). Am I right ?

I've prepared a modification to `Watch()` class to be able to add a callback which will be called on every change. I'll send PR to kubernetes-client/python-base with it soon.

mbohlool:
@tomplus The new swagger codegen supports asyncio. We need to remove our manual changes to api_client.py file and then use the new version of generator. Then we can update this example and check it in. I will create an umbrella issue for that. Thanks for being patient.

tomplus:
@mbohlool Thanks, I'll check the new swagger too.

mbohlool:
@tomplus The master now does not have any manual changes in api_client.py. Can you checkout the asyncio python client in swagger-codegen? I assume it is a simple flag and you should be able to change our generator script to set the flag and try the client. As you are more familiar with this topic than me, you would be a good judge on this. Thanks.

tomplus:
@mbohlool Sure, I'll do it over the weekend.

tomplus:
I've regenerated library using swagger-codegen with switch `--library asyncio.`. I had to do some small changes. Likely there is some issue related to SSL in swagger - I'm going to confirm it and submit patch. Also urlib3 is replaced by aiohttp but module `kube_config.py` imports it.

Finally I have working piece of asynchronous code:

```
import asyncio

async def main():
    # Configs can be set in Configuration class directly or using helper
    # utility. If no argument provided, the config will be loaded from
    # default location.
    config.load_kube_config()

    v1 = client.CoreV1Api()
    print("Listing pods with their IPs:")
    ret = await v1.list_pod_for_all_namespaces(watch=False)
    for i in ret.items:
        print("%s\t%s\t%s" % (i.status.pod_ip, i.metadata.namespace, i.metadata.name))


if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

I can submit my changes but I have one fundamental question. Using the asyncio library in swagger-codegen will generate new REST client based on aiohttp instead of urlib3. It means that we get completely new library... Do we want to start deploy new library ? I'm not sure. It would be better to have two different REST client inside one library.

Let me know what do you think .

mbohlool:
We may want to start with having `kubernetes-async` library in addition to existing library to give this a try and fix problems with it and see what people think about it. We may eventually replace main library with it but I don't feel comfortable supporting async one yet so we should omit it from final releases for now.

@dims @sebgoa what do you think?

dims:
@mbohlool agree with your assessment. We can experiment elsewhere and bring it back if it shows promise.

achimnol:
I discovered the end of this thread too late. (did the same regeneration that @tomplus had already did)
So... How is the progress going here? :eyes:

tomplus:
I'm working on it in my free time. There were some issues in swagger/asyncio generator which had to be fixed. I hope I'll publish async version in next month. Stay tuned...

tomplus:
At last I’ve prepared the new library [kubernetes_asyncio](https://github.com/tomplus/kubernetes_asyncio) which is based on this Python client, but uses the asyncio generator from swagger-codegen. I've added a script to create this client to [kubernetes-client/gen](https://github.com/kubernetes-client/gen) repo in this PR https://github.com/kubernetes-client/gen/pull/60. I've also decided to incorporate functionality from sub-repository [kubernetes-client/python-base](https://github.com/kubernetes-client/python-base) because a lot of changes according to asyncio were needed.

Please take a look. Do you think it's possible to migrate this project to official kubernetes-clients' repo ?

tomplus:
@mbohlool @sebgoa @dims @roycaihw Could you take a look? Thanks.

holdenweb:
"Thank you" doesn't come often enough in the open source world. The small startup where I work bet on Kubernetes and Python as base server technologies eighteen months ago, and so far we haven't regretted it (and AWS have come on board with a managed service). Async operations bundled in will be a huge win for the Python community. So, thank you to everyone who's working on this development (or has done so).

roycaihw:
@tomplus Thanks for working on this! The examples seem to be outdated (IIUC it's compatible with release-3.0), but we definitely want to give the asyncio library (https://github.com/tomplus/kubernetes_asyncio) a try and see what people think about it. 

Asyncio is introduced in Python 3 and we need to support Python 2.7, so we have to keep the old synchronous library in this package. I'd suggest we having a `kubernetes_asyncio` library along with the existing `kubernetes` library, and letting users to switch import path to try the asyncio library. The two libraries shouldn't have any feature/user-visible difference other than the import path change.

The `kubernetes_asyncio` library depends on changes in `python-base`, but IMO maintaining two copies of `python-base` is not what we want. I have to challenge you to backport the change to `python-base` and make it generic for both libraries and minimize the difference. 

We could have two new branches `python:kubernetes_asyncio` and `python-base:kubernetes_asyncio` (replacing the existing `python-base:master` submodule) checking out from existing masters for experimenting - seeing what changes are needed and releasing `7.0.0a2` version. New changes in master can be cherrypicked into these experimental branches. 

@yliaog @mbohlool @dims @sebgoa what do you think?

tomplus:
@roycaihw  Many thanks for your response. [kubernetes_asyncio](https://github.com/tomplus/kubernetes_asyncio) is almost identical from user point of view (only changing imports and async syntax are required). I watch the official repository and catch up changes. It will be much easier if the library became experimental/official. Thanks for considering this issue.

Unfortunately there are a lot of changes related to classes from [python-base](https://github.com/kubernetes-client/python-base). To be honest I had to rewrite [Watch](https://github.com/tomplus/kubernetes_asyncio/blob/master/kubernetes_asyncio/watch/watch.py), [Stream](https://github.com/tomplus/kubernetes_asyncio/tree/master/kubernetes_asyncio/stream) to be able to work in the asynchronous environment. We can try using the named branches but it may be confusing for contributors. I have other 2 options to discuss:

- Removing/merging python-base repository - I know it’s a controversial idea and I know the [motivation](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/csi-client-structure-proposal.md), but I don’t know another library which uses `python-base` with other API groups. Is it still important for you? BTW only Python and Go use base-repos, other client libraries (Java, JS, C#, Haskell, Ruby) don’t have such base repos. 

- I can create dedicated classes for asyncio with different names (like `AsyncWatch`, `AsyncStream` etc.) in `python-base` repo. This classes will be imported by `kubernetes_asyncio` and exported again with previous name. This trick allows us to have the common `python-base` and consistent libraries.

WRT final libraries, I think it’d be better to have libraries in separate repositories instead of branches, it’s more common to have a repository per package, it’s easy to install from repo, issues are tracked separately etc.

txomon:
Hello guys, is there any way to push this forward?

scottilee:
@tomplus is this still relevant? how should we move forward with this?

yliaog:
@tomplus I like your proposal for python-base below:

"I can create dedicated classes for asyncio with different names (like AsyncWatch, AsyncStream etc.) in python-base repo. This classes will be imported by kubernetes_asyncio and exported again with previous name. This trick allows us to have the common python-base and consistent libraries."

For async library, i'm thinking to have a separate directory inside python repo for asyncio, instead of a separate repo. then the releases / issues are all tracked in one repo. The current library and the async one should be kept independent, separate. of course we need to clearly state in the release note that the async library is new, and for preview.

The library has stopped supporting python 2 starting from v18.0.0. so the async library can be added in v18.0.0 or later.
https://github.com/kubernetes-client/python/blob/b13b350dba25e279765b3a8d055e5485b5289e2d/CHANGELOG.md#v18170a1

roycaihw:
> The library has stopped supporting python 2 starting from v18.0.0

it's pending https://github.com/kubernetes-client/python/issues/1413. The plan is to drop python 2 in v18.x.0b1

remram44:
I am using kubernetes-asyncio successfully and I'm a bit confused about what is remaining in this issue (and why it's going "stale" while the requested features exists).

yliaog:
@remram44 

kubernetes-asyncio currently lives under tomplus's, the issue is to merge it to the github.com/kubernetes-client/python

yliaog:
@tomplus python-base repo has been merged into [python](https://github.com/kubernetes-client/python/tree/master/kubernetes/base). This should make it easier to add the asyncio to this repo.

Skeen:
Any progress on this?

AvlWx2014:
@tomplus et al any idea if there's a future for this in the official client library? Seems like that was the last pending thing when this issue went stale and got closed.

PR #335, opened 2017-08-31
Add delete_custom_resource example

djkonro:


examples/delete_custom_resource.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import print_function
+
+from pprint import pprint
+from time import sleep
```
pokoli: I don't see sleep is used, so it can be removed.

mbohlool:
@djkonro friendly ping

djkonro:
@mbohlool , I have updated the code

scottilee:
@djkonro do you want to update this example or close it for now since it seems to be out-of-date. For example, `read_third_party_resource` is not an available method anymore.

Issue #341, opened 2017-09-16
Backport changes in configuration.py, api_client.py, and rest.py to swagger-codegen


mbohlool:
For velocity, we changed these files and disabled generation for them, but it is now time to backport changes to [swagger-codegen](https://github.com/swagger-api/swagger-codegen) so we can use the latest generator. These are items that need to be backported:

- [x] config.http_proxy_url is now config.proxy (already exists in upstream). Release Note: `config.http_proxy_url is deprecated. use configuration.proxy instead.`
- [x] (rest.py) port configuration.assert_hostname to upstream (https://github.com/swagger-api/swagger-codegen/pull/6508)
- [x] (rest.py) port configuration.connection_pool_maxsize to upstream (with multiprocessing.cpu_count() * 5 default value) (https://github.com/swagger-api/swagger-codegen/pull/6508)
- [x] (api_client.py) Upstream file does not have callback anymore, cleanup documentation in upstream about it (https://github.com/swagger-api/swagger-codegen/pull/6508)
- [x] (api_client.py) Remove ws_client from this file. We probably need to make a separate utility like Watch for exec/attach calls (with its out ws_streaming_protocol config). #344
- [x] (configuration.py) upstream configuration is not a singleton anymore, add a default configuration to upstream (https://github.com/swagger-api/swagger-codegen/pull/6554).
- [x] ws_client should not expect any customized config (such as ws_streaming_protocol) from configuration class.
- [ ] (requirements*.txt) let the generator generate these files and then add changes to main requirement files.
- [x] (configuration.py) no upstream api_client config. validate the change for our client.
- [ ] consider enabling asyncio feature of python generator (#323, #324)
- [x] update generator version to latest and validate generated client.

```release-note
- config.http_proxy_url is deprecated. use configuration.proxy instead.
- Exec and attach calls does not return a streaming object anymore. Use kubernetes.stream instead
- Configuration is not a singleton object anymore. Please use Configuraion.set_default to change default configuration.
- Configuration class does not support `ws_streaming_protocol` anymore. Use ApiClient.set_default_header for `sec-websocket-protocol` to have the same effect.
```

PR #344, opened 2017-09-22
Use stream module to call exec/attach calls

mbohlool:
As part of #341 this will add support for stream package added in https://github.com/kubernetes-client/python-base/pull/33.


dims:
LGTM 👍

PR #346, opened 2017-09-25
Fix duplicate ampersands in query parameters.

eyeofthefrog:


eyeofthefrog:
CLA has been signed.

mbohlool:
The file is moved to the base repo. Also this fix may not work all the time. I've simplified the command expansion to also fix this problem here https://github.com/kubernetes-client/python-base/pull/35

Issue #347, opened 2017-09-25
kubectl set image ?


FingerLiu:
Just want to update image of deployment simple as `kubectl set image`
Try patch_deployment but it seems that it has a few issues, And just can not get it work well.

Is there a simple api like `kubectl set image`  or will there be more docs of how to format the
body of patch_deployment to satisfy the server ?

FingerLiu:
Update:
After some test, this format works for me to update a image of an existing container:
``` python
from kubernetes import client, config
from kubernetes.client.rest import ApiException
from pprint import pprint
config.load_kube_config()
client.configuration.debug = True
api_instance = client.AppsV1beta1Api()
body = {"spec":{"template":{"spec":{"containers":[{"name":"image","image":"gcr.io/xxx/image:tag"}]}}}}
api_response = api_instance.patch_namespaced_deployment('image', 'default', body,pretty=True)
pprint.pprint(api_response)

```

mbohlool:
Would be nice if you can add that as an example (you just need to clean it up a little and make it similar to other examples).

FingerLiu:
Make a PR https://github.com/kubernetes-incubator/client-python/pull/352, please review it.

PR #352, opened 2017-09-28
Add example to patch a deployment

FingerLiu:
Add example in issue #347


mbohlool:
we have an example for deployment patch here: https://github.com/FingerLiu/client-python/blob/9ae080693cd16ce825a977cc167803ab1f7f1202/examples/deployment_examples.py#L56

Please let me know if this is different and I will reopen the PR.

PR #353, opened 2017-10-03
Use swagger-codegen HEAD to generate the client

mbohlool:
More information at issue: https://github.com/kubernetes-incubator/client-python/issues/341

This PR is depends on base client PR: https://github.com/kubernetes-client/python-base/pull/34


mbohlool:
@dims, yes for the first one. Second one is almost generated, I only removed some file from ignore list so the generator generates them.

PR #365, opened 2017-10-17
Change IntOrString field types from string to object

mbohlool:
IntOrString is better represented as object in python (and I can argue in most of the other languages too). That way we won't have any type check on it which is not ideal but it is better than consider it as string because both serialization and deserialization would be wrong that way.

fixes https://github.com/kubernetes-incubator/client-python/issues/18
(It is not a 100% fix but I think it is close enough to fix that issue. Only missing thing for IntOrString is typecheck which is not a big deal, if it is at all, in python but could be an issue in other languages).


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Master
+- IntOrString is now object thus it can be both int or string.
```
dims: `it can store either int or string` right?

mbohlool: Yes. The downside is it can be stored anything (to that point, any other field can be anything as python is not type safe) but server will complain about anything other than int or string.

mbohlool:
For some reason, this PR didn't get any new commits and I forced to close it in favor of #366

PR #366, opened 2017-10-17
Change IntOrString field types from string to object

### LINKED ISSUE #18: int-or-string
support kubernetes custom format "int-or-string".

mbohlool:
IntOrString is better represented as object in python (and I can argue in most of the other languages too). That way we won't have any type check on it which is not ideal but it is better than consider it as string because both serialization and deserialization would be wrong that way.

fixes #18
(It is not a 100% fix but I think it is close enough to fix that issue. Only missing thing for IntOrString is typecheck which is not a big deal, if it is at all, in python but could be an issue in other languages).


dims:
LGTM 👍

PR #370, opened 2017-10-19
Add MANIFEST.in, fixes #361

### LINKED ISSUE #361: PyPI source distribution missing requirements.txt
setup.py depends on the presence of requirements.txt and test-requirements.txt, however these files are not included within the source distribution on PyPI.

This means that the package cannot be installed from the source distribution.

allanadair:


dims:
@allanadair please see :) https://github.com/kubernetes-incubator/client-python/pull/369

allanadair:
Ah, thanks @dims, I wasn't watching.

PR #375, opened 2017-10-28
Ignore trailing slashes for server url

bacher09:
`kubectl` by default ignores trailing slashes for server url, but python client isn't. For example if you have server URL like this: `https://kubernetes.example:8443/` (with trailing slash), kubectl will work, but python client will fail with `kubernetes.client.rest.ApiException: (404)`.

I think it's nice to have consistent behavior across all clients.

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -144,7 +144,8 @@ class ApiClient(object):
             body = self.sanitize_for_serialization(body)
 
         # request url
-        url = self.configuration.host + resource_path
+        # ignore trailing slashes in base url
+        url = self.configuration.host.rstrip('/') + resource_path
 
         # perform request and return response
         response_data = self.request(method, url,


bacher09:
Hi, I already signed CLA. Is any way to trigger cla checks again, or it will be update automatically after some time ?

kubernetes/client/api_client.py
```diff
@@ -144,7 +144,8 @@ def __call_api(self, resource_path, method,
             body = self.sanitize_for_serialization(body)
 
         # request url
-        url = self.configuration.host + resource_path
+        # ignore trailing slashes in base url
+        url = self.configuration.host.rstrip('/') + resource_path
```
tomplus: I'm not sure if we should modify this file. It's generated by swagger-codegen (see the preamble of this file). @karlbunch suggested different solution (#388) - to fix the value when configuration object is created. How does it look?

bacher09: Haven't seen that notice. I thought about fixing this in configuration handling code too, but then rejected that idea. This approach isn't solving issue completely, since there is possibility that user will create configuration object manually (in python code).

PR #404, opened 2017-11-29
Fix the create_rollback api to return the correct object

nhumrich:


kubernetes/client/apis/extensions_v1beta1_api.py
```diff
@@ -355,7 +355,7 @@ def create_namespaced_deployment_rollback_with_http_info(self, name, namespace,
                                         body=body_params,
                                         post_params=form_params,
                                         files=local_var_files,
-                                        response_type='ExtensionsV1beta1DeploymentRollback',
+                                        response_type='ExtensionsV1beta1DeploymentStatus',
```
tomplus: @nhumrich Thanks. This code is generated from official specification (https://github.com/kubernetes/kubernetes/blob/master/api/swagger-spec/apps_v1beta1.json#L2002). Can you report an issue there ?

PR #410, opened 2017-12-05
Reapply "Implementation for /exec using websocket"

### LINKED ISSUE #58: Exec calls

### LINKED ISSUE #409: Upgrade request required when running exec command with python-kubernetes client
I saw this image already appeared in https://github.com/kubernetes-incubator/client-python/issues/58 and was solved, but now it's broken again in 4.0

dulek:
This commit reapplies PR #120 that was removed by #353. Below is the
original commit message.

inspired by the POC from @chekolyn

* Adds a new requirement on websocket-client
* Add a new class WSClient that uses WebSocketApp from
  the websocket-client.
* Make sure we pass Authorization header
* Make sure we honor the SSL settings in configuration
* Some of the code will get overwritten when we generate
  fresh classes from swagger definition. To remind us
  added a e2e test so we don't lose the changes
* Added a new configuration option to enable/disable failures
  when hostnames in certificates don't match

Fixes #58
Fixes #409

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -25,6 +25,7 @@ from six import PY3, integer_types, iteritems, text_type
 from six.moves.urllib.parse import quote
 
 from . import models
+from . import ws_client
 from .configuration import Configuration
 from .rest import ApiException, RESTClientObject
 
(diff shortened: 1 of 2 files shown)


dims:
@dulek the websocket related code is still there ... under https://github.com/kubernetes-client/python-base/blob/master/stream/ws_client.py

dulek:
@dims: Oh, so I assume only change to api_client.py is needed?

mbohlool:
@dulek api_client.py is an auto-generated file and we should not edit it. Please refer to [example here](https://github.com/kubernetes-incubator/client-python/blob/master/examples/exec.py) to know how to do exec calls.

dulek:
Oh. Thanks for pointing it out and sorry for all the fuss!

mbohlool:
@dulek You are welcome and no worries. Let me know how your experience went with exec calls. Any improvements to that process would be most welcome.

PR #416, opened 2017-12-13
make websocket-client dependency more open

prometheanfire:
fixes: https://github.com/kubernetes-incubator/client-python/issues/413

The issue as reported in https://github.com/kubernetes-incubator/client-python/issues/262
was not reproduced in versions of websocketclient over 0.40.0, so just mask
0.40.0.  This allows consumption of possible security fixes and allows the
client to be co-installable with more python libraries.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ urllib3>=1.19.1,!=1.21  # MIT
 pyyaml>=3.12  # MIT
 google-auth>=1.0.1  # Apache-2.0
 ipaddress>=1.0.17  # PSF
-websocket-client>=0.32.0,<=0.40.0 # LGPLv2+
+websocket-client>=0.32.0,!=0.40.0,!=0.41.*,!=0.42.* # LGPLv2+
 requests # Apache-2.0
 requests-oauthlib # ISC


prometheanfire:
I just finished signing it

dims:
@prometheanfire did you see these bad versions here? :
https://github.com/kubernetes-incubator/client-python/pull/270/commits/7a6054cef845b3d7b03977a03c3b7d37cbcdecdc

dims:
Can you please update this PR with those versions too?

prometheanfire:
@dims it looks like (based on @leeyal) that only 0.40.0 was currently 'bad'.  Are there not tests for the issue that original masked those versions?

dims:
@prometheanfire i don't see a test when the cap was made :(

dims:
cc @mbohlool

prometheanfire:
in that case, if it's confirmed I'll mask them too

dims:
@prometheanfire in the interest of caution (since the PR was filed by someone who had the problem), let's please mask them

mbohlool:
I see tests are passing for "0.44" and we banned up to "0.42". Are we concern about "0.43". Did you test it with websocket 0.43?

prometheanfire:
I was not able to reproduce the error at all, I was just masking what was reported elsewhere as known bad.

prometheanfire:
Is there anything preventing this from being merged at this point?

PR #423, opened 2017-12-29
Correct example of delete_namespaced_deployment

ektar:
grace_period_seconds, orpan_descendants, and propagation_policy are part of the V1DeleteOptions body parameter.  Body is a required parameter, and if it is empty and prop. policy is set then prop policy is ignored, resulting in orphans.  If prop policy is set in the body then orphans are removed as expected.  Replaced example code with working code.


tomplus:
@ektar you are right but the problem is that the official specification says that you can use DeleteOptions or arguments if you want and both solutions should work in the same way - but there is a bug https://github.com/kubernetes/kubernetes/issues/43329

This library is auto-generated from the specification and this patch can't be merged. We have the example (https://github.com/kubernetes-incubator/client-python/blob/master/examples/deployment_examples.py#L67) which you can rewrite if you want to explain it.

ektar:
I see, thanks for the response and pointer to the example, wasn't coming up in my searches.

mbohlool:
This file is auto generated (@tomplus already explained it). Please re-open a new PR with the right changes to the example file or fix it in upstream.

PR #440, opened 2018-01-26
Update master to kubernetes 1.9

roycaihw:


mbohlool:
I prefer manual and automatic changes to be in separate commits so we can review them easier. Here the only manual is the CHANGELOG and setting file, right?

roycaihw:
@mbohlool Yes. I split the commit.

PR #457, opened 2018-02-13
Update python-base submodule

roycaihw:


mbohlool:
Can you please add changes of python base to changelog?

roycaihw:
@mbohlool The only change in python-base is the org/repo url change. Do I need to add it to the changelog?

mbohlool:
in that case no :)

Issue #471, opened 2018-03-03
Incompatibility issue with nsenter python library


huang195:
I have a piece of code that makes use of both the kubernetes python client and nsenter. This code used to work when the kubernetes python client was at version 3. However, starting at version 4, it no longer works, and I was getting the following error from nsenter:

```
  File "/usr/local/lib/python2.7/dist-packages/nsenter/__init__.py", line 109, in __enter__
    raise OSError(e, errno.errorcode[e])
OSError: [Errno 22] EINVAL
```

After a bit of digging,  it seems the problem is due to `setns` (what `nsenter` eventually calls) does not allow the caller (my program) to be multi-threaded (see here: http://man7.org/linux/man-pages/man2/setns.2.html). Coincidentally, starting in version 4 of the kubernetes python client code, `multiprocessing.pool` was introduced to handle asynchronous requests (see here: https://github.com/kubernetes-client/python/blob/release-4.0/kubernetes/client/api_client.py#L19). I was able to confirm this is the culprit - after disabling the use of `multiprocessing.pool` (luckily, its use was completely confined to `api_client.py`, and changing only a few lines of code sufficed), my code worked the same way as v3 of kubernetes python client.

My question is if it is possible to add a constructor flag to enable/disable the use of `multiprocessing.pool` for handling async requests. It seems like a small change that I can take a stab at it.

dims:
@huang195 yes, please do take a stab at it.

Issue #226, opened 2017-05-21
Use a threadpool for async requests


yuvipanda:
Currently, a new thread is created for each HTTP request. Would be nicer to have the option of using a ThreadPool with a ThreadPoolExecutor instead.

We can already wrap this ourselves when using, but would be nice to do this in the library itself so async usage is cleaner

dims:
@yuvipanda : we are using urllib3.PoolManager, please see https://github.com/kubernetes-client/python-base/blob/7c8e59b58d9a769f1ca3f818aa91c8a6538b51b1/rest.py#L110

yuvipanda:
Aaah, whoops! Didn't notice that - it was doing Thread() last time I looked (was a while ago!). Thanks :)

yuvipanda:
I'm actually curious about https://github.com/kubernetes-incubator/client-python/blob/f399ece006d43970a9d25052798782db1407fd05/kubernetes/client/api_client.py#L331 - which starts an entirely new thread for each request still. This is a thread pool, unrelated to the urllib3 connection pool...

yuvipanda:
Note: I've started using a threadpool for this in my applications and performance is much better!

mbohlool:
That is a good idea. Would I suggest you contribute it back to `swagger-codegen`. Their templates should be straightforward to edit with your changes. When you did, we can import the changes here.

PR #473, opened 2018-03-03
added async_req parameter (default to True) to ApiClient constructor

huang195:
Taking a stab at issue #471 
@dims ptal

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -59,12 +59,14 @@ class ApiClient(object):
         'object': object,
     }
 
-    def __init__(self, configuration=None, header_name=None, header_value=None, cookie=None):
+    def __init__(self, configuration=None, header_name=None, header_value=None, cookie=None, async_req=True):
         if configuration is None:
             configuration = Configuration()
         self.configuration = configuration
 
-        self.pool = ThreadPool()
+        self.async_req = async_req
+        if async_req:
+            self.pool = ThreadPool()
         self.rest_client = RESTClientObject(configuration)
         self.default_headers = {}
         if header_name is not None:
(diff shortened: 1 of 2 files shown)


kubernetes/client/api_client.py
```diff
@@ -59,12 +59,14 @@ class ApiClient(object):
         'object': object,
     }
 
-    def __init__(self, configuration=None, header_name=None, header_value=None, cookie=None):
+    def __init__(self, configuration=None, header_name=None, header_value=None, cookie=None, async_req=True):
```
roycaihw: This file is auto-generated by [swagger-codegen](https://github.com/kubernetes-client/gen). You could contribute to swagger-codegen [python template](https://github.com/swagger-api/swagger-codegen/blob/d5f36191998396f4810155dd19cecdf697a72e95/modules/swagger-codegen/src/main/resources/python/api_client.mustache). When you did, we can import the changes here. Ref: https://github.com/kubernetes-client/python/issues/226

huang195: @roycaihw sure, can migrate the code change there, but other than that, the code looks ok?

roycaihw: the code change makes sense to me, although I'm not a python expert :)

dims: @huang195 looks good to me. please do follow the path mentioned by @roycaihw

huang195: I created the PR here: https://github.com/swagger-api/swagger-codegen/pull/7782

roycaihw:
It looks like the upstream fix was stale. We recently bumped our dependency to use up-to-date upstream generator https://github.com/kubernetes-client/python/pull/931. If this issue has been fixed upstream, we should have picked the fix up

/close

Issue #445, opened 2018-01-27
python kubeconfig loading doesn't match kubectl config loading...


brendandburns:
I had a .kube/config file with duplicate clusters with the same name (but different information) in the config file, the three different entries each pointed to different clusters, two were dead, but one was alive (I had be repeatedly creating and deleting clusters)

`kubectl` command line happened to chose the working one and worked correctly.

python client library chose the first one in the file (which wasn't working) and failed.

This lead to a very confused experience until I figured out what was going on.

It will be difficult to emulate the `kubectl` parsing exactly, but we should detect duplicate entries and throw an exception.  (and I would argue that `kubectl` probably should too)

This actually probably applies to all client libraries in different languages.

@mbohlool

brendandburns:
We need a spec for this or we risk getting on a treadmill...

--brendan

________________________________

roycaihw:
@brendandburns Totally agree. Just curious if `kubectl` in your test was picking the last one or behaving differently. I will open a PR to add the duplication detection.

PR #475, opened 2018-03-08
Update python-base submodule

roycaihw:
ref https://github.com/kubernetes-client/python/issues/445


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v6.0.0b1
+- Update python-base submodule: config loader now raises exception on duplicated name in kubeconfig [kubernetes-client/python-base#47](https://github.com/kubernetes-client/python-base/pull/47)
```
mbohlool: I don't think we want "Update python-base submodule:". That is an implementation detail and users don't care about it.

roycaihw: Updated

roycaihw:
@mbohlool Done. Thanks for reviewing this PR!

PR #477, opened 2018-03-09
API change for v6.0.0b1

roycaihw:
Committing generated API change for v6.0.0b1:

- add PATCH to custom_objects_api
- documentation update: never let cluster-scoped resources skip webhooks

also updating README and CHANGELOG


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v6.0.0b1
+- Update to Kubernetes 1.10 cluster
+- New API: add PATCH to CustomObjectsApi [kubernetes-client/gen#53](https://github.com/kubernetes-client/gen/pull/53)
+- Documentation update: never let cluster-scoped resources skip webhooks [kubernetes/kubernetes#58185](https://github.com/kubernetes/kubernetes/pull/58185)
```
mbohlool: Just curious, are you manually generating this? I would be beneficial to all clients if we can automate this. I would create a tracking issue in gen repo for it.

roycaihw: Yes, I manually checked the diff and tracked down the original PRs. It would help a lot if `gen` could detect the openapi spec change since last time we generated the client

Issue #476, opened 2018-03-09
Copy Files to a pod


ltamaster:
Is there a way to copy files to a pod?, I am trying with connect_get_namespaced_pod_exec but I think is very a limited workaround.

Thanks

aogier:
Hi, unfortunately it doesn't seem there is another way to cope with that as even the [kubectl implementation](https://github.com/kubernetes/kubernetes/blob/e61b9b448bfe00030c90e80c0db0ee2c06b96515/pkg/kubectl/cmd/cp.go#L164) or [this ticket](https://github.com/fabric8io/kubernetes-client/issues/868) seems to suggest this is the only way.  
Have you managed to copy files via exec() ?

ltamaster:
Hi @aogier,
Yes,  I used the exec to copy files. My user case is coping scripts to pods (a rundeck plugin), so using exec is enough. However,  my workaround won't work with binary files.

aogier:
Hi @ltamaster, AFAICT current examples/implementation use `tar` on a standard stream pipe, like:

```
source -> tar -> stdout -> kubernetes -> stdin -> tar -> target
```

so I'd expect it should work for any file type, as long as `tar` can cope with them. If you don't mind share your current implementation I could try to help you make things work and maybe promote it to an example script to give to others too :)

ltamaster:
Hi @aogier ,

I am doing something like this:

```
    # Calling exec interactively.
    exec_command = [shell]
    resp = stream(api.connect_get_namespaced_pod_exec, name, namespace,
                  command=exec_command,
                  stderr=True, stdin=True,
                  stdout=True, tty=False,
                  _preload_content=False)

    file = open(source_file, "r")

    commands = []
    commands.append("cat <<'EOF' >" + destination_file + "\n")
    commands.append(file.read())
    commands.append("EOF\n")

    while resp.is_open():
        resp.update(timeout=1)
        if resp.peek_stdout():
            print("STDOUT: %s" % resp.read_stdout())
        if resp.peek_stderr():
            print("STDERR: %s" % resp.read_stderr())

        if commands:
            c = commands.pop(0)
            resp.write_stdin(c)
        else:
            break

    resp.close()
```

aogier:
Ah, I've seen that now. There are some problems with strings send over the websocket, this will in turn also become a python 2/3 issue. I think the problem reside in `kubernetes.stream.WSClient` implementation, I've trying patching his `write_channel()` method that way:

```python
[...]

PR #494, opened 2018-03-24
binary websocket support, plus examples

aogier:
Hello, in #476 we noted that we currently cannot stream binary data to/from pods so no feature like `kubectl cp` (and probably others) could be implemented.

I've worked on that, two PR came out. The first is kubernetes-client/python-base#52 and when it will be merged, this one will then unlock binary streams and provide couple examples on how to do it.

Problem: I had to patch core_v1_api.py swagger generated code, as swagger itself seems to ignore [binary type](https://github.com/kubernetes/kubernetes/compare/master...aogier:feature/binary-type) specification. I'm currently investigating on that.  
In the meanwhile I could produce a temporary patch in `scripts/` and apply it when `client-update`ing but I'd like to know what do you think about that part.

Thank you, ciao

### DIFF
--- a/examples/exec.py
+++ b/examples/exec.py
@@ -1,4 +1,6 @@
+import tarfile
 import time
+from tempfile import TemporaryFile
 
 from kubernetes import config
 from kubernetes.client import Configuration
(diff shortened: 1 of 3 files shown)


aogier:
I think I've successfully patched swagger in swagger-api/swagger-codegen@676da33d2089fffb332140a35b48865d93d031e7 although my tests aren't supported by a successful CI run because starting branch obsolescence.

Basically I'm using six.binary_type that translate to `str` on py2 (no-op), and `bytes` on py3 and this need to be checked throughly against py3 tests in python-base (afaict python-client tests are empty, and for sure they return OK).

I'll work on testing against our needs but seems promising !

PR #509, opened 2018-04-11
Only depend on ipaddress in python2

asottile:
The ipaddress module became a standard module in python3.3.

This uses [environment markers][pep508] to only select the `ipaddress` module
in python2.7.  In setuptools, the most compatible way to accomplish this is
through the `extras_require` field (as suggested in the [wheel][wheel] docs).

[pep508]: https://www.python.org/dev/peps/pep-0508/#id23
[wheel]: https://wheel.readthedocs.io/en/latest/#defining-conditional-dependencies

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from setuptools import find_packages, setup
+from setuptools import setup
 
 # Do not edit these constants. They will be updated automatically
 # by scripts/update-client.sh.
@@ -27,8 +27,18 @@ DEVELOPMENT_STATUS = "4 - Beta"
 # prerequisite: setuptools
 # http://pypi.python.org/pypi/setuptools
 
+EXTRAS = {}
+REQUIRES = []
 with open('requirements.txt') as f:
-    REQUIRES = f.readlines()
+    for line in f:
+        line, _, _ = line.partition('#')
+        line = line.strip()
+        if ';' in line:
+            requirement, _, specifier = line.partition(';')
+            for_specifier = EXTRAS.setdefault(':{}'.format(specifier), [])
+            for_specifier.append(requirement)
+        else:
+            REQUIRES.append(line)
 
 with open('test-requirements.txt') as f:
     TESTS_REQUIRES = f.readlines()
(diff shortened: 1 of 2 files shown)


asottile:
CC @yliaog @roycaihw -- would love a review :)

setup.py
```diff
@@ -27,8 +27,18 @@
...
+REQUIRES = []
 with open('requirements.txt') as f:
-    REQUIRES = f.readlines()
+    for line in f:
+        line, _, _ = line.partition('#')
+        line = line.strip()
+        if ';' in line:
+            requirement, _, specifier = line.partition(';')
+            for_specifier = EXTRAS.setdefault(':{}'.format(specifier), [])
+            for_specifier.append(requirement)
```
yliaog: what is the purpose of "for_specifier.append(requirement)" but for_specifier is not kept anywhere?

yliaog:
could you please write a test case for it?

asottile:
I'm not sure how I would test this, it's really only a quality of the installation itself -- generally `setup.py` isn't really testable (either it installs or it doesn't!).  tox covers this behaviour (it invokes the installation).

I've noted in `travis-ci` that the python2.7 is still installing `ipaddress` and the py35 / py36 installs are not -- meaning it works as intended.

yliaog:
would you be able to add a line in test-requirements.txt to test it?

setup.py
asottile: `for_specifier` is the list inside `EXTRAS[f":{spec}"]` -- I'd use `collections.defaultdict(list)` here but setuptools requires specific types here.

asottile:
> would you be able to add a line in test-requirements.txt to test it?

the `setup.py` is invoked by `tox` already, so no need to add anything here

setup.py
asottile: The trick here is [`setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault)

yliaog:
i think it's possible to write a test case to check which python version is run, and if python version is 2.7, then check if ipaddress module is installed, and also where it is installed, etc.

But i agree manually checking that it works is probably good enough.

asottile:
Presumably the tests would crash on an `ImportError` for `ipaddress` if it weren't being installed.  I think that's covered :)

PR #515, opened 2018-04-17
Add example for exec with `tty` enabled.

xychu:
Including resize pty via `RESIZE_CHANNEL`.

### DIFF
--- a/examples/exec_tty.py
+++ b/examples/exec_tty.py
@@ -1,0 +1,267 @@
+# This example used code(the `Interceptor` related part) from
+# http://sqizit.bartletts.id.au/2011/02/14/pseudo-terminals-in-python/
+# which is licensed under the MIT license:
+# Copyright (c) 2011 Joshua D. Bartlett
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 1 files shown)


examples/exec_tty.py
```diff
@@ -0,0 +1,267 @@
+# This example used code(the `Interceptor` related part) from
+# http://sqizit.bartletts.id.au/2011/02/14/pseudo-terminals-in-python/
+# which is licensed under the MIT license:
+# Copyright (c) 2011 Joshua D. Bartlett
```
roycaihw: Sorry that I couldn't get to this earlier. I'm concerned with the MIT licensing and worried if having another license violates our [CLA](https://github.com/cncf/cla/). Did a quick search in our code base and I didn't see any source code under MIT license in client-python and kubernetes/kubernetes (third party vendors don't apply). 

@yliaog @mbohlool Do we have existing best-practice / SIG in community that we can consult with?

I'd be happy to review & merge this PR after we sort out the license question.

yliaog: https://www.cncf.io/blog/2017/02/01/cncf-recommends-aslv2/

CNCF recommends Apache License. Not sure if MIT License is ok though.

adubkov:
Sorry for interfere here but I have a related problem, maybe you can point me to right direction.

I'm trying to add similar thing like `exec -ti` in my app and i'm using this way to interact with tty using asyncio module:

```
        packed = fcntl.ioctl(sys.stdout.fileno(),
                             termios.TIOCGWINSZ,
                             struct.pack('HHHH', 0, 0, 0, 0))
        rows, cols, h_pixels, v_pixels = struct.unpack('HHHH', packed)
        ws.write_channel(4, json.dumps({"Height": rows, "Width": cols}))

        async def forward_channel_to_stream(in_stream, out_stream, channel):
            while True:
                data = await loop.run_in_executor(None, in_stream.read_channel, channel)
                await loop.run_in_executor(None, out_stream.write, data)

        async def forward_stream_to_channel(in_stream, out_stream, channel):
            while True:
                line = await loop.run_in_executor(None, in_stream.readline)
                await loop.run_in_executor(None, out_stream.write_channel, channel, line)

        stdout = forward_channel_to_stream(ws, sys.stdout, sys.stdout.fileno())
        stderr = forward_channel_to_stream(ws, sys.stderr, sys.stderr.fileno())
        stdin = forward_stream_to_channel(sys.stdin, ws, sys.stdin.fileno())
        loop = asyncio.get_event_loop()
        loop.run_until_complete(asyncio.gather(stdin, stdout, stderr))
        loop.close()
```

It works ok, but i have one problem - when i'm typing any command, lets say `ls` and press enter it print some extra codes like this:
```
/ # ls
^[[34;5Rbashrc              datadir             etc                 lib                 media               proc                sbin                start_bncr.sh       start_full.sh       thundertool         txdb
bin                 dev                 home                lib64               mnt                 root                srv                 start_cdnserver.sh  sys                 tini                usr
config              entrypoint.sh       keystore            logs                opt                 run                 start_accel.sh      start_comm.sh       thunder             tmp                 var
/ # ^[[34;5R
```

It looks like color codes or caret move, not sure why it doesn't convert to corresponding codes in my local bash.

I would appreciate any suggestions/

PR #562, opened 2018-07-10
Update python-base submodule to pick up changes

roycaihw:
Remove required idp-certificate-authority-data in kubeconfig for oidc token refresh;
Watch properly decodes resourceVersion from custom object response.

And also update CHANGELOG about the upcoming breaking change for substituting the currently being-used `async` keyword, which is reserved in Python 3.7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# v7.0.0a1
+**NOTE:**
+- We are doing an Alpha release for v7.0.0 because there will be an **ACTION REQUIRED** API breaking change in this client to support Python 3.7, which substitutes the currently being-used `async` reserved keyword [kubernetes-client/gen#67](https://github.com/kubernetes-client/gen/pull/67)
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v7.0.0a1
+**NOTE:**
+- We are doing an Alpha release for v7.0.0 because there will be an **ACTION REQUIRED** API breaking change in this client to support Python 3.7, which substitutes the currently being-used `async` reserved keyword [kubernetes-client/gen#67](https://github.com/kubernetes-client/gen/pull/67)
+
```
yliaog: so the PR has not been merged, do you want to wait for it?

roycaihw: I'd like to release v7 alpha first, and we can release v7 stable afterwards to introduce the breaking change. What do you think?

roycaihw:
Tests are green after rebase on master. Merging

PR #569, opened 2018-07-11
Release 7.0.0a1

roycaihw:
Update version constants for 7.0.0a1 release.

/cc @yliaog


scripts/constants.py
```diff
@@ -18,13 +18,13 @@
...
 # client version for packaging and releasing.
-CLIENT_VERSION = "7.0.0-snapshot"
+CLIENT_VERSION = "7.0.0a1"
 
 # Name of the release package
 PACKAGE_NAME = "kubernetes"
 
 # Stage of development, mainly used in setup.py's classifiers.
-DEVELOPMENT_STATUS = "4 - Beta"
+DEVELOPMENT_STATUS = "3 - Alpha"
```
yliaog: why was it Beta before?

roycaihw: The master branch hosts a `X.0.0-snapshot` client version and the development status is Beta. During release we create a pre-release branch (release-7.0 in this case) from master, and modify these constants in the pre-release branch from beta snapshot to the actual versions.

PR #573, opened 2018-07-16
docs: update create_namespaced_pod_eviction example

davidxia:
to include required parameter.

Otherwise, you get
```
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Name parameter required.","reason":"BadRequest","code":400}
```

### DIFF
--- a/kubernetes/docs/CoreV1Api.md
+++ b/kubernetes/docs/CoreV1Api.md
@@ -3526,9 +3526,9 @@ configuration.api_key['authorization'] = 'YOUR_API_KEY'
 
 # create an instance of the API class
 api_instance = kubernetes.client.CoreV1Api(kubernetes.client.ApiClient(configuration))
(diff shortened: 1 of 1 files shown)


roycaihw:
Thanks for contributing! The documentation under `kubernetes/docs` is auto-generated by [swagger-codegen](https://github.com/swagger-api/swagger-codegen/blob/0eca1ff1016805aec285b2d7aa8805a30c0d7805/modules/swagger-codegen/src/main/resources/python/api_doc.mustache#L41-L60) from kubernetes openapi [spec](https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/swagger.json). And these files get re-generated every time when we do a new release. 

It's difficult to make the generator understand required values when initializing a request body. If it's not too much to ask, I would suggest you to add an example under https://github.com/kubernetes-client/python/tree/master/examples instead :)

PR #579, opened 2018-07-23
Add Azure Active Directory Library to dependencies.

brendandburns:
This is to support:

https://github.com/kubernetes-client/python-base/pull/77

btw, git submodules are terrible... :(


requirements.txt
```diff
@@ -9,3 +9,4 @@ ipaddress>=1.0.17;python_version=="2.7"  # PSF
 websocket-client>=0.32.0,!=0.40.0,!=0.41.*,!=0.42.* # LGPLv2+
 requests # Apache-2.0
 requests-oauthlib # ISC
+adal>=1.0.2
```
roycaihw: nit: document the license (# MIT) to be consistent with the other lines

brendandburns: done.

brendandburns:
Comment addressed, please re-check.

Thanks!

mbohlool:
> btw, git submodules are terrible... :(

I agree. another option is to make the python-base a library that added as requirements here instead of git-submodule but that means keeping two release process or a more complicated one that include releasing both packages together which can be worse. do you have any other suggestions?

PR #588, opened 2018-08-02
regenerate latest client gen to get custom object status scale api

juliantaylor:


kubernetes/client/apis/custom_objects_api.py
```diff
@@ -672,6 +672,240 @@ def get_cluster_custom_object_with_http_info(self, group, version, plural, name,
...
+                                        files=local_var_files,
+                                        response_type='object',
+                                        auth_settings=auth_settings,
+                                        async_req=params.get('async_req'),
+                                        _return_http_data_only=params.get('_return_http_data_only'),
+                                        _preload_content=params.get('_preload_content', True),
+                                        _request_timeout=params.get('_request_timeout'),
+                                        collection_formats=collection_formats)
+
+    def get_cluster_custom_object_status(self, group, version, plural, name, **kwargs):
```
MahdiZareie: I checked out to this branch and I tested this method.
this method returns 404 for an existing object,his sister (`get_namespaced_custom_object`)  with exactly the same parameters, returns the object successfully.

MahdiZareie: I need this method :-S
https://github.com/kubernetes-client/python/issues/607

juliantaylor: which version of kubernetes are you using? custom object subresources are enabled by default in kubernetes 1.11, in 1.10 you have to enable it with a feature gate.
You might also need to define the status in the custom resource definition, but I'm not sure about that, see https://github.com/kubernetes/sample-controller/blob/master/artifacts/examples/crd-status-subresource.yaml

juliantaylor:
it appears a new client update was already done in master so this branch can be closed

PR #615, opened 2018-08-30
fix typo

zachvalenta:


zachvalenta:
have received email from Linux Foundation confirming signature of contributor license

roycaihw:
/lgtm
/approve

Looks like this file is copied from k/k and the upstream file's typo hasn't been fixed yet: https://github.com/kubernetes/kubernetes/blob/master/hack/cherry_pick_pull.sh#L19

PR #640, opened 2018-10-08
fix Docs to use True instead of true

davegallant:
Python does not syntactically accept lowercase true. It must be True.


tomplus:
These files are auto-generated by swagger-codegen...

yliaog:
yes, please fix swagger-codegen, otherwise it will be overwritten in the next release.

davegallant:
where is the swaggger type defined for boolean? It does not appear to be in the .mustache templates here: https://github.com/zalando/connexion#connexion

PR #649, opened 2018-10-15
Generate snapshot client in preparation for 8.0.0b1 release

roycaihw:
... against kubernetes 1.12.1 release (temp 1.12.2 openapi spec). 

Also update CHANGELOG. I verified that there are only [two cherry-pick commits](https://github.com/kubernetes/kubernetes/commits/release-1.12/api/openapi-spec/swagger.json) that modified the upstream openapi spec after 1.12.0 is cut. Therefore the CHANGELOG only collects those two changes.

In my previous [PR](https://github.com/kubernetes-client/python/pull/648) updating python-base submodule, I forgot that API changes should be checked into master branch. I should have combined these two PRs into one to save review bandwidth... The next step is to merge master into release-8.0 branch and re-generate client (should be only version constants change)


kubernetes/README.md
```diff
@@ -3,7 +3,7 @@ No description provided (generated by Swagger Codegen https://github.com/swagger
 
 This Python package is automatically generated by the [Swagger Codegen](https://github.com/swagger-api/swagger-codegen) project:
 
-- API version: v1.12.1
+- API version: v1.12.2
```
roycaihw: As discussed in https://github.com/kubernetes-client/python/pull/631#discussion_r221404512, currently Kubernetes latest release is 1.12.1. We are generating this client against the temporary 1.12.2 openapi spec file. I will file an issue in [gen repo](https://github.com/kubernetes-client/gen) to track the UX improvement.

/assign @yliaog

Issue #660, opened 2018-10-23
Travis CI - server uses dated k8s version v1.7


micw523:
The Travis CI build uses an outdated server k8s version of v1.7.0 via localkube, causing testcases for PR #655 build 1009 to fail (https://travis-ci.org/kubernetes-client/python/builds/444893877) for an e2e_test case using app/v1.

To use API app/v1, we should at least use k8s version 1.9. Although, localkube has been superseded by kubeadm and should probably be replaced.

Will investigate how to change CI to use kubeadm.

PR #661, opened 2018-10-23
[WIP] Update localkube version to 1.9.4 in Travis CI

micw523:
This PR addresses #660.
Update localkube version to 1.9.4 so that we could use app/v1beta2 and app/v1 in Travis CI.


micw523:
Unable to upgrade version; close the PR and request for help

PR #665, opened 2018-10-27
Update Travis CI to use most current k8s server version

micw523:
This PR corrects behavior reported in #660 that causes e2e tests which uses API introduced after 1.7.0 to fail.

Changes:
- switch local k8s cluster from depreciated localkube to minikube (https://github.com/redspread/localkube)
- upgrade server version to most current release
- Raised Linux build platform from 14.04 to 16.04 (14.04 only supports running minikube with localkube, restricting highest version to 1.10) https://kubernetes.io/docs/setup/independent/install-kubeadm/
- Mount / to avoid dns issues (https://github.com/LiliC/travis-minikube/blob/minikube-30-kube-1.12/.travis.yml)

/assign @roycaihw

### DIFF
--- a/scripts/kube-init.sh
+++ b/scripts/kube-init.sh
@@ -30,6 +30,11 @@ trap "clean_exit" EXIT
 # Switch off SE-Linux
 setenforce 0
 
+# Mount root to fix dns issues
+# Define $HOME since somehow this is not defined
+HOME=/home/travis
+sudo mount --make-rshared /
+
 # Install docker if needed
 path_to_executable=$(which docker)
 if [ -x "$path_to_executable" ] ; then
@@ -40,7 +45,7 @@ fi
 docker --version
 
 # Get the latest stable version of kubernetes
-export K8S_VERSION=$(curl -sS https://storage.googleapis.com/kubernetes-release/release/stable.txt)
+K8S_VERSION=$(curl -sS https://storage.googleapis.com/kubernetes-release/release/stable.txt)
 echo "K8S_VERSION : ${K8S_VERSION}"
 
 echo "Starting docker service"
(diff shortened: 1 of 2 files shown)


scripts/kube-init.sh
```diff
@@ -54,48 +59,51 @@ wget -O kubectl "http://storage.googleapis.com/kubernetes-release/release/${K8S_
...
+export MINIKUBE_HOME=$HOME
+export MINIKUBE_DRIVER=${MINIKUBE_DRIVER:-none}
+
+# Used bootstrapper to be kubeadm for the most recent k8s version
+# since localkube is depreciated and only supported up to version 1.10.0
+echo "Starting minikube"
+sudo minikube start --vm-driver=$MINIKUBE_DRIVER --bootstrapper=kubeadm --kubernetes-version=$K8S_VERSION --logtostderr
+
+# Make sure minikube information is not outdated
+sudo minikube update-context
```
roycaihw: Why do we need to run `minikube update-context`? I assume it modifies the kubeconfig file to point to the minikube cluster and I'd expect `minikube start` does that for us already

scripts/kube-init.sh
```diff
@@ -54,48 +59,51 @@ wget -O kubectl "http://storage.googleapis.com/kubernetes-release/release/${K8S_
...
+
+export MINIKUBE_OK="false"
```
roycaihw: Let's not export this variable and make it local

scripts/kube-init.sh
```diff
@@ -30,6 +30,11 @@ trap "clean_exit" EXIT
 # Switch off SE-Linux
 setenforce 0
 
+# Mount root to fix dns issues
+# Define $HOME since somehow this is not defined
+export HOME=/home/travis
```
roycaihw: This is unexpected. The [travis doc](https://docs.travis-ci.com/user/environment-variables#default-environment-variables) mentioned that default the environment variables are available and `HOME=/home/travis (do not depend on this value)`. 

If the environment variable is truly missed, I'm wondering if [`~` like the previous implementation](https://github.com/kubernetes-client/python/blob/529c0d3c2fbceca0f38d1cd3ae2eb3d5cd660a07/scripts/kube-init.sh#L76-L77) works for you?

scripts/kube-init.sh
```diff
@@ -30,6 +30,11 @@ trap "clean_exit" EXIT
...
+sudo mount --make-rshared /
```
roycaihw: Which dns issue does this solve?

roycaihw:
cc @dims 
as you previously implemented the `kube-init` script

scripts/kube-init.sh
micw523: I think $HOME works in 14.04, but somehow it’s not there. I was super surprised as well. Yes, ~ works. You cannot see it since I squashed the commits with push -f...

scripts/kube-init.sh
micw523: Frankly I am not sure. I’ve seen others using it and we had disable dns in our previous script. I think ```kubedns``` sometimes does not like the CI environments.

scripts/kube-init.sh
micw523: It's effectively a safeguard: https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/update-context.go#L36

scripts/kube-init.sh
roycaihw: Read https://github.com/LiliC/travis-minikube/blob/minikube-30-kube-1.12/.travis.yml. SGTM

roycaihw:
LGTM. Please squash the commits before merging

PR #685, opened 2018-11-10
Support Python 3.7 for Travis CI

micw523:
This PR prepares for a Python 3.7 support from Travis CI.
Edit:
Travis CI do support Python 3.7 now. It requires OpenSSL 1.0.2+ and Ubuntu 16.04+
and luckily we switched to the platform on our Travis CI update...
See https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-425720905

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,5 +77,6 @@ setup(
         "Programming Language :: Python :: 3.4",
         "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
     ],
 )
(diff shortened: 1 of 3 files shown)


micw523:
/hold cancel
just couldn't remember the commands...
/this-is-not-fine

setup.py
```diff
@@ -77,5 +77,6 @@
         "Programming Language :: Python :: 3.4",
         "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
```
roycaihw: This will add 3.7 to the `pypi supported versions` badge in README (probably in next release)

PR #698, opened 2018-12-08
Update master branch to 9.0.0-snapshot client

roycaihw:
This PR updates python-base submodule and generates client for 9.0.0-snapshot. 

The next step is to fork a release-9.0 branch from master and re-generate the client with version constant updates. 

/assign @yliaog


roycaihw:
/hold

adding CHANGELOG for v6.1.0 and v8.0.1

roycaihw:
/hold cancel

CHANGELOG added

kubernetes/client/apis/admissionregistration_v1alpha1_api.py
```diff
@@ -268,18 +268,18 @@ def delete_collection_initializer_configuration_with_http_info(self, **kwargs):
                                         _request_timeout=params.get('_request_timeout'),
                                         collection_formats=collection_formats)
 
-    def delete_initializer_configuration(self, name, body, **kwargs):
```
micw523: This looks like it’s potentially going to break some codes if they called the function with func(name, body). Now user needs to call with func(name, body=xxx)

README.md
```diff
@@ -83,17 +83,18 @@ supported versions of Kubernetes clusters.
 
 #### Compatibility matrix
```
micw523: Do we indicate that we drop v6 support here or in the actual release?

kubernetes/client/apis/admissionregistration_v1alpha1_api.py
micw523: Looks like all delete methods will have this change. Probably want to document this

roycaihw: It was documented in API change. I added a duplicate of the body param changelog in Breaking Change.

README.md
roycaihw: We support 3 stable releases, so we will drop v6 support after we release v9 GA version.

PR #710, opened 2018-12-21
Update CoreV1Api.md

rj03hou:
modify create eviction of a Pod example, need to add name and namespace into body, otherwise will proceduce "Name parameter required." 400 error


rj03hou:
already sign a Contributor License Agreement (CLA).

micw523:
I believe this file is automatically generated and would require an upstream patch.

PR #718, opened 2019-01-08
Update PyYAML (CVE-2017-18342)

ramnes:


micw523:
We should not request a beta version for a production repo like this. Also, version 4.1 already fixed the problem.

requirements.txt
```diff
@@ -3,7 +3,7 @@ six>=1.9.0  # MIT
 python-dateutil>=2.5.3  # BSD
 setuptools>=21.0.0  # PSF/ZPL
 urllib3>=1.23  # MIT
-pyyaml>=3.12  # MIT
+pyyaml>=4.2b1  # MIT
```
micw523:
```suggestion
pyyaml>=4.1  # MIT
```

micw523:
@ramnes 
Thanks! I just realized that PyYaml 4.1 is indeed pulled from PyPI. However, kubernetes-client/python-base#111 just got merged. The yaml ```load``` has been replaced by ```safe_load``` in the python-base repo. Now the old versions of PyYAML should be fine as well.

Also, in new versions of PyYAML, calling ```safe_load``` has not been depreciated; instead it has been set as an equivalent fucntion of ```load```. See https://github.com/yaml/pyyaml/blob/master/lib/yaml/__init__.py#L77. Therefore we don't need to revert the code if someone is using PyYAML 4.2 or higher when it is released.

Therefore, I think we can safely close the PR - what do you think?

ramnes:
Oh, I didn't see that other PR on `python-base`, nice!

There are still some `yaml.load` lying around here though, [including one in `utils`](https://github.com/kubernetes-client/python/blob/4d4ce21265f58f79c55f6755e3ba60f4360ba172/kubernetes/utils/create_from_yaml.py#L42) which I guess might be dangerous. I'll close this PR and open a new one to replace the unsafe calls by precaution, if you don't mind.

PR #719, opened 2019-01-08
Use yaml.safe_load

ramnes:
See #718 for rationale.

@micw523


micw523:
Could you leave the one in utils in its original form just for now? #673 will change the use of ```load``` into ```load_all``` which should be fine when it gets merged. The other two look good to me.

ramnes:
@micw523 If you change #673 to use `safe_load_all` rather than `load_all`, then yes! Otherwise, I can just rebase on your PR when it's inside the tree.

micw523:
@ramnes sounds good! I'll make the change in #673.

ramnes:
@micw523 done

PR #724, opened 2019-01-11
Update python-base ref to include yaml security fix

xvello:
Update the `python-base` submodule to ship https://github.com/kubernetes-client/python-base/pull/111

Tested it with a kubeconfig file specifying a client-certificate and a certificate-authority, no regression detected.

As this addresses a security issue, I think it deserves a 8.0.2 release. What would be the process to cherry-pick it and get a release?


xvello:
hi @roycaihw 
Any chance we could cut a release (either 8.0.2 or 9.0.0 if ready) with this security fix?

PR #727, opened 2019-01-16
Added support for json files in kubectl create -f functionality

prafull01:
"I Signed it"


micw523:
The YAML loader can parse most JSON correctly so I don’t think this is necessary.

PR #732, opened 2019-01-21
feat: remove dependency on mvn from updating script

tomplus:
The scripts use docker based generator so a locally installed Maven is not necessary.


tomplus:
/remove-lifecycle stale

IMO it's worth to merge it.

roycaihw:
/approve

@tomplus good catch!

Issue #676, opened 2018-11-02
Setup.py will fail on Windows


micw523:
This issue references #479, as it is a restriction on the current repo.
This repo uses symbolic link when managing the submodules, so that the folders in /kubernetes/base can be referenced in /kubernetes.
Windows does not have symbolic link, so Windows developers will have to manually copy the folders over after doing a --recursive clone.
This should be a simple enough fix in case someone wants to take a crack at it.

jenetiks:
I'd like to take a stab at this but I have a few questions first. 
1. Would you be looking for a specific python version (2 vs 3) and more specifically, a subversion?
2. base/config and config/ are always in the same location?
3. run python setup.py install script from a specific directory?

micw523:
@jenetiks
1. We should just look for python in the environment - although, we do support Python 2.7. The file setup.py though will be configured to handle this. Personally I don’t expect any developer to clone this repository with Python 2.6 or lower, but that might be up to debate.

2. When we clone submodule with git the folders in the /kubernetes/base is supposed to be symbolically linked in the /kubernetes directory. In Windows this is not done. For a Windows developer, we’ll need to have a script copying all the base/* folders into */. 

3. It probably will have to be run in a certain directory, likely the root of the repo or scripts. It needs to be documented well where it is supposed to be run though.

4. I’ll prefer to be a bat script. Although I don’t develop this repo in Windows, I do use Anaconda and python environments variables are only available under the Anaconda Prompt (cmd). 

Correct me if you have other thoughts. /cc @roycaihw

Sai-Adarsh:
> /priority backlog

Hello @micw523 Gave a PR that solves the issue.

Added shell script to check python environment, prints the python version, copy files from /kubernetes/base to /kubernetes/ and finally runs python setup.py install.

PR #734, opened 2019-01-21
Fixes Setup.py error on Windows #676

Sai-Adarsh:
Fixes Setup.py error on Windows, added shell script to check python environment, prints the python version, copy files from /kubernetes/base to /kubernetes/ and finally runs python setup.py install issue #676

### DIFF
--- a/scripts/setup-fix.bat
+++ b/scripts/setup-fix.bat
@@ -1,3 +1,1 @@
 where python >null && echo Python environment found.. && python --version && xcopy ..\kubernetes\base\* ..\kubernetes\ /E && cd ../ && python setup.py install
-
-
--- a/scripts/setup-fix.bat
+++ b/scripts/setup-fix.bat
@@ -1,5 +1,5 @@
 (	python --version>nul 2>&1 && (
-	echo Python environment found..
+	echo Python environment found.
 	)
 	python --version
 	copy ..\kubernetes\base\* ..\kubernetes\ 
--- a/scripts/setup-fix.bat
+++ b/scripts/setup-fix.bat
@@ -1,5 +1,8 @@
-( 	where python >null
-	echo Python environment found..
+( 	python --version>nul 2>&1 && (
+		echo Python environment found..
+	) || (
+	    echo Python environment not found..
+	)
 	python --version
 	copy ..\kubernetes\base\* ..\kubernetes\ 
 	cd ../ && python setup.py install )
--- a/scripts/setup-fix.bat
+++ b/scripts/setup-fix.bat
@@ -1,7 +1,5 @@
-( 	python --version>nul 2>&1 && (
-		echo Python environment found..
-	) || (
-	    echo Python environment not found..
+(	python --version>nul 2>&1 && (
+	echo Python environment found..
 	)
 	python --version
 	copy ..\kubernetes\base\* ..\kubernetes\ 
--- a/scripts/setup-fix.bat
+++ b/scripts/setup-fix.bat
@@ -1,11 +1,3 @@
-if ! [ -x "$(command -v python)" ];
-then 
-	echo "Python environment notfound.."
-else 
-	echo "Python environment found.."
-fi
-python --version
-cp -R ../kubernetes/base/* ../kubernetes/
-cd ../ && python setup.py install
+where python >null && echo Python environment found.. && python --version && xcopy ..\kubernetes\base\* ..\kubernetes\ /E && cd ../ && python setup.py install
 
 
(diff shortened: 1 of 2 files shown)


Sai-Adarsh:
@micw523 gave a PR. could you check this out.

scripts/setup-fix.bat
```diff
@@ -0,0 +1,11 @@
+if ! [ -x "$(command -v python)" ];
+then 
+	echo Python environment notfound..
+else 
+	echo Python environment found..
+fi
+python --version
+cp -R ../kubernetes/base/* ../kubernetes/
```
yliaog: what's the reason to do the 'cp -R' above?

scripts/setup-fix.bat
```diff
@@ -0,0 +1,11 @@
+if ! [ -x "$(command -v python)" ];
+then 
+	echo Python environment notfound..
```
yliaog: probably add double quotes around the echo string

scripts/setup-fix.bat
```diff
@@ -0,0 +1,11 @@
...
+else 
+	echo Python environment found..
```
yliaog: ditto

scripts/setup-fix.bat
Sai-Adarsh: > what's the reason to do the 'cp -R' above?

"When we clone submodule with git the folders in the /kubernetes/base is supposed to be symbolically linked in the /kubernetes directory. In Windows this is not done. For a Windows developer, we’ll need to have a script copying all the base/* folders into */. "

^ is what @micw523 requested in his issue. so 'cp -R' does the job.

scripts/setup-fix.bat
Sai-Adarsh: > ditto

cool. will do the necessary changes

micw523:
To be honest this looks like bash instead of a Windows bat. You could probably use it in a Cygwin but I think Windows developers would prefer a genuine bat.

Sai-Adarsh:
@micw523 done with the changes. could you check on the new commit :)

Sai-Adarsh:
> I have a few comments -
> 
> * Could you break your command into several lines? It is difficult to decipher at its current state
> * xcopy is considered an external program and has been deprecated since '07. Could you try using copy or robocopy?

done with the changes.

scripts/setup-fix.bat
```diff
@@ -0,0 +1,5 @@
+( 	where python >null
+	echo Python environment found..
```
micw523: In this case Python environment found will be printed regardless.

micw523: ```python > null``` and ```found.```

Sai-Adarsh: > `python > null` and `found.`

cool. done with the changes.

Sai-Adarsh:
> Please fix your indents - the conditional statements are not aligned.
> Also, one period for output message is sufficient.
> On the last line `../` - the slash is not needed and on Windows you're supposed to use backslash anyway so `cd ..` should be good enough

done.

scripts/setup-fix.bat
micw523:
```suggestion
	echo Python environment found.
```

scripts/setup-fix.bat
```diff
@@ -0,0 +1,6 @@
+(	python --version>nul 2>&1 && (
+	echo Python environment found.
```
roycaihw: just out of curiosity: was the double quote change @yliaog  suggested reverted for some reason?

Sai-Adarsh: > 
> 
> just out of curiosity: was the double quote change @yliaog suggested reverted for some reason?

works fine without double qoute, so thought of reverting the change :)

Sai-Adarsh:
> 
> 
> /hold
> 
> Please add comments in the bat describing what it's for. Probably also name the file "windows-setup-fix.bat" to make it more clear. LGTM otherwise

cool.

Sai-Adarsh:
> 
> 
> /hold
> 
> Please add comments in the bat describing what it's for. Probably also name the file "windows-setup-fix.bat" to make it more clear. LGTM otherwise

@roycaihw could you check on the new changes?

roycaihw:
Could you add the comments to the beginning of the file? Basically I was asking for a description so that people can tell what it's for and when it should be used. I think we don't need inline comments.

Something like "Install python client in Windows. Windows doesn't have symbolic link that this repo uses. This batch script provides a workaround for symbolic link by copying the folders over."

Also please squash your commits before submit

/hold cancel

micw523:
Will lg()tm after squash

Sai-Adarsh:
> 
> 
> Will lg()tm after squash

cool.

PR #748, opened 2019-02-06
Update base module and regenerate client for 9.0.0b1

roycaihw:
contains several improvements. Note that [moving adal under extra_require](https://github.com/kubernetes-client/python/pull/668) is still WIP in main repo (merged in python-base repo)


yliaog:
kubernetes/client/api_client.py.orig is in the commit, is it intentional?

roycaihw:
@yliaog it's from https://github.com/kubernetes-client/gen/blob/d0821e65aa3474ce69b6bd64219d9c2082c391b7/openapi/python.sh#L64-L66

removed. I don't think we need to keep the patch backup here. I will send a change to `gen` to clean it up (like what we do for: https://github.com/kubernetes-client/python/blob/7486e196b68d83317f18f6ef9a4cc35f842d7bda/scripts/update-client.sh#L73

roycaihw:
the backup turns out to be from an unintentional mismatch. sent a cleanup https://github.com/kubernetes-client/gen/pull/102

PR #764, opened 2019-02-18
Added support for copy files from and to the pod

prafull01:
Added support to copy files/folder from and to the pods using python client.


rushant001:
it worked forme.thank you very very very much!!!

PR #767, opened 2019-02-23
Add Rbac support for creating from YAML

### LINKED ISSUE #708: create_from_yaml is converting rbac.authorization.k8s.io/v1 incorrectly
The create_from_yaml function in utils is gets an error when trying to create resources that use the rbac api.  For example, when trying to create the role below with create_from_yaml.
```
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  namespace: default
  name: pod-reader
```
It results in the following error.
```
>           k8s_api = getattr(client, fcn_to_call)(k8s_client)
E           AttributeError: module 'kubernetes.client' has no attribute 'Rbac.authorizationV1Api'
```
rbac.authorization.k8s.io/v1 is being converted into Rbac.authorizationV1Api instead of RbacAuthorizationV1Api.  Additional logic is needed to convert this correctly.

micw523:
Fixes #708.

Improves parsing of apiVersion. This bug fix is pending #673 to be merged and may need rebase.

/hold

### DIFF
--- a/kubernetes/utils/create_from_yaml.py
+++ b/kubernetes/utils/create_from_yaml.py
@@ -97,8 +97,10 @@ def create_from_yaml_single_item(
     # Take care for the case e.g. api_type is "apiextensions.k8s.io"
     # Only replace the last instance
     group = "".join(group.rsplit(".k8s.io", 1))
-    fcn_to_call = "{0}{1}Api".format(group.capitalize(),
-                                     version.capitalize())
+    # convert group name from DNS subdomain format to
+    # python class name convention
+    group = "".join(word.capitalize() for word in group.split('.'))
+    fcn_to_call = "{0}{1}Api".format(group, version.capitalize())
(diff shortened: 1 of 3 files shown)


kubernetes/utils/create_from_yaml.py
```diff
@@ -97,8 +97,8 @@ def create_from_yaml_single_item(
     # Take care for the case e.g. api_type is "apiextensions.k8s.io"
     # Only replace the last instance
     group = "".join(group.rsplit(".k8s.io", 1))
-    fcn_to_call = "{0}{1}Api".format(group.capitalize(),
-                                     version.capitalize())
+    group = "".join(word.capitalize() for word in group.split('.'))
```
roycaihw:
```suggestion
    # convert group name from DNS subdomain format to python class name convention
    group = "".join(word.capitalize() for word in group.split('.'))
```

PR #770, opened 2019-02-28
Create from dicts added to utils

vakker:
I modified the k8s utils files to be able to create form dicts, not just yamls. Let me know what you think.


oz123:
@roycaihw this PR can be closed, since it's been implemented already.

Issue #722, opened 2019-01-11
`utils` and `utils.create_from_yaml` improvement suggestions


oz123:
`create_from_yaml` should accept a string like object, like `yaml.load` and `safe_safe` do.
Then, it's the user's task to read the file.
To help users we can add a function `create_from_file` which accept a file path. `create_from_file` will than call `create_from_yaml`.

The new signatures should be:

```
def create_from_yaml(k8s_client, yaml_str, verbose=False, **kwargs):
      """
      load yaml string and apply using `create_from_map`
      """
      stream = yaml.safe_load_all(yaml_str)
      for obj in stream:
            create_from_map(k8s_client, obj, verbos)

      ...

def create_from_file(k8s_client, path_str, verbose=False, **kwargs):
      """
      read a file and continue with create_from_yaml
      """
      ...

```

Additionally, we should add:

```
 def create_from_json(k8s_client, json_str, verbose=False, **kwargs):
       """
       read a json like string  and apply by using create_from_map
       """
       ...

 def create_from_map(k8s_client, dict_like, verbose=False, **kwargs):
       """
       does all the logic previously in create_from_file.
       """
```

This approach brings the python client a little bit closer to 

`kubectl create -f <filename>`  

which accepts either files and streams from stdin, file, formatted as json and yaml).

I will happily prepare patch for this.

micw523:
The YAML loaders should parse (most) JSON correctly especially those created by kubectl so I don’t think the JSON part is needed.

I don’t think you use the -f flag if you’re parsing from stdin though.

oz123:
It's not usual to do this, but I have seen people do this in scripts, for example:
```
dnscheck_pod='{"apiVersion":"v1","kind":"Pod","metadata":{"labels":{"k8s-app":"dnscheck"},"name":"dnscheck","namespace":"default"},"spec":{"containers":[{"image":"tutum/dnsutils","name":"dnscheck","command":["sleep"],"args":["86400"]}],"securityContext":{}}}'; 
echo ${dnscheck_pod} | kubectl --kubeconfig=${KUBECONFIG} apply -f -; 
```

In any case, would you accept a PR with my suggestions?

micw523:
I take 50% of what I said back - ```create -f -``` takes input from stdin. It’s not super unusual either - this kind of usage appears on the official documentation. It would require an extra parameter if we implement in Python: file name as ```-``` and a string input.

[...]

PR #783, opened 2019-03-18
`utils` and `utils.create_from_yaml` improvements

oz123:
As discussed in #722. 

Seems this had quite some positive feedback.

### DIFF
--- a/kubernetes/utils/create_from_yaml.py
+++ b/kubernetes/utils/create_from_yaml.py
@@ -14,60 +14,119 @@
 
 
 import re
-import sys
 from os import path
 
 import yaml
-from six import iteritems
 
 from kubernetes import client
 
 
-def create_from_yaml(k8s_client, yaml_file, verbose=False, **kwargs):
+def create_from_file(k8s_client, yaml_path, verbose=False, **kwargs):
     """
-    Perform an action from a yaml file. Pass True for verbose to
-    print confirmation information.
-    Input:
-    yaml_file: string. Contains the path to yaml file.
-    k8s_cline: an ApiClient object, initialized with the client args.
+    Perform an action from a yaml file.
+
+    :param k8s_client: an ApiClient object, initialized with the client args
+    :yaml_path: the path on the file system to the yaml file
+    :type yaml_path: str
+    :param verbose: print confimation information
+    :type verbose: bool
 
     Available parameters for performing the subsequent action:
-    :param async_req bool
-    :param bool include_uninitialized: If true, partially initialized resources are included in the response.
-    :param str pretty: If 'true', then the output is pretty printed.
-    :param str dry_run: When present, indicates that modifications should not be persisted. An invalid or unrecognized dryRun directive will result in an error response and no further processing of the request. Valid values are: - All: all dry run stages will be processed
+    :param async_req:
+    :type async_req: bool
+    :param include_uninitialized: include partially initialized resources in
+    the response.
+    :type include_uninitialized: bool
+    :param str pretty: pretty print the output
+    :param str dry_run: When present, indicates that modifications should not
+    be persisted. An invalid or unrecognized dry_run directive will result in
+    an error response and no further processing of the request.
+    Valid values are: - All: all dry run stages will be processed
     """
+    with open(path.abspath(yaml_path)) as f:
+        k8s_api = create_from_yaml(k8s_client, f.read(), verbose=verbose,
+                                   **kwargs)
 
-    with open(path.abspath(yaml_file)) as f:
-        yml_object = yaml.load(f)
-        # TODO: case of yaml file containing multiple objects
-        group, _, version = yml_object["apiVersion"].partition("/")
-        if version == "":
-            version = group
-            group = "core"
-        # Take care for the case e.g. api_type is "apiextensions.k8s.io"
-        # Only replace the last instance
-        group = "".join(group.rsplit(".k8s.io", 1))
-        fcn_to_call = "{0}{1}Api".format(group.capitalize(),
-                                         version.capitalize())
-        k8s_api = getattr(client, fcn_to_call)(k8s_client)
-        # Replace CamelCased action_type into snake_case
-        kind = yml_object["kind"]
-        kind = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', kind)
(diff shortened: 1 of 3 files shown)


kubernetes/utils/create_from_yaml.py
```diff
@@ -14,60 +14,119 @@
...
 from os import path
 
 import yaml
-from six import iteritems
 
 from kubernetes import client
 
 
-def create_from_yaml(k8s_client, yaml_file, verbose=False, **kwargs):
+def create_from_file(k8s_client, yaml_path, verbose=False, **kwargs):
```
juliantaylor: create_from_stream using a file like object would be more flexible than create from file path
That way you support anything file like at once and not only files on disk.

kubernetes/utils/create_from_yaml.py
```diff
@@ -14,60 +14,119 @@
...
+    Valid values are: - All: all dry run stages will be processed
+    """
+    with open(path.abspath(yaml_path)) as f:
+        k8s_api = create_from_yaml(k8s_client, f.read(), verbose=verbose,
+                                   **kwargs)
+
+    return k8s_api
+
+
+def create_from_yaml(k8s_client, yaml_str, verbose=False, **kwargs):
```
juliantaylor: this function would be unnecessary if you add create_from stream as you could just add your yaml string to a io.StringIO() object

kubernetes/utils/create_from_yaml.py
```diff
@@ -14,60 +14,119 @@
...
+
+    return k8s_api
+
+
+def create_from_map(k8s_client, yml_object, verbose, **kwargs):
     """
+    performe an action from a valid parsed yaml object (as dict).
+
+    :param k8s_client: an ApiClient objcet initialized with the client args
+    :yml_object dict: a parsed yaml object
```
juliantaylor: I don't think one needs to explicitly reference yaml here as what the function wants is a kubernetes manifest in python dictionary form
though it probably also does not harm as yaml is the most common representation form of that data.

kubernetes/utils/create_from_yaml.py
oz123: That is true. I will send a fix later .

juliantaylor: a concern is also backward compatibility
for that this function should probably be kept with the old behavior of using file paths

Though I am no developer I don't know the rules this project has for this.
it seems the 9.0 release has not yet been cut so maybe one can still change it before release

test-requirements.txt
```diff
@@ -10,4 +10,6 @@ recommonmark
 codecov>=1.4.0
 pycodestyle
 autopep8
-isort
\ No newline at end of file
+isort
+jwt
+adal
```
roycaihw: why do we change test requirements?

micw523: I don't think we're using jwt or adal?

micw523:
I think I'm still holding my original position in #722: we don't need many function names. We should be able to do everything in ```create_from_yaml``` (potentially with an extra input parameter), since the word ```yaml``` isn't specific to a file anyway. If absolutely needed we can make an alias on ```create_from_yaml```.

Also, we're pushing to merge 673, so it's likely that you'll need a rebase soon...

test-requirements.txt
oz123: @micw523, the test failed with import error without jwt and adal.

oz123: I checked again:
```
$ grep -nr adal kubernetes | grep -v \.pyc
kubernetes/base/config/kube_config.py:24:import adal
kubernetes/base/config/kube_config.py:223:        context = adal.AuthenticationContext(
```

jwt is a dependency of adal, and I'm not sure adal is specifically required here.
I removed jwt. How do you want me to proceed with adal?

kubernetes/utils/create_from_yaml.py
oz123: @juliantaylor the function create_from_yaml now accepts a file or a string, which will be then parsed.

kubernetes/utils/create_from_yaml.py
oz123: I totally agree, but it seems that currently there will be no acceptance for that.

kubernetes/utils/create_from_yaml.py
oz123: this is why I added `create_from_map`

micw523:
Which branch did you fork it from? If you forked it from the current master I’ll need to do some investigation.

Adal as a requirement got moved to optional in a previous PR so that might be the problem.

oz123:
> Which branch did you fork it from? If you forked it from the current master I’ll need to do some investigation.
> 
> Adal as a requirement got moved to optional in a previous PR so that might be the problem.

I forked from the master on a fresh clone I did yesterday.

micw523:
I just had a fresh fork from the master branch. adal is not required for testing. Is it breaking anything that you submitted?

micw523:
I checked your past Travis builds. It doesn't seem like adal is affecting you. Please remove the added packages.

oz123:
> I checked your past Travis builds. It doesn't seem like adal is affecting you. Please remove the added packages.

done. I completely removed that commit.

oz123:
@micw523 the new master branch includes `create_from_yaml_single_item` which obsoletes my work.
I think a lot of people would appreciate a more explicit function name, because `create_from_yaml` is still misleading. A better name would have been `create_from_yaml_file`. In order to keep backward compatabilty `create_from_yaml_file` could be added in the future as a thin wrapper around `create_from_yaml_file` and if people use `create_from_yaml` use a warning, but still read the file and parse the yaml.
My two cents.

micw523:
I’m not sure. The ```create_from_single_item``` is there for a multi-resource yaml and / or yaml lists. It is NOT imported into the package nor you can use it directly with inputs from a string if your yaml string / dict contains a list type or multiple resources. This idea is not obsolete if these cases are handled.

Do you still want to work on this, or you’re going to leave it in my hands?

oz123:
@micw523 I would prefer starting on a clean branch and a clean PR. This has become too opaque for me. Do you mind if I submit a new PR?

micw523:
That’s cool, go for it :)

PR #785, opened 2019-03-19
Re-generate client for 9.0.0 release

roycaihw:
mostly version constant update. One bug fix in custom objects API to add an optional parameter

will update the CHANGELOG and README in master branch later

/assign @yliaog


micw523:
May want to drop support for v6.0 since we're rolling out a stable version?

yliaog:
i don't see the file custom_objects_spec.json for the bug fix

roycaihw:
> May want to drop support for v6.0 since we're rolling out a stable version?

@micw523 updated

> i don't see the file custom_objects_spec.json for the bug fix

@yliaog the file is in the [gen repo](https://github.com/kubernetes-client/gen/blob/master/openapi/custom_objects_spec.json). We use gen repo from HEAD to re-generate the client

Issue #741, opened 2019-01-30
Refresh token/api-key periodically


aparamon:
Currently, authorization token/api-key is only initialized on loading config:
https://github.com/kubernetes-client/python-base/blob/bd9a8525e9215f7f01c32a321beb9a605cf0402b/config/kube_config.py#L420
https://github.com/kubernetes-client/python-base/blob/bd9a8525e9215f7f01c32a321beb9a605cf0402b/config/kube_config.py#L510
When working with Amazon EKS via [aws-iam-authenticator](https://github.com/kubernetes-sigs/aws-iam-authenticator) though, token/api-key expires relatively quickly.

It is proposed to introduce a configurable option to specify token/api-key time-to-live. On API call the time should be checked, and if expired the token/api-key should be refreshed by calling https://github.com/kubernetes-client/python-base/blob/bd9a8525e9215f7f01c32a321beb9a605cf0402b/config/kube_config.py#L350 again.
Alternatively, API client could check for `401 Unauthorized` return code and refresh token (at most once per API call).

marsewe:
See also https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/63 and https://stackoverflow.com/questions/48151388/kubernetes-python-client-authentication-issue .

Eronarn:
Calling `config.load_kube_config()` again doesn't seem to recreate a token in my case, even though the corresponding kubectl command would have automatically done so. So I think there's some staleness check here that isn't actually occurring.

houqp:
config for an existing client is cached unfortunately :( After calling `load_kube_config`, you will have to recreate a new api client to pick up the refreshed token. Currently, the `expirationTimestamp` returned from `aws-iam-authenticator` is ignored by `KubeConfigLoader`.

wing328:
@houqp I wonder if you can send the patch upstream to [OpenAPI Generator](https://github.com/OpenAPITools/openapi-generator) as that's what the project going to use moving forward to generate API clients: https://github.com/kubernetes-client/gen/issues/93

houqp:
Good call @wing328, do you know if https://github.com/kubernetes-client/python/pull/738 and https://github.com/kubernetes-client/gen/pull/97 are the right PRs to watch?

houqp:
@wing328 i have ported the patches to openapi-generator, could you help review them?

roycaihw:
[...]

PR #791, opened 2019-03-21
Retry API call on token expiration

marsewe:
doesn't fit too well to the kubernetes-client design. 
Maybe s.o. has a nicer approach to fix https://github.com/kubernetes-client/python/issues/741 .

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -21,9 +21,11 @@ from multiprocessing.pool import ThreadPool
 from datetime import date, datetime
 
 # python 2 and python 3 compatibility library
+from retry_decorator import *
 from six import PY3, integer_types, iteritems, text_type
 from six.moves.urllib.parse import quote
 
+from ..config import kube_config
 from . import models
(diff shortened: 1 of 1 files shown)


micw523:
I think this is generated by swagger-codegen

roycaihw:
/remove-lifecycle rotten

we had something tackling similar problem https://github.com/kubernetes-client/python-base/pull/92 from a different angle, and it seems to work well so far. Maybe we can go with that pattern here as well

my main concerns:
- it's better to be consistent for different auth providers
- if we (python client) can do the refresh without throwing the exception, it's better to not bother the user with a 401
- when we do the refresh, we should persist the updated token to the kubeconfig file (I'm not sure if #92 does this)

roycaihw:
OTOH, if we go with catching the exception and refreshing, we can patch the generated client like [this one does](https://github.com/kubernetes-client/python/blob/master/scripts/rest_client_patch.diff)

roycaihw:
Closing due to inactivity. Please reopen as needed. 

/close

PR #795, opened 2019-03-26
Allow creation from dict

oz123:
This is a fix for #722.


kubernetes/utils/create_from_yaml.py
```diff
@@ -30,7 +31,7 @@ def create_from_yaml(
     Perform an action from a yaml file. Pass True for verbose to
     print confirmation information.
     Input:
-    yaml_file: string. Contains the path to yaml file.
+    yaml_file: string. Contains yaml or a  path to yaml file.
```
micw523: This is a little ambiguous. Does this mean yaml string, parsed yaml dict, or yaml file?
I think we should say "contains a yaml string, path to yaml file, or a dict object" or something like that

kubernetes/utils/create_from_yaml.py
```diff
@@ -54,35 +55,46 @@ def create_from_yaml(
         processing of the request.
         Valid values are: - All: all dry run stages will be processed
     """
+    if path.exists(yaml_file):
+        with open(path.abspath(yaml_file)) as f:
+            yaml_file = f.read()
+    else:
+        yaml_file = io.StringIO(yaml_file)
```
micw523: I'm not sure we need a stringio here. Also, rewriting the input to the function is kind of ambiguous.

kubernetes/utils/create_from_yaml.py
```diff
@@ -54,35 +55,46 @@ def create_from_yaml(
...
+
+    yml_document_all = yaml.safe_load_all(io.StringIO(yaml_file))
+    # Load all documents from a single YAML file
+    for yml_document in yml_document_all:
+        create_from_map(k8s_client, yml_document, verbose=verbose,
+                        **kwargs)
+
+
+def create_from_map(k8s_client, yml_document, verbose=False, **kwargs):
```
micw523: If you do it this way, the verbose does not need a "False" to it. You can call directly from your previous function with ```create_from_map(k8s_client, yml_object, verbose, **kwargs)```

kubernetes/utils/create_from_yaml.py
```diff
@@ -54,35 +55,46 @@ def create_from_yaml(
         processing of the request.
         Valid values are: - All: all dry run stages will be processed
     """
+    if path.exists(yaml_file):
```
micw523: If you want it to handle a dict type input, you could check it here.

micw523:
Also you may want to add a test case each for dict type creation (if you'll be doing it) and for creating from string.

kubernetes/utils/create_from_yaml.py
oz123: I actually meant this to be public, so people can use it without a previous function. Did you meant to put this comment on line 64?

kubernetes/utils/create_from_yaml.py
oz123: If I don't wrap the file with a io.StringIO, I will get later the following exception from
`yaml.safe_load_all`
```
E       ValueError: I/O operation on closed file.
```

kubernetes/utils/create_from_yaml.py
micw523: Oh OK. Please import the function in ```__init__.py``` then.
I'm a bit leaning towards calling it ```create_from_dict``` since the hash maps (?) are dicts in Python. What do you think?

kubernetes/utils/create_from_yaml.py
oz123: Thanks for the suggestion. I fixed this.

kubernetes/utils/create_from_yaml.py
oz123: I'd rather be explicit and have a second function do this instead of having on function to do all.
I added yaml string to load_from_yaml in order not to break previous code.

kubernetes/utils/create_from_yaml.py
oz123: Yeah, I guess we can do this. It's more pythonic.

oz123: done.

oz123:
> Also you may want to add a test case each for dict type creation (if you'll be doing it) and for creating from string.

Done. Thanks for your suggestions.

kubernetes/utils/create_from_yaml.py
```diff
@@ -30,7 +30,7 @@ def create_from_yaml(
     Perform an action from a yaml file. Pass True for verbose to
     print confirmation information.
     Input:
-    yaml_file: string. Contains the path to yaml file.
+    yaml_file: string. Contains yaml string or a  path to yaml file.
```
juliantaylor: I may be a bit odd to expect a yaml string as input.
Wrapping preexisting strings into file-like objects is normally done by the caller the reasoning being if you have already have stream you have to read it into memory just for the function to convert it back to a memory stream.
Wrapping a pre-existing memory object into a stream on the otherhand is cheap as it just boils down to a memory view.

The usual interface is file path or file-like object. a file-like is also still acceptable for the argument name yaml_file, while that name does not really fit for a string.

micw523: Is it just me or there are two spaces between "a" and "path"?

kubernetes/utils/create_from_yaml.py
juliantaylor: having the behavior of the function depend on the state of the filesystem is not good
the function should always decide based on the type of the input whether it will try to treat it as a file path or something else.
when using this string is data or a path this a bit difficult to do as they are the same type.
but for a file-like interface one option would be to check for a read function and else treat it as a file path
or consider strings and path objects as paths and everything else as file-like

kubernetes/utils/create_from_yaml.py
oz123: No, it's not just you. You are correct.

micw523: The reason we want to support string input is that ```kubectl create -f -``` takes a stdin input. I know it's kinda weird since bash does not really have dicts, but the functionality is there...

https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#create

juliantaylor: that kind of my point stdin is not a string, stdin is a file-like object
so to use this function like `kubectl create -f -` you call `create_from_yaml(sys.stdin)`
the mapping of a potential `-` commandline argument to stdin should be done by the caller

oz123: This is now fixed.

kubernetes/utils/create_from_yaml.py
oz123: @juliantaylor, basically I agree with you. I don't like Java style function with the same name but different signature. Which is also what we have here, yaml_file is two things and is completely not obvious from the function name that it expects a file. 
As said before, I'd like to have the function renamed to `create_from_file` which is explicit, but also breaks old code. So I understand why this won't be accepted. 
As a compromise, we could change the argument name from  `yaml_file` to `yaml_or_file` which is a bit more verbose, but can give a good first hint what the function expects.
@micw523 what do you think?

oz123:
@micw523 any news on this? The master branch is advancing and I am afraid this PR will rot soon.

micw523:
@oz123 Is it possible if you fast forward the submodules with respect to the upstream? Some things are broken in the CI and I think it's due to the updates in the submodules.

oz123:
@micw523 I am not sure what's the best way. I did in my branch
```
git submodule foreach git pull origin master
```
I hope this fixes the broken CI.

oz123:
@micw523 I have fixed most of the tests. Now just one test is failing (`TestUtils.test_create_apps_deployment_from_yaml_obj`) will look into this in the coming days.

micw523:
I checked the CI history and it seems that you have a type error problem when you used the stringIO.

oz123:
> I checked the CI history and it seems that you have a type error problem when you used the stringIO.

Thanks, I will look at it this weekend. I had some other issues I missed when I pushed the code.
My last commits fixed those, but it's definitely not ready.

I have notices that sometimes when the TestSuite fails it doesn't clean resources properly.
Is it worth opening an issue for that? I'd like to fix this.

oz123:
@micw523 I finally got the time to fix all the failing tests. I would appreciate further feedback.

oz123:
@micw523 ping ... any chance of reviewing this soon?

dee-me-tree-or-love:
Friendly bump, is this still on the review? :)

dee-me-tree-or-love:
Friendly bump, is this still on the review? :)

kubernetes/utils/create_from_yaml.py
```diff
@@ -54,38 +58,58 @@ def create_from_yaml(
...
+        exceptions = create_from_dict(k8s_client, yml_document, verbose,
+                                      **kwargs)
+        if exceptions:
+            fail_exceptions.extend(exceptions)
+
+    if fail_exceptions:
+        raise FailToCreateError(fail_exceptions)
+
+
+def create_from_dict(k8s_client, yml_document, verbose=False, **kwargs):
```
juliantaylor: if its create_from_dict the argument should not be named `yml_document`
maybe just document, manifest or just data

maybe the whole function should just be called `create_from_data`?
that would also fit the imo reasonable usecase of the function supporting creating objects from the models of this very library.

oz123: @juliantaylor, the most trivial and important data structure in Python is dict, hence it's very Python to expect something like this. A function named so is easy to understand, already from the name.
When I see a function called `create_from_data` I think "hmm, which data ? Is it pickle? is it some object? Is a protocolbuffer?" So I disagree here.
However, I  do agree about renaming `yml_document` to `data`.

juliantaylor: you can create objects from all of these as long as it has a mapping interface
the api functions that in the end create the objects also take dictionaries or the models so it would be consistent.
Though there is the `to_dict` function which might work it looks it returns the correct data but interestingly the call_api function does not seem to use it for serialization.
If `to_dict` works then it is probably better to keep the function simple. A comment in the documentation on how to use create with model objects would be helpful then.

kubernetes/utils/create_from_yaml.py
juliantaylor: my main concern here is more that the behavior of the function depends on the state of the filesystem
also os.path will not work on a file descriptor/stream which is the primary way of how I would like the function to work

```
os.path.exists(io.StringIO())
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-d85685824f54> in <module>()
----> 1 os.path.exists(io.StringIO())

/usr/lib/python3.7/genericpath.py in exists(path)
     17     """Test whether a path exists.  Returns False for broken symbolic links"""
     18     try:
---> 19         os.stat(path)
     20     except OSError:
     21         return False

TypeError: stat: path should be string, bytes, os.PathLike or integer, not _io.StringIO
```

kubernetes/utils/create_from_yaml.py
oz123: My latest commit has added a docstring that specifically mentions that data is expected to be a dictionary. Do you think it's lacking? Can you please suggest a better doc-string?

juliantaylor: It is fine for now, I'll have to do some testing if the `to_dict` function does the right thing (and if not that is what should be fixed).

kubernetes/utils/create_from_yaml.py
```diff
@@ -60,46 +66,77 @@ def create_from_yaml(
...
+        if exceptions:
+            fail_exceptions.extend(exceptions)
+
+    if fail_exceptions:
+        raise FailToCreateError(fail_exceptions)
+
+
+def create_from_dict(k8s_client, data, verbose=False, **kwargs):
+    """
+    Perform an action from a dictionary containing one or more valid kubernetes
```
juliantaylor: "one or more" is not really correct is it?
it only takes one kubernetes object, that one object can be a kubernetes-list but it still one object you are inputting
e.g. you cannot input two different types of objects, like a service and a deployment.

if it would take more than one object data would need to be a list/sequence of dicts/mappings

oz123: Yes, you are correct. The dictionary itself contains one item, this item by itself can contain more objects. See my last commit.

kubernetes/utils/create_from_yaml.py
oz123: This PR mostly tries to fix the issue with adding the ability to create items from dictionaries. IMHO, the function `create_from_yaml` should not read a file from the file system, this should be the responsibility of the user. But since we had to keep it backwards compatible, I chose this path.
Can you make a concrete suggestion, other wise I would like to resolve this discussion and have this PR merged.

juliantaylor: I have made two suggestions already:
- only assume a path if it is a pathlike object
- duck typing by trying read and fallback to open

indeed the other function allows the user to solve the problem already and this functions interface can be left as is.
But it isn't as is, this whole if statement can be removed and replaced with `with open: yaml.safe_load_all()`

oz123: Julian, I am sorry, your comment is still unclear to me. Please, can you create a gist with your suggestion of how to implement this function?

juliantaylor: as you have added the other function it is probably best to just leave this function as is and modify the `create_from_dict` to not catch its exception as it is very unusual api for a function that can only throw one exception to return it instead of just raising it.

This function then handles the compound exception:

```
def create_from_yaml(yaml_file, ...):
    # Load all documents from a single YAML file
    with open(yaml_file) as f:
        yml_document_all = yaml.safe_load_all(f)

    fail_exceptions = []

    for yml_document in yml_document_all:
        try:
            create_from_dict(k8s_client, yml_document, verbose, **kwargs)
        except client.rest.ApiException as api_exception:
            fail_exceptions.append(api_exception)

    if fail_exceptions:
        raise FailToCreateError(fail_exceptions)

Comment (juliantaylor) [reply kubernetes/utils/create_from_yaml.py]: oh then one is missing the support for list objects ...
why does it support list objects anyway if you cannot apply them as one object?


Ok in that case I would suggest to change the `create_from_dict` function to only support single objects.
the `create_from_yaml` then uses that when it is not a list object, and implements handling of the list objects itself.
That way `create_from_dict` is a sane function with a reasonable interace users can add their special cases on top and the odd `create_from_yaml` deprecated

Comment (juliantaylor) [reply kubernetes/utils/create_from_yaml.py]: the implementation would the look approximately like this:
(with the list loop removed from the `create_from_dict`, possibly with a good error message in case someone tries to apply such an object)

```python
def create_from_yaml(yaml_file, ...):
    # Load all documents from a single YAML file
    with open(yaml_file) as f:
        yml_document_all = yaml.safe_load_all(f)

    fail_exceptions = []

    for data in yml_document_all:
        if "List" in data["kind"]:
          # Could be "List" or "Pod/Service/...List"
          # This is a list type. iterate within its items
          kind = data["kind"].replace("List", "") 
          for yml_document in data["items"]:
              # Mitigate cases when server returns a xxxList object
              # See kubernetes-client/python#586
              if kind is not "": 
                  yml_document["apiVersion"] = data["apiVersion"]
                  yml_document["kind"] = kind
              try:
                  create_from_dict(k8s_client, yml_document, verbose, **kwargs)
              except client.rest.ApiException as api_exception:
                  fail_exceptions.append(api_exception)
        else:
            try:
                create_from_dict(k8s_client, data, verbose, **kwargs)
            except client.rest.ApiException as api_exception:
                fail_exceptions.append(api_exception)

    if fail_exceptions:
        raise FailToCreateError(fail_exceptions)

Comment (juliantaylor) [reply kubernetes/utils/create_from_yaml.py]: As this has been going on for a while I should probably clarify what I as a library user actually want:

I have a kubernetes manifest of some basic object type (represented in some standard way python, so dictionary or a model object) which I want to apply to the cluster.
When applying I do not want to have to care about which api function is the correct one to apply, I want to call a simple function to apply this single object to the cluster or fail using standard language methods (= throw an exception)
The function should figure out which api function to use to apply it (based on its kind) and whether it needs to create or patch the object in.
I want a very basic building block which is currently missing.
The `create_from_dict` you added is very close to what I want, simplify it further by removing support for List objects and it is perfect.

Reading something from a file, apply multiple objects, handling application errors, etc. can all be implemented using the building block.

Comment (oz123) [reply kubernetes/utils/create_from_yaml.py]: Thank for you for adding the code example. 
I am still not convinced your suggestions are better. 
To begin with, I don't like that `create_from_dict` can't use a `List` type. It's a valid k8s object, and it is needed. 
Further, if we adopt your implementation we use `create_from_yaml` to create from file, which IMHO is a misleading name. My desire was to first add the ability to create from from string, and in a following PR split `create_from_yaml` to `create_from_yaml` and `create_from_file`. IMHO, this is much easier for the users to understand what each of this functions does (even if it is at cost of having another function).
Can we agree on leaving it like this, and let future changes to deal with it?

Comment (juliantaylor) [reply kubernetes/utils/create_from_yaml.py]: `List` is not a kubernetes object type in the same sense as the non lists are, it is a wrapping structure to return multiple objects in.
There is no kubernetes api to apply a `List` object in a single step so the lowest level apply function does not need to support it.

`create_from_yaml` is already a misleading name, it was always `create_from_yaml_file` but I do not particularly care for about that function.

What I care about is a proper api for the `create_from_dict` and that not the same mistakes are made that were done when `create_from_yaml` was added by baking in too much usecase specific magic.
My remaining issue with that function is only the error handling, it should not return an exception for the user to raise.
It should do its job or raise an exception, either by only supporting single object application or if you have a good usecase to require support for `Lists` raise  the `FailedToCreateError` compound exception.

Comment (oz123) [reply kubernetes/utils/create_from_yaml.py]: OK, I accept this, and now `create_from_dict` throws the exception instead of  `create_from_yaml`. I hope that concludes it.

File: kubernetes/utils/create_from_yaml.py [hunk 8]
```diff
@@ -60,46 +66,69 @@ def create_from_yaml(
         processing of the request.
         Valid values are: - All: all dry run stages will be processed
     """
+    if path.exists(yaml_file):
+        with open(path.abspath(yaml_file)) as f:
```
juliantaylor: this block of code is still unnecessary, see https://github.com/kubernetes-client/python/pull/795/files#r295811946

kubernetes/utils/create_from_yaml.py
```diff
@@ -60,46 +66,69 @@ def create_from_yaml(
...
+            content = f.read()
+            try:
+                yaml_file = StringIO(content)
+            except TypeError:
+                yaml_file = StringIO(content.decode('utf-8'))
+
+    yml_document_all = yaml.safe_load_all(yaml_file)
+
+    for yml_document in yml_document_all:
+        create_from_dict(k8s_client, yml_document, verbose,
```
juliantaylor: as this function supports list objects this needs to catch the exception and build a new compound exception from it and raise that.

oz123: This was the case before, I understood that this is not desired. I will revert this.

oz123: This is already the case, `create_from_dict` gathers the exception, instead of throwing it immediately. When it is done, it throws the exception, which in turn will propagate to `create_from_yaml`.

kubernetes/utils/create_from_yaml.py
oz123: Correct. I removed this.

kubernetes/utils/create_from_yaml.py
juliantaylor: no create_from_dict gathers the exception from one object (which can be a list object)
create_from_yaml applies multiple objects which are all applied and only at the end an exception is 
the current code does not do that, it aborts after the first failure

as implemented you have to catch each exception and create a new one with all exceptions at the end (basically the same code as in from_dict but with the compound exception

juliantaylor: not resolved,
https://github.com/kubernetes-client/python/pull/795/files#r295928529

kubernetes/utils/create_from_yaml.py
```diff
@@ -60,46 +66,69 @@ def create_from_yaml(
         processing of the request.
         Valid values are: - All: all dry run stages will be processed
     """
+    if path.exists(yaml_file):
+        with open(path.abspath(yaml_file)) as f:
+            content = f.read()
```
juliantaylor: what is your intention here?
why are you using StringIO instead of just calling `yml_document_all = yaml.safe_load_all(f)`?

micw523: I think that was my comment from the old PR too.

micw523:
Sorry - I'm defending my PhD soon and am not able to review many PRs.

kubernetes/utils/create_from_yaml.py
oz123: @juliantaylor, it feels like I am not getting through here. As I said before, I am using StringIO, so I can pass either a string containing a valid yaml or a file name.
This was desired by the reviewers of this PR in order to keep it backwards compatible. Originally I added 2 functions:
1. `create_from_file` which took over the functionality of `create_from_yaml`
2. `create_from_dict` 
Also, I modified `create_from_yaml` to read a string.
This was rejected in favor of having a single function with both functionality (I don't like this either, but again, the `create_from_yaml` is not reflecting correctly what the function does).

juliantaylor: it is not necessary for backward compatilbiity the old code only dealt with filenames (which is not a good api)

even if you need it, why the indirection over stringio? the yaml library can deal with file like inputs just fine

juliantaylor: most python libraries, including yaml loading from "files" accept file-like objects or data-strings as input.
the old `create_from_yaml` unfortunately uses file paths as input which is really poor as unnecessarily you tie the function to the filesystem and you cannot reasonably distinguish between something being a file path or data as they both just bytes.

Due to this original design changing this function to do something more useful is very ugly and I think it is not worthwhile.
Instead your new _from_dict function should be sufficient as it at least allows you to do more flexible things.

oz123: > it is not necessary for backward compatilbiity the old code only dealt with filenames (which is not a good api)
Yes, the current API is not good and missleading.

> even if you need it, why the indirection over stringio? the yaml library can deal with file like inputs just fine.
This is not entirely correct, here how yaml loads a file name:
```
In [13]: !echo "{'foo': 'bar'}" > foo.txt
In [14]: yaml.safe_load('foo.txt')
Out[14]: 'foo.txt'
```
In any case, since you strongly object this, I will remove this.

kubernetes/utils/create_from_yaml.py
oz123: fixed in f30f107b4258aca3a87968881d3486ba668c4776

kubernetes/utils/create_from_yaml.py
oz123: @juliantayloer, StringIO usage is now completely remove.

juliantaylor: yaml does not load filenames, it is better designed than the `create_from_yaml´
yaml loads file-like objects:

```
yaml.load(open("foo.txt"))

File: kubernetes/utils/create_from_yaml.py [hunk 11]
```diff
@@ -59,47 +51,74 @@ def create_from_yaml(
...
+    Raises:
+        FailToCreateError which holds list of `client.rest.ApiException`
+        instances for each object that failed to create.
+    """
     with open(path.abspath(yaml_file)) as f:
         yml_document_all = yaml.safe_load_all(f)
-        api_exceptions = []
-        # Load all documents from a single YAML file
+
+        failures = []
```
juliantaylor: the indentation level can be lowered here
one does not need the file descriptor anymore at this point as the yaml was already loaded

oz123: This is incorrect. See my git commit log:

    ```
    with open('foo.txt') as f:
        y = yaml.safe_load_all(f)
    
    for i in y:
        print(i)
    
    \# raises ValueError: I/O operation on closed file.
    ```
    
    Hence, we indent the whole method body into the open block.
    
    with open('foo.txt') as f:
        y = yaml.safe_load_all(f)
        for i in y:
            print(i)

juliantaylor: oh it returns a generator, I forgot about that, nevermind then

juliantaylor:
looks good to me

oz123:
> Sorry - I'm defending my PhD soon and am not able to review many PRs.

Good luck! Can you reassign this to another member?

The the next person who reviews this, the CI pipeline is currently failing in all e2e tests. The reason seems to be that k8s is not even running in the CI instance. Can someone take care of this?

roycaihw:
/assign

I'm looking at the test failure

micw523:
Hey @roycaihw, I think k8s 1.15 is released and I'm not sure if the current minikube is ready for it.

micw523:
@oz123 @roycaihw I just checked - the current minikube only supports up to v1.14.3. What do you want to do?

oz123:
@micw523 I am not familiar with the test suite, can one specify the k8s version, and temporarily pin it to 1.14.X?

roycaihw:
@micw523 Thanks. Please feel free to send a pull to ping the [k8s version](https://github.com/kubernetes-client/python/blob/master/scripts/kube-init.sh#L48) to 1.14

oz123:
> @micw523 Thanks. Please feel free to send a pull to pin the [k8s version](https://github.com/kubernetes-client/python/blob/master/scripts/kube-init.sh#L48) to 1.14

@roycaihw see #856.

oz123:
@micw523 @roycaihw can this please moved forward , this is really wanted. See also #770 which tries to achieve the same thing.

micw523:
This LGTM if you remove the changes in kube-init.sh since we already merged that.

oz123:
@micw523, I removed the commit and forced push the branch again. This should now be legit.

micw523:
See #890 for the test problems - the default config file requests 2 CPUs to run and sometimes Travis only provides 1. I don't know why this is happening since we didn't have a problem for quite a while using minikube, but we probably want to configure minikube to use only 1 CPU in the future.

roycaihw:
I restarted the test job and it passed. Seems to be a flake

/approve

tdmalone:
Thanks for this! Bit of a kubernetes noob here, but could I ask when this might be released? (I realise I can install from source in the meantime, but keen to stick to released versions if possible)

roycaihw:
@tdmalone This will be included in v11.0.0a1, which is coming out soon. See https://github.com/kubernetes-client/python/pull/931/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1e

Issue #793, opened 2019-03-25
e2e tests fail with k8s v1.14.0


tomplus:
As we can found here: https://github.com/kubernetes/minikube/issues/3959 minikube can't start the latest version of k8s v1.14.0. Upgrading minkube to v 1.0 will be necessary when it's been released.

micw523:
/assign
Got it. Let me take a look.

PR #796, opened 2019-03-26
Temporary patch for failed e2e tests

micw523:
Temporary workaround for #793. Use k8s version 1.13.5 with minikube.
When minikube v1.0 is released this shall be replaced with an upgraded version of minikube. Marked with TODO.


micw523:
Minikube v1.0 is out - superceded by #797

PR #807, opened 2019-04-05
Drop Python3.4

oz123:
Python3.4 become EOL on 2019.03.18.

 https://www.python.org/dev/peps/pep-0429/

See also https://github.com/kubernetes-client/python-base/pull/127


roycaihw:
/lgtm
/approve

ref https://github.com/kubernetes-client/python-base/pull/127

PR #824, opened 2019-05-03
Fix AttributeError in create_from_yaml

tnqn:
Some models don't have attribute 'status', like V1ConfigMap,
V1ClusterRole, and V1NetworkPolicy. AttributeError would be raised if
these resources are created via create_from_yaml with verbose enabled.

This patch checks if attribute 'status' exists before accessing it.


micw523:
Thanks for fixing this! /lgtm

PR #831, opened 2019-05-13
Update doc: ExtensionsV1beta1Api on delete methods

caiotedim:


yliaog:
i think it is generated, it will be overwritten during next release.

roycaihw:
@yliaog was correct. This needs to be fixed in upstream code generator
/close

PR #844, opened 2019-05-30
Update nginx-deployment.yaml

jjasghar:
Updated nginx stable version


scottilee:
Hey @jjasghar thanks for your contribution but it looks like this will be taken care of along with similar problems in https://github.com/kubernetes-client/python/pull/888.

roycaihw:
Thanks for you contribution. Closing in favor of #888 which updates exhaustive places

/close

PR #855, opened 2019-06-20
add function to parse canonical quantities (e.g. resources)

juliantaylor:
This utility function is useful to parse values like the 200m or 300Gi
memory and cpu resources stored in kubernetes manifests.
The function only supports the "canonical form" which has no fractional
digits.
Currently no support for negative numbers as those to my knowledge don't appear in manifests, though it could be added for completeness.

A concern is the return value of the function, currently it is a float.
Though maybe a Decimal would be better as that represents the intention of the number somewhat better (exact representation for typically used numbers, typically arithmetic is more exact etc.) but using a somewhat rarely used type as a return value could have unforeseen consequences elsewhere (.e.g. type float checks would fail, other libraries may not support decimals properly etc.)


juliantaylor:
I don't think the test failures are related to my change

roycaihw:
/assign 

test failure is unrelated https://github.com/kubernetes-client/python/pull/795#issuecomment-504217262

juliantaylor:
rebased on master, tests are succeeding now

juliantaylor:
any chance for a review?

juliantaylor:
rebased

roycaihw:
LGTM in general. I'm also concerned about the floating number output which might defeat the purpose of the [canonical quantity](https://github.com/kubernetes/kubernetes/blob/c68eeb230af92ea834eff70ba84c79e1c4c42dd9/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity.go#L71-L80) which guarantees no precision lost

given it's on client side and currently only in parsing canonical quantities (instead of writing), perhaps it's okay to use floating numbers

/approve
/cc @yliaog 
WDYT?

scottilee:
@yliaog can you take a look when you get a moment?

kubernetes/utils/quantity.py
```diff
@@ -0,0 +1,70 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import re
+
+
+def parse_quantity(quantity):
```
yliaog: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity_test.go#L219

some cases are missing here.

juliantaylor: yes this function currently only parse the canonical form of the quantities, which is what you get in the api objects. It is not intended to re-implement the quantities in kubernetes.

it is probably reasonably straightforward to do by returning decimal objects instead of floats.
Imo that is not unreasonable, but it means user have to work fully in decimals or convert to float/integer as you cannot mix both in computations.

Should I change it to decimals and be able to parse all number representations?

juliantaylor:
Updated to use Decimal as output, add more testcases and added nano(n) and micro(u) which is apparently supported by Quantities (but not listed in its documentation)

kubernetes/utils/quantity.py
```diff
@@ -0,0 +1,73 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from decimal import Decimal, InvalidOperation
+
+
+def parse_quantity(quantity):
+    """
+    Parse kubernetes canonical form quantity like 200Mi to a decimal number.
```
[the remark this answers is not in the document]
juliantaylor: done

PR #856, opened 2019-06-21
Don't use latest stable version for minikube

oz123:
Sometimes, minikube lags behind kubernetes which can lead
to a breakage of the test suite, since minikube will fail to start
with the latest kubernetes version.
See for example [this issue](https://github.com/kubernetes/minikube/issues/4371).

This change defaults to leave the decision of which k8s version to use,
to minikube itself. This is defined in:

https://github.com/kubernetes/minikube/master/pkg/minikube/constants/constants.go

However, if one really desires it is still possible to pass

`--kubernetes-version=X.Y.Z`

to minikube initialization start command via an environment variable
before invoking the test suite:

export $MINIKUBE_ARGS="--kubernetes-version=X.Y.Z"

This allows ofcourse passing other flags to minikube also.

### DIFF
--- a/scripts/kube-init.sh
+++ b/scripts/kube-init.sh
@@ -44,9 +44,19 @@ else
 fi
 docker --version
 
-# Get the latest stable version of kubernetes
+# Get the latest stable version of kubernetes, this is not always what minikube
+# installs per default
+# See:
+# https://github.com/kubernetes/minikube/blob/master/pkg/minikube/constants/constants.go
 K8S_VERSION=$(curl -sS https://storage.googleapis.com/kubernetes-release/release/stable.txt)
 echo "K8S_VERSION : ${K8S_VERSION}"
+
+# You can pass variables to minikube using MINIKUBE_ARGS
+# If using tox you can export TOX_TESTENV_PASSENV.
+# For example, you can run:
+# $ export TOX_TESTENV_PASSENV="MINIKUBE_ARGS=--kubernetes-version=1.X.Y"
+# now tox will run minikube with the specified flag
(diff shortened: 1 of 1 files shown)


micw523:
If you take a look at the CI log you'll find that the behavior your changes introduced is to download k8s 1.15 first, and then minikube installs k8s 1.14 again. This is a little redundant in my opinion - I think we could simply pin the k8s version to 1.14 for now(?)

oz123:
Why would you want to introduce a future bug? Every-time Kubernetes changes behavior you risk minikube breaking. The "working" Kubernetes version for minikube is hardcoded in minikube, and I think you should just minikube choose it's own version.
If the python-client needs to test a specific k8s API there is still the possibility of overriding it with the environment variable.

scripts/kube-init.sh
```diff
@@ -44,10 +44,19 @@ else
 fi
 docker --version
 
-# Get the latest stable version of kubernetes
+# Get the latest stable version of kubernetes, this is not always what minikube
+# installs per default
+# See: https://github.com/kubernetes/minikube/master/pkg/minikube/constants/constants.go
```
roycaihw: nit:

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/constants/constants.go

roycaihw:
I have one nit on the link, then LGTM. We should have better mechanism to verify server-client skew, instead of tweaking minikube ad hoc

scripts/kube-init.sh
oz123: Thanks, I update the link.

PR #873, opened 2019-07-12
kube-init.sh: die is not a command

TomasTomecek:
die is not a shell command

### DIFF
--- a/scripts/kube-init.sh
+++ b/scripts/kube-init.sh
@@ -108,7 +108,8 @@ done
 # Shut down CI if minikube did not start and show logs
 if [ $MINIKUBE_OK == "false" ]; then
   sudo minikube logs
-  die $LINENO "minikube did not start"
+  echo "minikube did not start (line: ${LINENO})"
+  exit 1
 fi
 
 echo "Dump Kubernetes Objects..."


yliaog:
do you know why 'die' is not available?

micw523:
/cc
I did a little investigation.
"die" is actually a Perl command and not a bash command. When I refreshed our CI script I kept the "die" in place b/c the code was there and I thought it would let the program exit.

I think it would be OK as long as the exit return value makes sense.

TomasTomecek:
> Why do you want to return 33?

would you ask me this question if I wrote 1, 2 or 3? I dunno, it's just a number, no meaning.

micw523:
I probably still would have asked if you returned 2 or 3. 1 is commonly used for error exits isn't it?

Another reason I asked is that the original behavior (probably) wants to return the line number that it exited from. I'll have to leave the judgment to Roy and Yu on this one. I'm fine with it if they say it's ok.

scripts/kube-init.sh
```diff
@@ -108,7 +108,8 @@ done
 # Shut down CI if minikube did not start and show logs
 if [ $MINIKUBE_OK == "false" ]; then
   sudo minikube logs
-  die $LINENO "minikube did not start"
+  echo "minikube did not start"
```
yliaog: problem echo the $LINENO too

yliaog:
https://shapeshed.com/unix-exit-codes/

1 is likely a better choice.

I'm curious about what happened? the script has been in existence since 2017. what change has caused the problem?

TomasTomecek:
> I'm curious about what happened? the script has been in existence since 2017. what change has caused the problem?

I used the script in a vagrant VM to run tests - sadly minikube did not start and since there is no `set -e`, bash just said 'die: command not found' and kept on going

Issue #602, opened 2018-08-15
Kubes ops communicating via websockets and API hanging


bonneaud:
Hi,

**This is a Bug Report**
- [ ] Feature Request
- [x] Bug Report

**Problem:**
We're using Kubernetes clusters with pods running Airflow. We then have other pods running tasks that send commands to Airflow. To do that, these ops use websockets (kubernetes.stream) and Kubernetes' core_v1_api to get the information about the pod running Airflow. We run commands in one pod, from another pod, by using connect_get_namespaced_pod_exec [1] via the Kubernetes Python API, which, if we understand correctly, is a wrapper for 'kubectl exec'. Our code looks like:
stream(self.api.connect_get_namespaced_pod_exec,
            self.web_pod_name,
            self.namespace,
            command=['airflow'] + cmd,
            container='web',
            stdin='True',
            stdout='True')

Most of the times, our tasks are able to send commands and get their outputs back from Airflow's pods. Once in a while though, the tasks seem to hang forever on the stream call. When the task hangs, it is accessible via Kubes and shown as running, it simply is silent. We don't see any error.

[1] https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#connect_get_namespaced_pod_exec

**Proposed Solution:**
We don't have a solution right now, except to delete the pod via Kube and re-create it.

**Environment**:
- Kubernetes version (use `kubectl version`): 1  (1.9.7-gke.5)
- Cloud provider or hardware configuration: Google Cloud / GKE cluster
- OS (e.g. from /etc/os-release): Linux

abeltre1:
@bonneaud  I have the same issue, have you found a solution to it? 
alternatively, we can use os.system(<command>) as a hack.

PaulFurtado:
@bonneaud I may have tracked down the issue you're hitting, I filed an issue here: https://github.com/kubernetes-client/python-base/issues/106

arzarif:
@PaulFurtado - are you hitting this specifically when using CRI-O? The OP's description of this bug matches my own exactly - intermittent, indefinite hangs when invoking connect_get_namespaced_pod_exec. In my case, however, I'm using just Docker. It doesn't look like the OP specified which runtime they were using.

PaulFurtado:
@arzarif the issue I filed is relevant to every runtime, however, for the commands I tested, something about the framing of the data causes it to be very reproducible when cri-o is the runtime.

arzarif:
[...]

PR #881, opened 2019-07-16
tests: remove timeout from update()

TomasTomecek:
needed in https://github.com/kubernetes-client/python-base/pull/143

### DIFF
--- a/kubernetes/e2e_test/test_client.py
+++ b/kubernetes/e2e_test/test_client.py
@@ -13,6 +13,7 @@
 # under the License.
 
 import json
+import select
 import time
 import unittest
 import uuid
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -9,7 +9,7 @@ deps = -r{toxinidir}/test-requirements.txt
        -r{toxinidir}/requirements.txt
 commands =
    python -V
-   py.test -vvv -s --ignore=kubernetes/e2e_test
+   py.test -vvv -s --showlocals --ignore=kubernetes/e2e_test
```
roycaihw: Does this allow us to see the variable values (e.g. the `timeout`) in the test (when the test fails)? I like this

kubernetes/e2e_test/test_client.py
```diff
@@ -105,11 +106,13 @@ def test_pod_apis(self):
...
         resp.write_stdin("exit\n")
-        resp.update(timeout=5)
+        select.select([resp.sock.sock], [], [], timeout=5)
+        resp.update()
         line = resp.read_channel(ERROR_CHANNEL)
         status = json.loads(line)
         self.assertEqual(status['status'], 'Success')
-        resp.update(timeout=5)
+        select.select([resp.sock.sock], [], [], timeout=5)
+        resp.update()
```
roycaihw: I'm not sure if we want to ask existing users to do this. On the plus side this would be a hard break (unknown parameter) instead of some undefined behavior :(

cc @yliaog

tox.ini
TomasTomecek: yup, that's exactly why I did this change

TomasTomecek: but while looking at the logs in CI, it doesn't show them; that's odd

kubernetes/e2e_test/test_client.py
TomasTomecek: yes, that's a very good point: I was really just trying to remove the select call from the update method for sake of #602 and https://github.com/kubernetes-client/python-base/issues/106

okay, let me close this one and try the solution you proposed in the base PR

Issue #884, opened 2019-07-17
Fix and Update Existing Examples


scottilee:
The `*.py` files in the `examples/` directory need updates for various reasons. Some of them could also use better file names and more commenting within the files (alternatively, there could be a `README.md` that provides more information about setup and each file). Below are the files that I have tested and found issues for.

- [x] **create_deployment.py**: seems redundant with `deployment_examples.py` and `create_deployment_from_yaml.py`, I think it should be merged with one of those
- [x] **create_deployment_from_yaml.py**: fix the associated `nginx-deployment.yaml` which creates pods that go to a state of crashloopbackoff
- [x] **create_thirdparty_resource.md**: in progress of being replaced with a custom resource definition example with https://github.com/kubernetes-client/python/pull/848/
- [x] **deployment_examples.py**: add documentation
- [x] **example1.py**: add documentation and use more descriptive filename
- [x] **example2.py**: add documentation and use more descriptive filename
- [x] **example3.py**: add documentation and use more descriptive filename
- [x] **example4.py**: add documentation and use more descriptive filename, also the `print("Active host is %s" % configuration.host)` line throws the error `AttributeError: module 'kubernetes.client.configuration' has no attribute 'host'` -- in progress with https://github.com/kubernetes-client/python/pull/846/files
- [x] **exec.py**: add documentation explaining what this does
- [x] **in_cluster_config.py**: create YAML for in-cluster pod
- [x] **manage_node_labels.py**: add documentation that it is for minikube only
- [x] ~~**nginx-deployment.yaml**: using this with `create_deployment_from_yaml.py` creates nginx pods that immediately go to a state of crashloopbackoff~~ fixed by https://github.com/kubernetes-client/python/pull/888
- [x] **remote_cluster.py**: the `list_pod_for_all_namespaces` method returns a 401 unauthorized error, probably needs RBAC?

Examples without issues:
- **multiple_clusters.py**
- **job_examples.py** (https://github.com/kubernetes-client/python/pull/786/files)

roycaihw:
This is awesome! Thanks for looking into the examples and putting together this summary. 

Some additional thoughts:
- examples for creating jobs: https://github.com/kubernetes-client/python/issues/872#issuecomment-511988833, [the e2e test](https://github.com/kubernetes-client/python/blob/master/kubernetes/e2e_test/test_batch.py)
[...]

PR #888, opened 2019-07-23
Deprecate extensions/v1beta1 deployment

micw523:
- Fix a test with duplicate name introduced in #782.
- Deprecate extensions/v1beta1 deployment. (#884, comment)
- Update examples to use up-to-date nginx (#884, comment)


examples/create_deployment.py
```diff
@@ -27,7 +27,7 @@ def main():
 
     with open(path.join(path.dirname(__file__), "nginx-deployment.yaml")) as f:
         dep = yaml.safe_load(f)
-        k8s_beta = client.ExtensionsV1beta1Api()
+        k8s_beta = client.AppsV1Api()
```
roycaihw: please rename `k8s_beta` to something that makes more sense, e.g. `apps_v1`. Otherwise lgtm

micw523:
Found out that ```deployment_example.py``` needs update too. Do you want to include the changes in this PR or we should do a separate one?

roycaihw:
@micw523 let's do all the extensions/v1beta1 change in one PR

micw523:
@roycaihw done. Although ```deployment_example.py``` and the Jupyter notebooks seem to be using old apis.

examples/nginx-deployment.yaml
```diff
@@ -1,17 +1,21 @@
...
+      app: nginx
   template:
     metadata:
       labels:
         app: nginx
     spec:
       containers:
       - name: nginx
-        image: nginx:1.7.9
+        image: nginx:1.15.4
```
scottilee: Is there anything wrong with using `latest` so we don't have to keep bumping it in the future? Obviously not good practice but given it's an example I think it should be fine.

micw523: I'm keeping it on pace with https://k8s.io/examples/controllers/nginx-deployment.yaml. Otherwise I don't have a good reason :/

Issue #882, opened 2019-07-16
Python Client Documentation Site Needs Updating


scottilee:
The Python Kubernetes Client documentation site (https://kubernetes.readthedocs.io/en/latest/) is out-of-date and could use better attention to formatting.

Per our discussion in the Python Client meeting this morning please investigate:

- if someone is working on developing or replacing this site
- if auto-generation of the site is broken or needs to be setup

cc @roycaihw

micw523:
Looks like we need a markdown engine for Sphinx. tbh if we convert the readme to rst this site should be working again.

@roycaihw , is this website automatically being deployed?

roycaihw:
Thanks for pointing this out. The site looks stale with v1.8 APIs, so I don't think it's automatically being deployed. 

From the [original PR](https://github.com/kubernetes-client/python/pull/95/files#r95669875):
- the intention was to have the test catch out-of-date doc, but we aren't getting the signal from travis-ci
- either running test or a script should update the doc and allow the contributor to check-in the up-to-date files; we are missing documentation around how to do this
- I tried running `python setup.py build_sphinx` locally and it generates some html pages under `/doc/build`, which matches [this config](https://github.com/kubernetes-client/python/blob/02358ec60ff847e860eb58f037bde43cc7adcf90/setup.cfg#L6-L7). I'm not sure if `python setup.py upload_sphinx` would be the next step here to deploy the content

Since the site has been stale for a long time, I'm not sure if anyone is relying on it (I wasn't aware of this site before), but I think it has value in being a nice web page aggregating the READMEs (generated and manually maintained) in this repo

cc @dims Do you have suggestions about how we can improve this site?

micw523:
I was able to do something like this using our existing README after some exploration.
![readme](https://user-images.githubusercontent.com/13098656/61678894-5757eb00-acd2-11e9-874b-172f5ead8b90.PNG)
It's possible to generate the website using Sphinx but likely requires some effort. I'll assign myself for now since it could take a while.
/assign

oz123:
@micw523 I have experience with rendering markdown in sphinx. Would you like me to work on this?

micw523:
Sounds good - feel free to reply that you’ve taken the issue.

I think there are a few things we should do:
- the current repo uses Sphinx 1.3 and is way out of date. We should think of maybe moving on to a higher version.
[...]

PR #893, opened 2019-07-24
Docs update

oz123:
Update the configuration for building sphinx documentation with markdown. 
Now README.md is fully rendered, and it is looking much better.

Do note, this is the first PR in a series of PRs to come. I don't want to PR to be too big,
so it does not take 3 months to merge.

This PR is a partial fix for #882.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,7 @@ target/
 .idea/*
 *.iml
 .vscode
+
+# created by sphinx documentation build
+doc/source/README.md
+doc/_build
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,0 +1,21 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line.
+SPHINXOPTS    = -c source
+SPHINXBUILD   = sphinx-build
+SPHINXPROJ    = kubernetes-python
+SOURCEDIR     = source
+BUILDDIR      = build
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
+# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
+html:
+	$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	@echo "\nDocs rendered successfully, open _/build/html/index.html to view"
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,11 @@
+Building the documentation
+==========================
+
+Install the test requirements with:
+
+```
+$ pip install -r ../test-requirements.txt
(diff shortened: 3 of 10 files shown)


micw523:
I think the reason your build failed is probably due to out-of-date sphinx specified in the test-requirements.

oz123:
@micw523 I didn't realize that the CI tests the docs. That is a *good* thing.
The requirements are missing a package. I will update the CI to install the correct dependencies.

micw523:
Code lgtm. I’ll need to go render it myself to take a look though.

oz123:
@micw523 there is an online version of the updated docs in 
https://k8s-python.readthedocs.io/en/latest/README.html.

d

Also, here what needs to be done to install the requirements correctly:
sphinx

micw523:
LGTM. The commits should probably be squashed since our bot does not have the capability. What do you think @roycaihw?

oz123:
> Sorry, but could you investigate why the [the contributing](https://k8s-python.readthedocs.io/en/latest/contributing.html) page is not rendered correctly on the website you've shown?

Check:

https://raw.githubusercontent.com/oz123/python/docs_update/doc/source/contributing.rst
That is easy to explain:

The file is rst, but the content is markdown (`.. include:: ../../CONTRIBUTING.md`, just dumps it to the file). Now, sphinx does the following if the filename ends  with rst it parses it as rst. So since the content is markdown isn't understood by the parser it's left as is.

That said, I can probably fix it later today.

micw523:
Um, maybe it's just me not finding it, but what is the purpose of the ```requirements-docs``` file now?

oz123:
> Um, maybe it's just me not finding it, but what is the purpose of the `requirements-docs` file now?

Well, it's overlapping with test-requirements.txt. However, IMHO it should be seperated. If I removed the content of requirements-docs from  test-requirements.txt it would break the CI.
I'd like to avoid a large PR, and at this stage not to remove this content but also leave requirements-docs as is as an explicit reminder to documentation contributors.
I would also like to remove those requirements from test-requirements.txt  and fix the CI pipeline in a future PR.

micw523:
I think that the test requirements file is being checked by the ```setup.py``` file to determine which package to install. These packages inside will be installed on the VM when we run the CI. Having said that, you could make the ```setup.py``` file to look at your ```requirement-docs``` file and install the packages in there. I'm not sure if that's good practice though (?)

oz123:
> You could make the `setup.py` file to look at your `requirement-docs` file and install the packages in there. I'm not sure if that's good practice though (?)

That would be too much magic packed in `setup.py` and is a bad thing. Some people use setup.py as a replacement for makefile or decent package build. While you can do it, IMHO it is bad. `setup.py` should take care of setting up your code to work in your python environment. Reading and compiling the docs has little to do with. Most linux packages tend also to separate this (e.g debian which separates python packages to python3-xxx and python3-xxx-doc.
I think this client should follow this practice.

micw523:
/lgtm
Given that the doc test does almost nothing right now, I'm gonna lgtm this.

ConscriptMR:
> @micw523 there is an online version of the updated docs in
> https://k8s-python.readthedocs.io/en/latest/README.html.
> 
> d
> 
> Also, here what needs to be done to install the requirements correctly:
> sphinx

I can't find any actual documentation on this site :(

Issue #872, opened 2019-07-12
error in create_namespaced_job 'property'  - object has no attribute 'swagger_types'


simondh:
error in create_namespaced_job
'property' object has no attribute 'swagger_types'

I have traced this to an error in /kubernetes/client/api_client.py, sanitize_for_serialization, where the recursion calls this function to 'sanitize' a single entity of type Property. This falls through to the obj_dict comprehension, where it bombs as the object has no .swagger_type attibute.
I have fixed it locally with this change (2 lines in bold). The code that triggers the issue is attached
[k8-python-error.txt](https://github.com/kubernetes-client/python/files/3386411/k8-python-error.txt). I can submit a pull request if desired. 

        if isinstance(obj, dict):
            obj_dict = obj
        **elif isinstance(obj, property):
            return None**
        else:
            # Convert model obj to dict except
            # attributes `swagger_types`, `attribute_map`
            # and attributes which value is not None.
            # Convert attribute name to json key in
            # model definition for request.
            obj_dict = {obj.attribute_map[attr]: getattr(obj, attr)
                        for attr, _ in iteritems(obj.swagger_types)
                        if getattr(obj, attr) is not None}

roycaihw:
thanks for the report!

I looked at `k8-python-error.txt` and I'm not sure if it's a bug in `api_client.py`

I pulled together a small example mimicking the [k8s doc](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#running-an-example-job) and it worked for me:
```
from kubernetes import client, config

def main():
    config.load_kube_config()
    container = client.V1Container(name="pi", image="perl",
        command=["perl", "-Mbignum=bpi", "-wle", "print bpi(2000)"])
    template = client.V1PodTemplateSpec(spec=client.V1PodSpec(containers=[container], restart_policy="Never"))
    job = client.V1Job(api_version="batch/v1", kind="Job",
        metadata=client.V1ObjectMeta(name="test-job"),
        spec=client.V1JobSpec(template=template, backoff_limit=4))
    batch_v1 = client.BatchV1Api()
    api_response = batch_v1.create_namespaced_job(body=job, namespace="default")
    print("Job creation status='%s'" % str(api_response.status))

if __name__ == '__main__':
    main()
```

```
Job creation status='{'active': None,
 'completion_time': None,
 'conditions': None,
 'failed': None,
 'start_time': None,
 'succeeded': None}'

$ kubectl get jobs
[...]

PR #898, opened 2019-07-26
Added additional examples

ratanboddu:
Added the following examples :
- Jobs
- Hooks
- Namespaces
- Services
- Ingress
- Deployment Strategies
       -Ramped
       -Recreate

### DIFF
--- a/examples/create_ingress.py
+++ b/examples/create_ingress.py
@@ -1,0 +1,55 @@
+# Copyright 2016 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from kubernetes import client, config
+
+
+def main():
+    # Fetching and loading Kubernetes Information
+    config.load_kube_config()
+    # For incluster details
+    # config.incluster_kube_config()
+
+    extension = client.ExtensionsV1beta1Api()
+
+    body = client.ExtensionsV1beta1Ingress(
+        api_version="v1beta1",
+        kind="Ingress",
+        metadata=client.V1ObjectMeta(name="ingress-example", annotations={"nginx.ingress.kubernetes.io/rewrite-target": "/"}),
+        spec=client.ExtensionsV1beta1IngressSpec(
+            rules=[client.ExtensionsV1beta1IngressRule(
+                host="abc.xyz.com",
+                http=client.ExtensionsV1beta1HTTPIngressRuleValue(
+                    paths=list[client.ExtensionsV1beta1HTTPIngressPath(
+                        path="/api",
+                        backend=client.ExtensionsV1beta1IngressBackend(
+                            service_name="ingress",
+                            service_port=5000
+
+                        )
+                    )]
+                )
+            )
+            ]
+        )
+    )
+
+    extension.create_namespaced_ingress(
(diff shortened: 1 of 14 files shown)


examples/create_ingress.py
```diff
@@ -0,0 +1,57 @@
...
+# limitations under the License.
+
+from kubernetes import client, config
+
+
+def main():
+    # Fetching and loading Kubernetes Information
+    config.load_kube_config()
+    # For incluster details
+    # config.incluster_kube_config()
```
roycaihw: [`load_incluster_config`](https://github.com/kubernetes-client/python-base/blob/4f7bcfaa91d8f95d358f8215b4efacda734492c2/config/incluster_config.py#L89) (also in the other files)

examples/create_ingress.py
```diff
@@ -0,0 +1,57 @@
...
+
+    extension = client.ExtensionsV1beta1Api()
```
roycaihw: [the official doc](https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource) recommends using `networking.k8s.io/v1beta1` api instead of extensions v1beta1

examples/create_ingress.py
```diff
@@ -0,0 +1,57 @@
...
+
+    body = client.ExtensionsV1beta1Ingress(
+        api_version="v1beta1",
```
roycaihw: missing api group in `api_version` field

examples/create_ingress.py
```diff
@@ -0,0 +1,57 @@
...
+                        )
+                    )]
+                )
+            )
+            ]
+        )
+    )
+
+    extension.create_namespaced_ingress(
+        namespace="kube-client",
```
roycaihw: this requires the `kube-client` namespace to be created first. I'd suggest using the `default` namespace for a standalone example (also in the other files)

examples/create_job.py
```diff
@@ -0,0 +1,88 @@
...
+from kubernetes import client, config
+
+
+def main():
+    # Fetching and loading Kubernetes Information
+    config.load_kube_config()
+    # For incluster details
+    # config.incluster_kube_config()
+
+    extension = client.BatchV1Api()
```
roycaihw: name the api object corresponding to the api being used, e.g. `batch_v1`

examples/create_job.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Volume
+    volume = client.V1Volume(
+        name="test-volume",
+        empty_dir=client.V1EmptyDirVolumeSource(medium=""))
+
+    # Container
+    container = client.V1Container(
+        name="jobtest",
+        image="nginx:1.7.9",
```
roycaihw: running nginx doesn't sound like a good use case for job to me

examples/create_job.py
```diff
@@ -0,0 +1,88 @@
...
+    job = client.V1Job(
+        kind="Job",
+        metadata=client.V1ObjectMeta(name="jobtest"),
+        spec=spec_pod
+    )
+
+    extension.create_namespaced_job(
+        namespace="kube-client",
+        body=job
+    )
```
roycaihw: will the example print some information showing the job succeeded? 

I'd suggest creating the job template in the [official doc](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#running-an-example-job) and print the pod logs. I have a snippet written for the first step already https://github.com/kubernetes-client/python/issues/872#issuecomment-511988833

ratanboddu:
> Thanks for contributing! I think the examples will be super helpful. I went through half of them and left some comments.
> 
> Since these examples are standalone, could you split this PR into smaller pieces (maybe 2~3 PRs, each containing a couple of exampels), which will be easier to review and keep track of?

As suggested I've raised a PR for creating an Ingress #933 and will create more PR's for the remaining examples in the coming days

scottilee:
@ratanboddu can you close this PR since you are splitting it out into other PRs?

lucheeseng827:
@roycaihw I still could not figure out the secret like docker secret. How do you pass the docker secret to authenticate the job to download from docker hub?

PR #900, opened 2019-07-28
Fix small grammatical typo for possessive "its"

rcjsuen:


rcjsuen:
/check-cla

I have completed the CLA.

micw523:
This file is automatically generated so you’ll need to patch upstream instead.

rcjsuen:
@micw523 Thank you for reviewing this. From looking at the `kubernetes/kubernetes` repository, it seems like their version of the `swagger.json` file [doesn't have this problem](https://raw.githubusercontent.com/kubernetes/kubernetes/23649560c060ad6cd82da8da42302f8f7e38cf1e/api/openapi-spec/swagger.json).

I guess it's just a matter of time before the `swagger.json` file in this repository gets updated then?

micw523:
Yes, that would be the expected result.

roycaihw:
Yes this was fixed upstream https://github.com/kubernetes/kubernetes/pull/77781. New releases against 1.15+ swagger will include the fix

/close

Issue #890, opened 2019-07-23
Test flake due to CPU requested in Travis CI


micw523:
See my personal build [here](https://travis-ci.org/micw523/python/jobs/562374061).

It seems that the default config requests 2 CPUs but sometimes Travis CI only provides 1. This behavior causes Minikube to crash and the py*-functional test would fail. We should request 1 CPU from minikube instead.

roycaihw:
minikube cpu usage is configurable using the [command line flag](https://github.com/kubernetes/minikube/blob/165c1afd331765aaa94d8db684b0045fb0713cc9/cmd/minikube/cmd/start.go#L130). I haven't tested if it can work with 1 cpu

Before that, 2 cpus doesn't sound like a huge value and I'm surprised that travis ci cannot guarantee that. My assumption is that workers running in parallel can share the cpus, and most of the time we only have one job running. I wonder if we can configure the worker pool to ensure the cpus assigned to each job: https://docs.travis-ci.com/user/enterprise/worker-configuration/#configuring-the-number-of-concurrent-jobs

/assign @micw523

micw523:
Sounds good. I'll do some investigations when I have time. Also let me check some of our earlier builds to see what our configurations were.

micw523:
Our previous builds were all configured for using 2 CPUs. I wonder if Travis CI had put some quota system in place.

micw523:
See #906 - failure attributed to upstream Travis CI

PR #906, opened 2019-07-30
Request 2 CPUs per Travis CI job

micw523:
Add an environmental variable according to [this help](https://docs.travis-ci.com/user/enterprise/worker-configuration/#configuring-the-number-of-concurrent-jobs) to avoid test flakes

### DIFF
--- a/scripts/kube-init.sh
+++ b/scripts/kube-init.sh
@@ -30,6 +30,9 @@ trap "clean_exit" EXIT
 # Switch off SE-Linux
 setenforce 0
 
+# Request at least 2 CPUs from Travis CI per job
+export TRAVIS_WORKER_DOCKER_CPUS=2
+
 # Mount root to fix dns issues
 # Define $HOME since somehow this is not defined
 HOME=/home/travis


micw523:
/hold
Need to see if CI likes it

micw523:
/hold cancel
/assign @roycaihw 
Looks OK on my personal build.

scripts/kube-init.sh
```diff
@@ -30,6 +30,9 @@ trap "clean_exit" EXIT
 # Switch off SE-Linux
 setenforce 0
 
+# Request at least 2 CPUs from Travis CI per job
+export TRAVIS_WORKER_DOCKER_CPUS=2
```
roycaihw: I did some search and it seems the default value for `TRAVIS_WORKER_DOCKER_CPUS` is 2 already. I'm not sure if this is the right fix

https://github.com/travis-ci/worker/blob/b6839eb59412dd46ec2feccc85660c2afd990772/.example.env#L49-L51
https://github.com/travis-ci/worker/blob/b6839eb59412dd46ec2feccc85660c2afd990772/backend/docker.go#L60

For the original issue https://github.com/kubernetes-client/python/issues/890, have you observed other builds in this repo hitting the same error?

micw523:
I also think the default is 2. The official Travis CI documentation also says at least 2 CPUs are needed to run the jobs. Nothing started flaking until a few days ago, when only 1 CPU is assigned to some of our jobs. Today’s and yesterday’s builds are fine though - I wonder if something happened on the Travis side

micw523:
The most recent one is from #795. For some others that were failing, I restarted the individual py-functional jobs a few times and they went green - I’m not sure if we have those records...

micw523:
Ok so it’s documented here - https://travis-ci.community/t/builds-often-fail-because-of-lack-of-resources-memory-cpu/4331/3

The failure is due to Travis CI.

PR #918, opened 2019-08-06
Temporary pin minikube to 1.2.0 due to regression in 1.3.0

roycaihw:


micw523:
I'm watching kubernetes/minikube#5014. I think we are running into the same problem.

roycaihw:
Added a comment. Retagging

PR #922, opened 2019-08-08
Update examples

oz123:
This is a partial fix for #884 which addresses the first two bullet points.


oz123:
@micw523 can you be more specific please with your comment? I don't want to address all examples at once.
I just want to commit these two updated examples. 
A follow up PR will add an example for `kubernetes.util.create_from_yaml`

micw523:
I just saw that you took off the yaml example. However Sphinx allows us to create automated documentation of the utils/ folder and package with automodule/autoclass. I was wondering if that was your intention after removing the yaml example.

oz123:
@micw523 that was not my intention. This pull request is only about updating the examples. I also included them in the sphinx docs.
Long term speaking, I would like to add more documentation inside the modules,
and there is also this issue https://github.com/swagger-api/swagger-codegen/issues/9630.
Is there something specific you want me to include in this PR?

micw523:
This sounds good to me. @roycaihw what do you think?

micw523:
/lgtm
Seems like I cannot call tide by using the robot?

PR #926, opened 2019-08-13
[DO NOT MERGE] Switch to openapi-generator

roycaihw:
ref https://github.com/kubernetes-client/gen/issues/93, https://github.com/kubernetes-client/gen/pull/97

This is for reviewing and testing the diff of generated code in master branch. The change is basically:
1. pick https://github.com/kubernetes-client/gen/pull/97
2. run `scripts/update-client.sh`

but split into logical commits for easier review.

I plan to:
- do a 11.0.0a1 release with openapi-generator in release branch (and keep using swagger-codegen in master branch for now)
- switch master branch to use openapi-generator and do a 11.0.0b1 release


roycaihw:
Test is green. This is ready for review: https://github.com/kubernetes-client/python/pull/926/commits/7387639060bf05de9cb34af5919aedf52de48146

/cc @micw523 @scottilee @tomplus @yliaog 

user-facing changes:
- `kubernetes.client.apis` package is renamed to `kubernetes.client.api`
  - https://github.com/swagger-api/swagger-codegen/pull/6839#discussion_r148603571
- generated code now uses absolute import instead of relative import
- api requests won't have `Content-Type` specified in HTTP headers
- the `swagger_types` attribute in all models is renamed to `openapi_types`

non-user-facing changes:
- `kubernete/.swagger-codegen` is deleted, openapi-generator uses `kubernetes/.openapi-generator` and also `.openapi-generator-ignore` instead of `.swagger-codegen-ignore` (renamed)
- various code cleanup

micw523:
I remember ```openapi-generator``` uses flake8 as the main Python linter and that's why you see the ```#noqa``` flags scattered everywhere in the code. We may want to make the switch too in the future since these flags are not recognized by ```pycodestyle```.

The code style test is green right now - I don't think we need to act right now but I may just send in a PR switching us to flake8 when this is merged.

scripts/update-client.sh
```diff
@@ -47,6 +47,13 @@ if [[ -z ${GEN_ROOT:-} ]]; then
     GEN_ROOT="${TEMP_FOLDER}/gen"
     echo ">>> Cloning gen repo"
     git clone --recursive https://github.com/kubernetes-client/gen.git "${GEN_ROOT}"
+    # pick https://github.com/kubernetes-client/gen/pull/97
+    cd "${GEN_ROOT}"
+    git remote add tomplus https://github.com/tomplus/gen.git
+    git fetch tomplus
+    # somehow I couldn't get a clean cherrypick of the commit
```
micw523: We need to remind ourselves about removing these lines in the future

kubernetes/docs/V1beta1PodDisruptionBudgetSpec.md
```diff
@@ -3,9 +3,9 @@
...
-**selector** | [**V1LabelSelector**](V1LabelSelector.md) | Label query over pods whose evictions are managed by the disruption budget. | [optional] 
+**max_unavailable** | [**object**](.md) | An eviction is allowed if at most \&quot;maxUnavailable\&quot; pods selected by \&quot;selector\&quot; are unavailable after the eviction, i.e. even in absence of the evicted pod. For example, one can prevent all voluntary evictions by specifying 0. This is a mutually exclusive setting with \&quot;minAvailable\&quot;. | [optional] 
+**min_available** | [**object**](.md) | An eviction is allowed if at least \&quot;minAvailable\&quot; pods selected by \&quot;selector\&quot; will still be available after the eviction, i.e. even in the absence of the evicted pod.  So for example you can prevent all voluntary evictions by specifying \&quot;100%\&quot;. | [optional] 
+**selector** | [**V1LabelSelector**](V1LabelSelector.md) |  | [optional]
```
micw523: Seems like the description of a few parameters have disappeared. The loss of descriptions is very consistent across docs.

examples/exec.py
```diff
@@ -2,7 +2,7 @@
 
 from kubernetes import config
 from kubernetes.client import Configuration
-from kubernetes.client.apis import core_v1_api
+from kubernetes.client.api import core_v1_api
```
scottilee: Is this right? I get an error: `ModuleNotFoundError: No module named 'kubernetes.client.api'`.

kubernetes/README.md
```diff
@@ -17,9 +17,9 @@ Python 2.7 and 3.4+
 If the python package is hosted on Github, you can install directly from Github
 
 ```sh
-pip install git+https://github.com/kubernetes-client/python.git
+pip install git+https://github.com/kubernetes-kubernetes.client/python.git
 ```
-(you may need to run `pip` with root permission: `sudo pip install git+https://github.com/kubernetes-client/python.git`)
+(you may need to run `pip` with root permission: `sudo pip install git+https://github.com/kubernetes-kubernetes.client/python.git`)
```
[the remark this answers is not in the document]
roycaihw: ack. This exists for swagger-codegen already. I think it's in how we configure the generator somewhere.

scripts/update-client.sh
roycaihw: add a note in https://github.com/kubernetes-client/python/pull/931#issue-307594451

kubernetes/README.md
```diff
@@ -68,7 +70,7 @@ except ApiException as e:
 
 ## Documentation for API Endpoints
 
-All URIs are relative to *https://localhost*
+All URIs are relative to *http://localhost*
```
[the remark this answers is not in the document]
roycaihw: This is from the code generator. The documentation change looks reasonable to me, since by default the generated client doesn't have TLS set up

examples/exec.py
roycaihw: could you print the package info in your code, to see if it's using the updated code from this pull? I suspect you have official kubernetes package installed in your env and the code is using  that by default

kubernetes/docs/V1beta1PodDisruptionBudgetSpec.md
roycaihw: The openapi spec didn't change. @tomplus Do you know what may cause the doc change in the generators?

examples/exec.py
micw523: You can also try modifying your PYTHONPATH variable if you’re under Linux and prioritize searching for the k8s package locally.

roycaihw:
included in https://github.com/kubernetes-client/python/pull/931

/close

PR #929, opened 2019-08-13
Restore latest release of minikube

micw523:
Testing if minikube v1.3.1 fixes the CI problem.


micw523:
/assign @roycaihw 
/assign @yliaog 
Seems like minikube v1.3.1 works fine.

Issue #866, opened 2019-07-04
Unsupported Media Type (HTTP 415) when patching custom resources


nolar:
## Actual behaviour

v10.0.0 was released this night. Since then, custom resource patching is broken:

HTTP 415 "Unsupported Media Type" when patching the custom resources.

```
Traceback (most recent call last):
  File "kopf-134.py", line 15, in <module>
    rsp = api.patch_namespaced_custom_object(**request_kwargs)
  File "/Users/svasilyev/.pyenv/versions/kopf/lib/python3.7/site-packages/kubernetes/client/apis/custom_objects_api.py", line 1951, in patch_namespaced_custom_object
    (data) = self.patch_namespaced_custom_object_with_http_info(group, version, namespace, plural, name, body, **kwargs)
  File "/Users/svasilyev/.pyenv/versions/kopf/lib/python3.7/site-packages/kubernetes/client/apis/custom_objects_api.py", line 2057, in patch_namespaced_custom_object_with_http_info
    collection_formats=collection_formats)
  File "/Users/svasilyev/.pyenv/versions/kopf/lib/python3.7/site-packages/kubernetes/client/api_client.py", line 334, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/Users/svasilyev/.pyenv/versions/kopf/lib/python3.7/site-packages/kubernetes/client/api_client.py", line 168, in __call_api
    _request_timeout=_request_timeout)
  File "/Users/svasilyev/.pyenv/versions/kopf/lib/python3.7/site-packages/kubernetes/client/api_client.py", line 393, in request
    body=body)
  File "/Users/svasilyev/.pyenv/versions/kopf/lib/python3.7/site-packages/kubernetes/client/rest.py", line 286, in PATCH
    body=body)
  File "/Users/svasilyev/.pyenv/versions/kopf/lib/python3.7/site-packages/kubernetes/client/rest.py", line 222, in request
    raise ApiException(http_resp=r)
kubernetes.client.rest.ApiException: (415)
Reason: Unsupported Media Type
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'Date': 'Thu, 04 Jul 2019 11:53:38 GMT', 'Content-Length': '263'})
[...]

PR #930, opened 2019-08-13
Add a test for custom object operations

roycaihw:
pick the test from https://github.com/kubernetes-client/python/pull/886 to prevent breaking change like https://github.com/kubernetes-client/python/issues/866 from happening

### DIFF
--- a/kubernetes/client/models/v1beta1_custom_resource_definition_status.py
+++ b/kubernetes/client/models/v1beta1_custom_resource_definition_status.py
@@ -101,8 +101,6 @@ class V1beta1CustomResourceDefinitionStatus(object):
         :param conditions: The conditions of this V1beta1CustomResourceDefinitionStatus.
         :type: list[V1beta1CustomResourceDefinitionCondition]
         """
-        if conditions is None:
-            raise ValueError("Invalid value for `conditions`, must not be `None`")
 
         self._conditions = conditions
 
--- a/kubernetes/e2e_test/test_client.py
+++ b/kubernetes/e2e_test/test_client.py
@@ -19,6 +19,8 @@ import uuid
 
 from kubernetes.client import api_client
 from kubernetes.client.apis import core_v1_api
+from kubernetes.client.apis import apiextensions_v1beta1_api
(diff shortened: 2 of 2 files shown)


kubernetes/e2e_test/test_client.py
```diff
@@ -234,3 +236,63 @@ def test_node_apis(self):
...
+            body={'kind':'Foo', 'apiVersion':api_version, 'metadata':{'name':name}})
+        print('E2E test CR created')
+
+        # perform an emtpy JSON merge patch on the custom object
+        resp = cr_api.patch_namespaced_custom_object(group=group,
+            version='v1',
+            plural='foos',
+            namespace='default',
+            name=name,
+            body={})
```
yliaog: do you want to check the resp? does it succeed?

roycaihw: added `.metadata.name` check to make sure that the response is a CRD, not a metav1.Status

also added a commit to make the CRD deserializer usable before we fix the upstream openapi spec: https://github.com/kubernetes/kubernetes/pull/64996

kubernetes/client/models/v1beta1_custom_resource_definition_status.py
```diff
@@ -101,8 +101,6 @@ def conditions(self, conditions):
         :param conditions: The conditions of this V1beta1CustomResourceDefinitionStatus.
         :type: list[V1beta1CustomResourceDefinitionCondition]
         """
-        if conditions is None:
-            raise ValueError("Invalid value for `conditions`, must not be `None`")
```
yliaog: this is generated, and will be overwritten in next release, right?

roycaihw: yes, the proper fix will come in 1.16 https://github.com/kubernetes/kubernetes/pull/64996

somehow in the test, the response is deserialized into a dict instead of a CRD object, which causes the attribute assertion to fail. Perhaps this workaround isn't ideal either. I will abandon this pull and verify the CR patch methods manually in the next couple of releases

/close

PR #935, opened 2019-08-19
fix typo

hacker-h:


micw523:
I think this needs to be fixed in the main repo instead. The document you're looking at is automatically generated.

roycaihw:
agree this should be fixed in upstream openapi spec

/close

Issue #812, opened 2019-04-11
How do you get exit code connect_get_namespaced_pod_exec ?


oz123:
Currently I have to simply do this:

```
        response = stream(self.api.connect_get_namespaced_pod_exec,
                          podname, 'kube-system',
                          command=exec_command,
                          stderr=False, stdin=False,
                          stdout=True, tty=False)

        if not response:
            do_something__()
```

However, I can't get the error code, or the stderror.
Looking at the source code I don't see an easy way.

tomplus:
@oz123 The error code arrives in the channel number 3 (error-channel) but it is not exposed by the stream() method.

As a workaround you can use:
```
client = stream(api.connect_get_namespaced_pod_exec,
                "myapp-78898448f4-pgnfl",
                "default",
                command=["false"],                  # <<< "false" returns with exit code = 1
                stderr=True, stdin=False,
                stdout=True, tty=False,
                _preload_content=False)

client.run_forever(timeout=60)
err = client.read_channel(ERROR_CHANNEL)

pprint(yaml.load(err))
```

and you should get
```
{'details': {'causes': [{'message': '1', 'reason': 'ExitCode'}]},
 'message': 'command terminated with non-zero exit code: Error executing in '
            'Docker Container: 1',
 'metadata': {},
 'reason': 'NonZeroExitCode',
 'status': 'Failure'}
```

I'm going to add a method to get it in more convenient way.

phracek:
CC @TomasTomecek

roycaihw:
fixed by https://github.com/kubernetes-client/python-base/pull/160

/close

sunchill06:
@tomplus 

When I run this as

resp = stream(v1.connect_get_namespaced_pod_exec, podName, namespace, command=exec_command, stderr=True, stdin=False, stdout=True, tty=False, _preload_content=preloadContent)

resp.run_forever(timeout=5)
print(resp.readline_channel(1))

Response:
**None**

However, print(resp.read_channel(1)) gives me proper output.

Any pointers why?

PR #956, opened 2019-09-17
Test getting the returncode of execution in a pod

oz123:
This tests demonstrate how to execute a command in a pod and
what behavior is expected. As discussed in the commit
https://github.com/kubernetes-client/python-base/pull/160/commits/bf367ed6ddc63369f76df0a07b248a6711328605 in python-base this
behavior would be familiar to Python users, as `subprocess.Popen`
has the same property.

Also, it's worth mentioning #812 which this PR is a suggested fix for.

### DIFF
--- a/kubernetes/e2e_test/test_client.py
+++ b/kubernetes/e2e_test/test_client.py
@@ -29,6 +29,26 @@ def short_uuid():
     return id[-12:]
 
 
+def manifest_with_command(name, command):
+    return {
+        'apiVersion': 'v1',
+        'kind': 'Pod',
+        'metadata': {
+            'name': name
+        },
+        'spec': {
+            'containers': [{
+                'image': 'busybox',
+                'name': 'sleep',
+                "args": [
+                    "/bin/sh",
+                    "-c",
+                    command
+                ]
+            }]
+        }
+    }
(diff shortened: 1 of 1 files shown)


kubernetes/e2e_test/test_client.py
```diff
@@ -117,6 +119,45 @@ def test_pod_apis(self):
...
+        )
+        for command, value in commands_expected_values:
+            client = stream(api.connect_get_namespaced_pod_exec, name, 'default',
+                                                       command=command,
+                                                       stderr=True, stdin=False,
+                                                       stdout=True, tty=False,
+                                                       _preload_content=False)
+
+            self.assertIsNone(client.returncode)
+            client.run_forever(timeout=60)
```
roycaihw: could you make the timeout smaller since the commands don't take long? maybe 10s? 

LGTM overall. Please rebase the python-base submodule after https://github.com/kubernetes-client/python-base/pull/160 is merged

oz123: Yes, sure. I will do this 10 sec. Thanks for merging  kubernetes-client/python-base#160 is merged
 so quickly.

PR #959, opened 2019-09-22
content-type for patch commands

tomplus:
Hi all.

In this PR I'd like to discuss possible solution for problems with selecting content-type when K8s' objects are patching. This problem is nicely documented in https://github.com/kubernetes-client/python/issues/866 and https://github.com/tomplus/kubernetes_asyncio/issues/68

This idea is to introduce a new parameter in ApiClient to be able to control a return from the `select_header_content_type` method and remove the problematic detection from rest.py. Optionally we can add this as an argument to each patch method.
It'll be breaking change if an application uses patch_* methods.

Please take a look. If we accept the solution I prepare PR in openapi-generator and temporary patch for this repo.

cc: @roycaihw @yliaog @HaraldGustafsson @nolar 

Thanks.

### DIFF
--- a/examples/custom_object.py
+++ b/examples/custom_object.py
@@ -29,6 +29,19 @@ spec:
     - name: v1
       served: true
       storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                cronSpec:
+                  type: string
+                image:
+                  type: string
+                replicas:
+                  type: integer
   scope: Namespaced
   names:
     plural: crontabs
@@ -68,6 +81,28 @@ def main():
         body=my_resource,
     )
     print("Resource created")
+
+    # get the resource and print out data
+    resource = api.get_namespaced_custom_object(
+        group="stable.example.com",
+        version="v1",
+        name="my-new-cron-object",
+        namespace="default",
+        plural="crontabs",
+    )
+    print("Resource details:")
+    pprint(resource)
+
+    # patch the resource
+    resource = api.patch_namespaced_custom_object(
+        group="stable.example.com",
+        version="v1",
+        name="my-new-cron-object",
+        namespace="default",
+        plural="crontabs",
+        body={'image': 'my-awesome-cron-image:2'},
+        content_type="application/merge-patch+json"
+    )
 
     # get the resource and print out data
     resource = api.get_namespaced_custom_object(
--- a/kubernetes/client/api/custom_objects_api.py
+++ b/kubernetes/client/api/custom_objects_api.py
@@ -2336,6 +2336,7 @@ class CustomObjectsApi(object):
         all_params.extend(
             [
                 'async_req',
+                'content_type',
                 '_return_http_data_only',
                 '_preload_content',
                 '_request_timeout'
@@ -2404,8 +2405,8 @@ class CustomObjectsApi(object):
             ['application/json'])  # noqa: E501
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.select_header_content_type(  # noqa: E501
-            ['application/merge-patch+json'])  # noqa: E501
+        header_params['Content-Type'] = local_var_params.get('content_type', self.api_client.\
+            select_header_content_type(['application/merge-patch+json'], 'PATCH', body_params))
 
         # Authentication setting
         auth_settings = ['BearerToken']  # noqa: E501
@@ -2505,6 +2506,7 @@ class CustomObjectsApi(object):
         all_params.extend(
             [
                 'async_req',
+                'content_type',
                 '_return_http_data_only',
                 '_preload_content',
                 '_request_timeout'
@@ -2573,8 +2575,8 @@ class CustomObjectsApi(object):
             ['application/json', 'application/yaml', 'application/vnd.kubernetes.protobuf'])  # noqa: E501
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.select_header_content_type(  # noqa: E501
-            ['application/merge-patch+json'])  # noqa: E501
+        header_params['Content-Type'] = local_var_params.get('content_type', self.api_client.\
+            select_header_content_type(['application/merge-patch+json'], 'PATCH', body_params))
 
         # Authentication setting
         auth_settings = ['BearerToken']  # noqa: E501
@@ -2674,6 +2676,7 @@ class CustomObjectsApi(object):
         all_params.extend(
             [
                 'async_req',
+                'content_type',
                 '_return_http_data_only',
                 '_preload_content',
                 '_request_timeout'
@@ -2742,8 +2745,8 @@ class CustomObjectsApi(object):
             ['application/json', 'application/yaml', 'application/vnd.kubernetes.protobuf'])  # noqa: E501
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.select_header_content_type(  # noqa: E501
-            ['application/merge-patch+json'])  # noqa: E501
+        header_params['Content-Type'] = local_var_params.get('content_type', self.api_client.\
+            select_header_content_type(['application/merge-patch+json'], 'PATCH', body_params))
 
         # Authentication setting
         auth_settings = ['BearerToken']  # noqa: E501
@@ -2846,6 +2849,7 @@ class CustomObjectsApi(object):
         all_params.extend(
             [
                 'async_req',
+                'content_type',
                 '_return_http_data_only',
                 '_preload_content',
                 '_request_timeout'
@@ -2920,8 +2924,8 @@ class CustomObjectsApi(object):
             ['application/json'])  # noqa: E501
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.select_header_content_type(  # noqa: E501
-            ['application/merge-patch+json'])  # noqa: E501
+        header_params['Content-Type'] = local_var_params.get('content_type', self.api_client.\
+            select_header_content_type(['application/merge-patch+json'], 'PATCH', body_params))
 
         # Authentication setting
         auth_settings = ['BearerToken']  # noqa: E501
@@ -3024,6 +3028,7 @@ class CustomObjectsApi(object):
         all_params.extend(
             [
                 'async_req',
+                'content_type',
                 '_return_http_data_only',
                 '_preload_content',
                 '_request_timeout'
@@ -3098,8 +3103,8 @@ class CustomObjectsApi(object):
             ['application/json', 'application/yaml', 'application/vnd.kubernetes.protobuf'])  # noqa: E501
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.select_header_content_type(  # noqa: E501
-            ['application/merge-patch+json'])  # noqa: E501
+        header_params['Content-Type'] = local_var_params.get('content_type', self.api_client.\
+            select_header_content_type(['application/merge-patch+json'], 'PATCH', body_params))
 
         # Authentication setting
         auth_settings = ['BearerToken']  # noqa: E501
@@ -3202,6 +3207,7 @@ class CustomObjectsApi(object):
         all_params.extend(
             [
                 'async_req',
+                'content_type',
                 '_return_http_data_only',
                 '_preload_content',
                 '_request_timeout'
@@ -3276,8 +3282,8 @@ class CustomObjectsApi(object):
             ['application/json', 'application/yaml', 'application/vnd.kubernetes.protobuf'])  # noqa: E501
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.select_header_content_type(  # noqa: E501
-            ['application/merge-patch+json'])  # noqa: E501
+        header_params['Content-Type'] = local_var_params.get('content_type', self.api_client.\
+            select_header_content_type(['application/merge-patch+json'], 'PATCH', body_params))
 
         # Authentication setting
         auth_settings = ['BearerToken']  # noqa: E501
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -494,7 +494,7 @@ class ApiClient(object):
         else:
             return ', '.join(accepts)
 
-    def select_header_content_type(self, content_types):
+    def select_header_content_type(self, content_types, method=None, body=None):
         """Returns `Content-Type` based on an array of content_types provided.
 
         :param content_types: List of content-types.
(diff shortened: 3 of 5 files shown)


kubernetes/e2e_test/test_client.py
```diff
@@ -55,9 +55,12 @@ class TestClient(unittest.TestCase):
     def setUpClass(cls):
         cls.config = base.get_e2e_configuration()
 
+    def setUp(self):
+        self.client = api_client.ApiClient(configuration=self.config,
+                                           prefered_content_type="application/strategic-merge-patch+json")
```
tomplus: We can use it as a default to make it backward compatible.

HaraldGustafsson: Assume you meant to have it as default in the ApiClient?

roycaihw: IIUC, defaulting to one content type doesn't make it backward compatible, since the old behavior was picking one of the two content types based on the `body` type. We have to keep the old behavior if we want to avoid breaking the old clients

examples/custom_object.py
```diff
@@ -46,7 +46,9 @@
 def main():
     config.load_kube_config()
 
-    api = client.CustomObjectsApi()
+    # passing prefered_content_type to choose the required type of patch
+    api_client = client.ApiClient(prefered_content_type="application/merge-patch+json")
+    api = client.CustomObjectsApi(api_client)
```
roycaihw: What if someone wants to send different types of patch in their client? Is it possible to expose an argument in each API to specify the content type for each request?

kubernetes/e2e_test/test_client.py
tomplus: IMO backward compatibility is not possible, because current behavior doesn't work for custom resources. If we assume that strategic-merg-patch is more common we can set is as a default. Other clients will have to change the preferred content type only...

examples/custom_object.py
tomplus: Sure, we can add it to each patch method but I thought that something like the default-content-type could help to write more consistent code. I've never mixed type of patch in my code yet.

kubernetes/e2e_test/test_client.py
roycaihw: > IMO backward compatibility is not possible, because current behavior doesn't work for custom resources

the current behavior allows JSON patch for custom resources, right? Can we keep the following behavior if `prefered_content_type` is None? 
1. for native types, a client can send `json-patch+json` if the body is a list, or `strategic-merge-patch+json` if the body is a map
2. for custom resources, a client can send `json-patch+json` (a list) only (apiserver doesn't support strategic merge patch for custom resources)

In future, we need to allow a client to send `merge-patch+json` and `apply-patch+yaml` for both native types and custom resources as well

examples/custom_object.py
roycaihw: I'm okay with either way

kubernetes/client/api_client.py
```diff
@@ -494,6 +497,8 @@ def select_header_content_type(self, content_types):
 
         if 'application/json' in content_types or '*/*' in content_types:
             return 'application/json'
+        elif self.prefered_content_type in content_types:
+            return self.prefered_content_type
```
roycaihw: This solves our problem. I'm not sure if this is general for other project as well. It works for us because PATCH is so far the only verb that doesn't take `application/json` and `*/*` as content type.

kubernetes/client/api_client.py
```diff
@@ -74,7 +74,10 @@ def __init__(self, configuration=None, header_name=None, header_value=None,
         self.cookie = cookie
         # Set default User-Agent.
         self.user_agent = 'Swagger-Codegen/10.0.0-snapshot/python'
-    
+
+        # prefered content-type if multiple options are allowed (eg. patch command)
```
roycaihw: 1) if multiple options are allowed and 2) if `application/json` is not allowed?

kubernetes/e2e_test/test_client.py
tomplus: Sure, I prepare changes for such approach.

I see one drawbacks - we'll keep the "magic" logic in rest.py and it won't be such generic to add it to the open-generator. But let's discuss it later with the code.

Thanks.

HaraldGustafsson: if the `select_header_content_type` takes the body and method as well then it can be intelligent about it's choice. We should be prepared for that custom resources will support strategic merge in the future when such strategy is defined in the CRD. So it is still a good idea to introduce the option to set the preferred type. 

Note that the previous supported type for custom resources was `application/merge-patch+json` and not `json-patch+json` which was newly added. So if you should have a default for custom objects you should select `application/merge-patch+json` if you want backward compatibility.

roycaihw: > Note that the previous supported type for custom resources was application/merge-patch+json and not json-patch+json which was newly added. So if you should have a default for custom objects you should select application/merge-patch+json if you want backward compatibility

Ah yes, you're right. Thanks!

tomplus: Now I’m leaning towards to expose content_type to each request (for patch mainly) to bypass the select_content_type function. This function now accepts "method" and "body" to try to select the better header. K8s introduces strategic-merge-patch which is k8s specific and this part has to be patched - instead of patching rest.py. Let me know what you think. Thanks.

examples/custom_object.py
tomplus: I've dropped `preferred_content _type` option - it didn't support situation where different type of methods have different valid content types.

kubernetes/client/api_client.py
tomplus: I've removed prefered_content_type.

kubernetes/client/api_client.py
tomplus: I've removed prefered_content_type.

examples/custom_object.py
```diff
@@ -78,6 +78,28 @@ def main():
...
 
+    # patch the resource
+    resource = api.patch_namespaced_custom_object(
+        group="stable.example.com",
+        version="v1",
+        name="my-new-cron-object",
+        namespace="default",
+        plural="crontabs",
+        body={'image': 'my-awesome-cron-image:2'},
+        content_type="application/merge-patch+json"
```
tomplus: This parameter is not necessary - select_header_content_type can manage with it.

kubernetes/client/apis/custom_objects_api.py
```diff
@@ -1656,8 +1657,8 @@ def patch_cluster_custom_object_with_http_info(self, group, version, plural, nam
             select_header_accept(['application/json'])
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.\
-            select_header_content_type(['application/merge-patch+json'])
+        header_params['Content-Type'] = params.get('content_type', self.api_client.\
+            select_header_content_type(['application/merge-patch+json']), 'PATCH', body_params)
```
HaraldGustafsson: This all started due to that we added 'application/json-patch+json' for custom objects, so should that not be part of the list of available content types?

HaraldGustafsson: Also assuming this is an example on the changes to all the patch methods, since there are several more in this file.

tomplus: Yes, should be but this PR uses current code from master branch (https://github.com/kubernetes-client/python/blob/master/kubernetes/client/apis/custom_objects_api.py#L1659). AFAIK @roycaihw removed this option when the problem appeared.

tomplus: Yes, I use this PR to discuss final solution and then I will change the openapi-generator to add this to all methods.

tomplus:
@roycaihw Could you take a look at the last concept? I'd like to prepare generator to manage with it. Thanks.

examples/custom_object.py
roycaihw: Custom resources support JSON patch, so this parameter can be used to hint JSON patch / JSON merge patch, right?

kubernetes/client/api_client.py
```diff
@@ -492,6 +492,21 @@ def select_header_content_type(self, content_types):
...
+                if isinstance(body, list):
+                    return 'application/json-patch+json'
+                else:
+                    # ---
+                    # kubernetes specific - application/strategic-merge-patch+json
+                    # applied as a patch or subclass
+                    if 'application/strategic-merge-patch+json' in content_types:
+                        return 'application/strategic-merge-patch+json'
+                    # ---
+                    return 'application/merge-patch+json'
```
roycaihw: Most native types support both strategic merge patch and JSON merge patch. Just to clarify, if a client wants to send JSON merge patch, it has to use the `content_type` parameter to hint, right?

kubernetes/client/apis/core_v1_api.py
```diff
@@ -14792,8 +14793,8 @@ def patch_namespaced_config_map_with_http_info(self, name, namespace, body, **kw
             select_header_accept(['application/json', 'application/yaml', 'application/vnd.kubernetes.protobuf'])
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.\
-            select_header_content_type(['application/json-patch+json', 'application/merge-patch+json', 'application/strategic-merge-patch+json'])
+        header_params['Content-Type'] = params.get('content_type', self.api_client.\
+            select_header_content_type(['application/json-patch+json', 'application/merge-patch+json', 'application/strategic-merge-patch+json'], 'PATCH', body_params))
```
roycaihw: Kubernetes recently added a new type `application/apply-patch+yaml`. However it's likely hard to auto-detect the yaml type based on the body

kubernetes/client/apis/custom_objects_api.py
roycaihw: I removed 'application/json-patch+json' from custom objects because the existing select_header_content_type couldn't support it without breaking old clients. With this change from @tomplus, we should be able to add 'application/json-patch+json' for custom objects

lizfeed:
Any update on this? this would be really handy for me.

examples/custom_object.py
tomplus: Yes, exactly.

kubernetes/client/api_client.py
tomplus: Yes, it's hard to detect which type of patch has been sent so it's better to expose content_type explicit.

examples/custom_object.py
```diff
@@ -29,6 +29,19 @@
     - name: v1
       served: true
       storage: true
+      schema:
```
tomplus, line 32: The schema is now required by CRD.

tomplus:
I'm back to this PR. I had to squash previous commits and rebase with the master branch (it was very old, before openapi-generator). I hope it helps to discuss the idea. This PR won't be merge because some changes have to be added to the generator. It's like an example how it will look like.

Thanks.

floriankoch:
Hi,
is there any update here? 

regards

Jc2k:
It would be really nice if someone could look at @tomplus's work. I seem to have to monkey patch around this every time i want to use this library.

krichter722:
The PR is still marked `[WIP]` in the title.

Jc2k:
So I read the last comment from the author as being about getting feedback from upstream about the direction/approach? My bad 😄

tomplus:
@krichter722 It's marked WIP because it shouldn't be merged in this form. I'm waiting for reviewing this PR to prepare changes in OpenAPI Generator

cc: @roycaihw

examples/custom_object.py
arcivanov: What about `application/apply-patch+yaml`? Why aren't we allowing to override from the caller?

tomplus:
I know it needs rebase but it's only to show and discuss the idea. At the end the generator will be extended.

@yliaog @roycaihw please take a look.

/hold

kubernetes/client/api_client.py
```diff
@@ -505,6 +505,21 @@ def select_header_content_type(self, content_types):
 
         content_types = [x.lower() for x in content_types]
 
+        if method and body and method == 'PATCH':
+            # try to select valid content_type
+            if 'application/json-patch+json' in content_types and \
+               'application/merge-patch+json' in content_types:
```
yliaog, line 511: what about changing  'application/merge-patch+json' in content_types: to

('application/merge-patch+json' in content_types or 'application/strategic-merge-patch+json' in content_types):

kubernetes/client/api/custom_objects_api.py
```diff
@@ -2573,8 +2575,8 @@ def patch_cluster_custom_object_scale_with_http_info(self, group, version, plura
             ['application/json', 'application/yaml', 'application/vnd.kubernetes.protobuf'])  # noqa: E501
 
         # HTTP header `Content-Type`
-        header_params['Content-Type'] = self.api_client.select_header_content_type(  # noqa: E501
-            ['application/merge-patch+json'])  # noqa: E501
+        header_params['Content-Type'] = local_var_params.get('content_type', self.api_client.\
+            select_header_content_type(['application/merge-patch+json'], 'PATCH', body_params))
```
yliaog, line 2579: want to double check, this PR would be backward compatible for both core native types, and custom resources, right?

kubernetes/client/api_client.py
tomplus: Good point. It should be implemented in this way.

kubernetes/client/api/custom_objects_api.py
tomplus: It's tricky question, in general it's backward compatible, but I can imagine that some calls can be interpreted differently by select_header_content_type - the current version usually uses the first patch method but this version tries to guess the content-type and results may be different... To be honest I don't think it's a real problem.

The positive aspect is that we will be able to stop removing the first element from the list as in this change https://github.com/roycaihw/client-python/commit/61f1fbe49af3be34cebe8781efa436ae8b84e2e2 and select_header_content_type should do the job.

yliaog: sounds good to me, @roycaihw do you have any concern?

roycaihw: +1 to it's not a real problem (at least for this repo, given how we use `select_header_content_type` today).

Today native types can only use `application/json-patch+json` or `application/strategic-merge-patch+json`. With this change, existing applications will still get the same result. New applications can use the exposed `content_type` to choose using `application/merge-patch+json`.

Custom resources today can only use `application/merge-patch+json` and will get the same result. With this change, applications will be able to pass in `application/json-patch+json` body.

kubernetes/client/api_client.py
roycaihw: Anything related to `application/strategic-merge-patch+json` will be a patch maintained by us IIUC

roycaihw:
Thanks for putting together this! It's a huge improvement to this client and can also benefit other generated clients. LGTM

tomplus:
My update - Openapi-generator already has required changes (PR #[10686](https://github.com/OpenAPITools/openapi-generator/pull/10686) & [10978](https://github.com/OpenAPITools/openapi-generator/pull/10978)) in the legacy and new python generator. It is in the master branch and will be released in v5.3.1. We can continue work on it when we switch to the latest version of generator - it's addressed here https://github.com/kubernetes-client/python/issues/1589.

aslafy-z:
So, we switched to golang..

PR #966, opened 2019-09-24
Address API Changes Introduced in k8s v1.16

### LINKED ISSUE #963: Travis CI is failing with Kubernetes 1.16
e.g. 
- https://github.com/kubernetes-client/python-base/pull/163
- https://travis-ci.org/kubernetes-client/python-base/jobs/588553023

1. dynamic client is failing with the following error, we should specify the version (`v1beta1`) in the test
```
ResourceNotUniqueError: Multiple matches found for {'kind': 'CustomResourceDefinition'}: [<Resource(apiextensions.k8s.io/v1beta1/customresourcedefinitions)>, <Resource(apiextensions.k8s.io/v1/customresourcedefinitions)>]
```
2. utils test is failing creating apps/extensions APIs with the following error, but I haven't seen upstream API change that could cause this
```
FailToCreateError: Error from server (Not Found): {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"the server could not find the requested resource","reason":"NotFound","details":{},"code":404}
```

micw523:
Fixes #963 .
This PR:
- Upgrades all apps/v1beta1 yamls to apps/v1. Ref https://github.com/kubernetes/kubernetes/pull/70672
- Changes tests originally written for extensions/v1beta1 to apps/v1.
- Add submodule commits for the exit code and dynamic client to fix e2e test errors.

### DIFF
--- a/kubernetes/e2e_test/test_apps.py
+++ b/kubernetes/e2e_test/test_apps.py
@@ -17,12 +17,12 @@ import uuid
 import yaml
 
 from kubernetes.client import api_client
-from kubernetes.client.apis import extensions_v1beta1_api
+from kubernetes.client.apis import apps_v1_api
 from kubernetes.client.models import v1_delete_options
 from kubernetes.e2e_test import base
 
 
-class TestClientExtensions(unittest.TestCase):
+class TestClientApps(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
@@ -30,14 +30,17 @@ class TestClientExtensions(unittest.TestCase):
 
     def test_create_deployment(self):
         client = api_client.ApiClient(configuration=self.config)
-        api = extensions_v1beta1_api.ExtensionsV1beta1Api(client)
+        api = apps_v1_api.AppsV1Api(client)
         name = 'nginx-deployment-' + str(uuid.uuid4())
-        deployment = '''apiVersion: extensions/v1beta1
+        deployment = '''apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: %s
 spec:
   replicas: 3
+  selector:
+    matchLabels:
+      app: nginx
   template:
     metadata:
       labels:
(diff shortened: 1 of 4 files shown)


kubernetes/e2e_test/test_extensions.py
```diff
@@ -1,94 +0,0 @@
...
-import uuid
-import yaml
-
-from kubernetes.client import api_client
-from kubernetes.client.apis import extensions_v1beta1_api
-from kubernetes.client.models import v1_delete_options
-from kubernetes.e2e_test import base
-
-
-class TestClientExtensions(unittest.TestCase):
```
roycaihw: do we have equivalent test for apps API?

kubernetes/e2e_test/test_apps.py
```diff
@@ -17,35 +17,37 @@
...
         name = 'nginx-deployment-' + str(uuid.uuid4())
-        deployment = '''apiVersion: extensions/v1beta1
+        deployment = '''apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: %s
 spec:
   replicas: 3
+  selector:
+	matchLabels:
```
roycaihw: nit: fix indentation

kubernetes/e2e_test/test_apps.py
```diff
@@ -56,28 +58,33 @@ def test_create_deployment(self):
         self.assertIsNotNone(resp)
 
         options = v1_delete_options.V1DeleteOptions()
-        resp = api.delete_namespaced_deployment(name, 'default', body=options)
+        resp = api.delete_namespaced_deployment(name, 'default', body={})
```
roycaihw: why don't we keep the v1 DeleteOptions?

kubernetes/e2e_test/test_apps.py
micw523: Fixed. The original indentation looked that way.

kubernetes/e2e_test/test_apps.py
micw523: It's optional - I kept it just in case. I think sometime in the past, body was not optional for the delete methods so that was necessary. Now it is not.

PR #969, opened 2019-09-25
support false values and missing fields in configs

goddenrich:
updated the python-base to include support for false values and missing fields in config files


micw523:
ref kubernetes-client/python-base#163

PR #975, opened 2019-10-02
Update py.test to pytest

micw523:
Refer to [this stackoverflow post](https://stackoverflow.com/questions/39495429/py-test-vs-pytest-command).

This should provide identical pytest coverage while allowing us to stay modern :)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ deps = -r{toxinidir}/test-requirements.txt
        -r{toxinidir}/requirements.txt
 commands =
    python -V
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -22,22 +22,22 @@ commands =
 [testenv:py27-functional]
 commands =
    python -V
-   {toxinidir}/scripts/kube-init.sh py.test -vvv -s []
```
roycaihw: should we update L12 as well?

micw523: Nice catch - updated!

roycaihw:
/approve

please squash the commits

PR #981, opened 2019-10-09
Merge azure refresh fix from python-base

fooka03:
This code replaces the hard-coded microsoft graph apiserver id with an attempt to load it dynamically from the kubeconfig before falling back to the graph id to maintain backwards compatibility.


fooka03:
Bump, @roycaihw or @yliaog are you able to review?

Issue #977, opened 2019-10-04
Convert JSON or Dict to client objects


remram44:
See also https://github.com/kubernetes-client/python/issues/340#issuecomment-526249075 (was closed for inactivity, NOT resolved)
See also #63

**Motivation** 
It is sometimes useful to be able to load Python objects from JSON, for example to read it from a config file. It should be possible to get Python objects from JSON, to pass around the code in a type-safe manner, for modification and ultimately sending through the client.

**Feature request** 
*There should be functions to convert JSON to objects and objects to JSON. This code already exists (it's required to talk to the API server), it should be exposed.*

**Workaround?** 
Right now, you can put this raw JSON into Python objects' fields, and the API client will apparently accept this, however if you use such mixed objects your code will probably break since their attribute names are different (API uses PascalCase, Python objects use snake_case). Even if taking care to handle that, you will reach the point where you have to handle things coming from the API client (Python objects using snake_case) differently from what you load from config files (JSON using PascalCase). Really messy.

Right now I am left considering dropping this lib altogether and issuing requests directly, since the point of the wrapper is to expose safe typed Python classes instead of the raw JSON, but this doesn't work if you use any kind of config file.

**Current status** 
It is somewhat possible to convert from a Python object to JSON using `client.api_client.sanitize_for_serialization()`, however converting from JSON to Python is completely impossible; the `deserialize()` method that exists expects a `requests.Response`, not JSON. Calling `client.api_client._ApiClient__deserialize()` works, but that is a private method, and you need to pass it the `kind` read from the JSON yourself.

Apologies for duplicating #340 but I cannot re-open it myself.

chekolyn:
I was just looking trying to do this. This would be a useful function.

yliaog:
@chekolyn could you take this issue, and work on a pr?

chekolyn:
@yliaog sure, I'll give it a shot. i think the deserializer() might have been auto generated by swagger codegen. So where would this be appropriate? utils?

micw523:
Yes that looks like generated code. I think utils should be fine, but somewhere in the base repo may work as well since we don’t necessarily require e2e tests on this one (do we?)

chekolyn:
I've been doing some tests and using a similar approach to [utils.create_from_yaml ](https://github.com/kubernetes-client/python/blob/master/kubernetes/utils/create_from_yaml.py)to infer the object  response_type to use in the already generated codegen deserialize function.

[...]

PR #989, opened 2019-10-19
Adding utils.deserialize

chekolyn:
Allows loading objects from json and yaml files.
See for feature request https://github.com/kubernetes-client/python/issues/977

Please review, this is my first pr.


micw523:
the tests would not work when you specify clusters on a local network like that - please use the e2e configurations.

Also, is it necessary to add several thousand lines of test cases? It’s very difficult to review like that...

chekolyn:
Sorry for all the lines, i just wanted to add some json files generated via kubectl and exported to json via command line.
It seems that it did catch an issue with python2.7. 
I looked into it and it seems the json.load is having issues with some json files; I have not found the root cause, but I have a hunch it might be unicode related.

chekolyn:
Hey guys, please let me know what else is necessary for this PR. I've reduce the number of json tests and the number of json files. Thank you.

chekolyn:
@roycaihw Sorry to bug, but I was wondering how can we move this PR forward.
If size is of the PR is an issue; I can continue to delete some of the test json files, but that would reduce some of the test cases.

kubernetes/utils/deserialize.py
```diff
@@ -0,0 +1,285 @@
...
+    or string of class name.
+    verbose: If True, print additional info.
+        Default is False.
+    """
+    # Mock api response from dict
+    resp_mock = RespMock()
+    resp_mock.data = json.dumps(data)
+
+    # Infer response_type from json data when not provided
+    if not klass:
```
roycaihw: I think having the caller to pass in the klass is the simplest and satisfies most use cases.

kubernetes/utils/deserialize.py
```diff
@@ -0,0 +1,285 @@
...
+    Raises:
+        FailToLoadError which holds list of load failures.
+    """
+    # If it is a list type, will need to iterate its items
+    load_exceptions = []
+    obj = None
+    if "List" in data["kind"]:
+        obj_list = []
+        # Could be "List" or "Pod/Service/...List"
+        # This is a list type. iterate within its items
```
roycaihw: I think api_client.deserialize can handle list types, e.g. `V1DeploymentList`. Why do we need to break the list down?

kubernetes/utils/deserialize.py
```diff
@@ -0,0 +1,285 @@
...
+        for yml_document in yml_document_all:
+            try:
+                obj = load_from_dict(
+                    yml_document, klass=klass, verbose=verbose)
+                # Prevent returning list of lists when doing multi
+                # document yaml
+                if isinstance(obj, list):
+                    objs.extend(obj)
+                else:
+                    objs.append(obj)
```
roycaihw: Could you justify the use case for deserializing a yaml list into a list of objects?

kubernetes/utils/deserialize.py
chekolyn: I'll trying to target both use-cases. The one we pass klass and the auto-inferring. I will prioritize the first one. For auto-inferring I am taking inspiration from the watch implementation.

kubernetes/utils/deserialize.py
chekolyn: I think this an excellent observation; I will explore this.

kubernetes/utils/deserialize.py
chekolyn: I don't think I can justify it if we can handle list type as you suggested above. I will explore that option first and this shall be removed if that is implemented.

kubernetes/utils/deserialize.py
```diff
@@ -0,0 +1,285 @@
...
+    # Take care for the case e.g. api_type is "apiextensions.k8s.io"
+    # Only replace the last instance
+    group = "".join(group.rsplit(".k8s.io", 1))
+    # convert group name from DNS subdomain format to
+    # python class name convention
+    group = "".join(word.capitalize() for word in group.split("."))
+    fcn_to_call = "{0}{1}Api".format(group, version.capitalize())
+    k8s_api = getattr(client, fcn_to_call)
+    # Replace CamelCased action_type into snake_case
+    kind = data["kind"]
```
roycaihw: The dict can be either a list of objects or a object, e.g. the following two types of responses
```
$ kubectl get po -ojson
{
    "apiVersion": "v1",
    "items": [
        {
            "apiVersion": "v1",
            "kind": "Pod",
            "metadata": {
...
```
v.s.
```
$ kubectl get po nginx-deployment-54f57cf6bf-cs6t7 -ojson
{
    "apiVersion": "v1",
    "kind": "Pod",
    "metadata": {
...
```
in the first case, the lookup function should be `list_namespaced_pod_with_http_info`

kubernetes/utils/deserialize.py
```diff
@@ -0,0 +1,285 @@
...
+    kind = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", kind)
+    kind = re.sub("([a-z0-9])([A-Z])", r"\1_\2", kind).lower()
+    # Expect the user to load namespaced objects more often
+    if hasattr(k8s_api, "read_namespaced_{0}_with_http_info".format(kind)):
+        fnc_lookup = "read_namespaced_{0}_with_http_info".format(kind)
+    elif hasattr(k8s_api, "read_{0}_with_http_info".format(kind)):
+        fnc_lookup = "read_{0}_with_http_info".format(kind)
+    # Try with the create if no read fnc found
```
roycaihw: is there any API that supports CREATE but doesn't support GET?

kubernetes/utils/deserialize.py
chekolyn: I think returning ObjectsList will work however there is one edge case where we need to account for multiple types of objects.
The following json  can have multiple objects with different kinds.

```kubectl get all --all-namespaces -o json > all.objs.json```

I guess in this case we need to traverse the list items and extract the different kinds of objects then we can output a list of ObjectsList. I think that should be the only time we return a python list with multiple K8s ObjectsList.

File: kubernetes/utils/deserialize.py (hunk 5 above)
Comment (chekolyn) [reply kubernetes/utils/deserialize.py]: I think this is for objects that don't have read_xxx  functions; I believe the get_{0}_with_http_info would be more appropriate to avoid confusion.

File: kubernetes/utils/deserialize.py (hunk 2 above)
Comment (roycaihw) [reply kubernetes/utils/deserialize.py]: `kubectl get all` sends multiple requests to the apiserver (you can see them with `kubectl get all -v=6`). This python client doesn't have a `get all` utility. If it does, for each request the response will be a single ObjectsList.

File: kubernetes/utils/deserialize.py (hunk 5 above)
Comment (roycaihw) [reply kubernetes/utils/deserialize.py]: I can't think of an API that has `create_{0}_with_http_info` but doesn't have `read_{0}_with_http_info`/`read_namespaced_{0}_with_http_info`. Could you give an example?

File: kubernetes/utils/deserialize.py (hunk 2 above)
Comment (chekolyn) [reply kubernetes/utils/deserialize.py]: I think in this situation I was thinking more of the json file output from kubectl  "all.objs.json" more than the way kubectl interacts with the API; in this context I was trying to support loading such files (with multiple object kinds) via the load_from_json() function.

Please see this https://github.com/kubernetes-client/python/issues/977#issuecomment-543009440 for some additional info.

Comment (roycaihw) [reply kubernetes/utils/deserialize.py]: It would be nice if we can return ObjectsList when the input is indeed a single ObjectsList. ObjectsList's are standard API classes, and a user should be able to give a "ObjectsList" as the klass hint. 

For the edge case with `kubectl get all`, I see it as a lower priority because:
- `kubectl get all` is a broken command. The term `all` is misleading and what people think it does is usually wrong: https://github.com/kubernetes/kubectl/issues/151. Currently it is suggested to not use `kubectl get all` and use `kubectl api-resources` instead.
- I haven't seen lots of requests for deserializing a list of objects with different kinds. And I think it can be broken down by iterating through the list.

I'm okay with adding a utility for deserializing `all.objs.json`, as long as we support the standard ObjectList kinds e.g. `V1DeploymentList`

Comment (chekolyn) [reply kubernetes/utils/deserialize.py]: Sound good @roycaihw I agree with the priorities you have outlined. Thank you for your insight.

File: kubernetes/utils/deserialize.py [hunk 6]
```diff
@@ -133,21 +139,48 @@ def response_type_from_dict(data, verbose=False):
...
+            fnc_lookup = None
     else:
-        fnc_lookup = None
+        # Expect the user to load namespaced objects more often
+        if hasattr(k8s_api, "read_namespaced_{0}_with_http_info".format(kind)):
+            fnc_lookup = "read_namespaced_{0}_with_http_info".format(kind)
+        elif hasattr(k8s_api, "read_{0}_with_http_info".format(kind)):
+            fnc_lookup = "read_{0}_with_http_info".format(kind)
+        # Try with the get if no read fnc found
+        # elif hasattr(k8s_api, "get_{0}_with_http_info".format(kind)):
```
chekolyn: Missed this, need to correct it the commented out code.

roycaihw:
Please let me know when it's ready for another round of review

kubernetes/utils/deserialize.py
```diff
@@ -171,8 +171,8 @@ def response_type_from_dict(data, verbose=False):
         elif hasattr(k8s_api, "read_{0}_with_http_info".format(kind)):
             fnc_lookup = "read_{0}_with_http_info".format(kind)
         # Try with the get if no read fnc found
-        # elif hasattr(k8s_api, "get_{0}_with_http_info".format(kind)):
-        #     fnc_lookup = "get_{0}_with_http_info".format(kind)
+        elif hasattr(k8s_api, "get_{0}_with_http_info".format(kind)):
+            fnc_lookup = "get_{0}_with_http_info".format(kind)
```
chekolyn: I remember encountering this before but can't replicate it now. Can't seem to find the object I encountered this with (I'm coding new system after switching jobs). I'm still unsure if we should just remove it.

roycaihw: `get_{0}_with_http_info` covers the following APIs
```
get_api_group
get_api_resources
get_api_versions
get_code
get_cluster_custom_object*
```

They all have models well-defined, except custom objects which are dynamically defined by user. This line looks good to me and I think we should keep it

chekolyn: Sounds good; thanks for your notes.

mgoerens:
Short message to say that I'm interested in seeing this feature merged, thankful for the work so far, and kind of sad to see that the PR is currently closed due to inactivity. Hope you'll be able to pick it up again :sun_with_face:

chekolyn:
Sorry guys the new normal has really limitted my time. I will try to pick it up again as this work it's almost done. 
/reopen

kubernetes/utils/deserialize.py
chekolyn: I have added the klass parameter as an optional one; if one is not given it will try to find the proper klass with the help of the response_type_from_dict function.

kubernetes/utils/deserialize.py
chekolyn: The implementation has been changed to return k8s list objects as requested.

chekolyn:
@roycaihw can we do another round of reviews? thank you.

bpereto:
thanks for you work, it will be useful in my projects :) 

I have some information about performance regarding the deserialization which I want to share. I don't think its relevant for merging this PR, but eventually interesting as a side note for users with a lot of objects.

A JSON file with 314 Objects (which is IMHO small) needs on my machine ~0.56 seconds to load with a for-loop with `utils.load_from_dict` per object (after parsing json by myself).   
The same amount of objects with a "kind: List" structure takes ~2.29 seconds to load with `utils.load_from_dict`.

As not a member of this project: the feature request works for me.

chekolyn:
Hello @bpereto,

Thank you for your interest in this topic; I have not done much profiling with large lists and your findings are most welcomed. 
I was wondering if you can add some of the code you were using to profile the deserialize; in your comment you mentioned the same function twice `utils.load_from_dict` I was wondering if you meant `utils.load_from_json` or `load_from_yaml ` on the second one.

chekolyn:
Hello @bpereto,

I was doing some tests on my end, but I've only seen some marginal variances.

These are some of my results:
```
(k8s_python) sa@leela:~/bin$ time python k8s_deserialize.py 
Just read file
'read_file'  2007.61 ms
Read file and count items
'read_file'  2040.86 ms
'get_items_count'  2177.36 ms
Items count: 5775
Read file and use utils.load_from_dict
'read_file'  1991.78 ms
Lookup info: {'group': 'Core', 'version': 'v1', 'kind': 'list', 'items_kind': 'pod', 'fcn_to_call': 'CoreV1Api', 'fnc_lookup': 'list_namespaced_pod_with_http_info'}
Lookup function found: list_namespaced_pod_with_http_info in k8s_api: <class 'kubernetes.client.apis.core_v1_api.CoreV1Api'> response_type: V1PodList lookup info: {'group': 'Core', 'version': 'v1', 'kind': 'list', 'items_kind': 'pod', 'fcn_to_call': 'CoreV1Api', 'fnc_lookup': 'list_namespaced_pod_with_http_info'}
'get_objs'  17893.79 ms
Use k8s utils.load_from_json
Lookup info: {'group': 'Core', 'version': 'v1', 'kind': 'list', 'items_kind': 'pod', 'fcn_to_call': 'CoreV1Api', 'fnc_lookup': 'list_namespaced_pod_with_http_info'}
Lookup function found: list_namespaced_pod_with_http_info in k8s_api: <class 'kubernetes.client.apis.core_v1_api.CoreV1Api'> response_type: V1PodList lookup info: {'group': 'Core', 'version': 'v1', 'kind': 'list', 'items_kind': 'pod', 'fcn_to_call': 'CoreV1Api', 'fnc_lookup': 'list_namespaced_pod_with_http_info'}
'get_objs_from_utils'  18325.33 ms

real	0m42.901s
user	0m41.048s
sys	0m1.853s
```


I'm using this script to test it and using a json file with about 5K items. 

```python
#!/bin/bin/env python3
#
#
import json
import kubernetes.utils as utils
import time

file_path = "/home/sa/Documents/all.pods.json"


def timeit(method):
    def timed(*args, **kw):
        ts = time.time()
        result = method(*args, **kw)
        te = time.time()
        if 'log_time' in kw:
            name = kw.get('log_name', method.__name__.upper())
            kw['log_time'][name] = int((te - ts) * 1000)
        else:
            print('%r  %2.2f ms'% \
                  (method.__name__, (te - ts) * 1000)
                  )
        return result
    return timed

@timeit
def read_file(file):
    with open(file) as myfile:
        data = myfile.read()
        data = json.loads(data)

        return data

@timeit
def get_items_count(file):
    count=0

    data = read_file(file_path)
    for i in data.get("items"):
        count +=1

    return count

@timeit
def get_objs(file):
    data = read_file(file)

    obj = utils.load_from_dict(data, verbose=True)

    return obj

@timeit
def get_objs_from_utils(file):
    return utils.load_from_json(file_path, verbose=True)


print("Just read file")
read_file(file_path)

print("Read file and count items")
print("Items count: {}".format(get_items_count(file_path)))

print("Read file and use utils.load_from_dict")
obj = get_objs(file_path)

print("Use k8s utils.load_from_json")
obj = get_objs_from_utils(file_path)
```

chekolyn:
This PR is still awaiting to be reviewed, thank you.

VoVAllen:
Still interested in this feature. Thanks for your contribution

vnredmon:
I'll echo the interest in this feature. It would be great to be able to move from a yaml file to k8s object smoothly. Thanks for all of your work @chekolyn!

jovan-absci:
Any plans to merge this? This is a really useful feature, thanks.

smoke:
This is pretty decent implementation, I hope PR will be re-opened and merged as this feature is quite needed.

For instance I want to define and provide a yaml template in file, defining all the Ops specific nits and bits like labels, namespace, service accounts, etc.
and from the python code to load the template, change few parts of it - e.g. some argument / command to be run and create / apply the final declaration.

dboreham:
Also interested in this functionality, to avoid needless custom parsing of yaml snippets into Python classes in one of our projects.

Issue #974, opened 2019-10-01
[RFC] Major breaking changes in the upcoming release


roycaihw:
Hi Kubernetes Python client developers, this is a request for comment issue. Please leave comments if you have any opinion/suggestion. Thanks!

**What happened**:
A majority of the code base is generated code. This repo is switching from using swagger-codegen to using openapi-generator. For more backgrounds:
- https://github.com/kubernetes-client/gen/issues/93
- https://github.com/kubernetes-client/gen/pull/97

Major breaking changes were detected when we released 11.0.0a1 with openapi-generator:
- [Changelog](https://github.com/kubernetes-client/python/blob/release-11.0/CHANGELOG.md#v1100a1)
- [11.0.0a1 release page](https://github.com/kubernetes-client/python/releases/tag/v11.0.0a1)
- [11.0.0a1 release PR](https://github.com/kubernetes-client/python/pull/931)

The breaking changes are:
1. [Major] `kubernetes.client.apis` package is renamed to `kubernetes.client.api`
2. [Major] `kubernetes` package code now uses absolute import instead of relative import
3. [Minor] The `swagger_types` attribute in all models is renamed to `openapi_types`

the major breaking changes come from https://github.com/swagger-api/swagger-codegen/pull/6839#discussion_r148603571, which is also in latest swagger-codegen. This repo didn't have these changes because we were depending on a old version of swagger-codegen.

**What versions are affected?**:
- All `8.0.x`, `9.0.x`, `10.0.x` versions are not affected
- `11.0.0a1` (corresponding to Kubernetes 1.15) is the only version that's affected currently
- The next release `11.0.0b1` and future releases (corresponding to Kubernetes 1.15, 1.16+) will be affected.

**Impact and fix**:

- All clients that import `kubernetes.client.apis` must change the import to `kubernetes.client.api` when upgrading to 11.0.0b1+, to avoid import error.
- Clients that use relative import and patch the `kubernetes` package locally may experience unexpected import error. Clients should install `kubernetes` package properly following the instructions in the [Readme](https://github.com/kubernetes-client/python#installation) or the [release page](https://github.com/kubernetes-client/python/releases).

roycaihw:
cc @dims @fabianvf @mbohlool @micw523 @oz123 @scottilee @tomplus @yliaog

oz123:
@roycaihw my 2 cents (appologies for being so scrooge, it's late here):
[...]

PR #995, opened 2019-10-29
Generate 11.0.0-snapshot client

roycaihw:
Updated python-base submodule, collected CHANGELOG from python-base and 11.0.0a1, and generated 11.0.0-snapshot client

Added an alias package with deprecation message for the major breaking change as https://github.com/kubernetes-client/python/issues/974#issuecomment-537256658 suggested
/cc @oz123 

TODO:
- [x] wait for test to be green (our test code in master branch still imports the deprecated package)
- [x] collect CHANGELOG for API changes since 11.0.0a1
- [ ] install locally and test the deprecation warning behavior


roycaihw:
diff between 11.0.0-snapshot and 11.0.0a1:
```
$ git diff 11.0.0-snapshot..upstream/release-11.0 --stat | grep -v "|   2 +-" | grep -v "examples\/" | grep -v "e2e_test\/"
 .travis.yml                                        |   4 -
 CHANGELOG.md                                       |  18 ++-
 README.md                                          |  36 ++++--
 RELEASE.md                                         |   8 +-
 kubernetes/README.md                               |   4 +-
 kubernetes/client/__init__.py                      |   4 +-
 kubernetes/client/api/custom_objects_api.py        |  34 ++----
 kubernetes/client/api_client.py                    |   4 +-
 kubernetes/client/apis/__init__.py                 |  11 --
 kubernetes/client/configuration.py                 |   6 +-
 kubernetes/docs/CustomObjectsApi.md                |  28 ++---
 kubernetes/utils/__init__.py                       |   1 -
 kubernetes/utils/quantity.py                       |  75 ------------
 scripts/constants.py                               |   4 +-
 scripts/rest_client_patch.diff                     |   4 +-
 scripts/swagger.json                               |  36 +-----
 scripts/update-client.sh                           |   7 ++
 setup.py                                           |   7 +-
 test-requirements.txt                              |   1 -
 tox.ini                                            |  29 ++++-
```

roycaihw:
@oz123 Ah I didn't enable the flag so I didn't see the warning message when I tested. Thanks for verifying it! I'm curious if the log level is configurable to make the warning visible all the time.

roycaihw:
Okay with https://github.com/kubernetes-client/python/pull/995/commits/950df115f7d44981fc0823e08299cd7f8a458efc the deprecation warning is printed by default when the old module gets imported. @oz123 WDYT? 

The warning filters are defined [here](https://docs.python.org/3/library/warnings.html#the-warnings-filter).

oz123:
@roycaihw I think it's a good idea to enable the warning per default.

roycaihw:
@oz123 Thanks for the review!

/assign @yliaog

drolando:
If I understand the docs correctly I think https://github.com/kubernetes-client/python/commit/950df115f7d44981fc0823e08299cd7f8a458efc is turning warnings on for everything. That means that if my code uses both the kubernetes client and other libraries, you're turning on the warnings also for those entirely unrelated libraries.
I didn't try this, but I think you should specify also `module` and make that apply only to the `kubernetes` module

roycaihw:
@drolando You're right. It turns on warning for unrelated paths. E.g.
```
import kubernetes.client.apis
import warnings
warnings.warn(
    "another warning.",
    DeprecationWarning
)
```

```
.../kubernetes/client/apis/__init__.py:13: DeprecationWarning: The package kubernetes.client.apis is renamed and deprecated, use kubernetes.client.api instead (please note that the trailing s was removed).
  DeprecationWarning
example.py:5: DeprecationWarning: another warning.
  DeprecationWarning
```
> I didn't try this, but I think you should specify also module and make that apply only to the kubernetes module

could you explain a bit more?

drolando:
I was looking at https://docs.python.org/3/library/warnings.html#the-warnings-filter and it says you can pass in a `module` argument to limit the scope.

So you'd probably want something like `warnings.filterwarnings('default', module='kubernetes')`

roycaihw:
@drolando Thanks for the suggestion. I verified that the example worked with https://github.com/kubernetes-client/python/pull/995/commits/b3164930dd1789dd66915acd6772f92f512cec47.

roycaihw:
/hold cancel
/cc @yliaog 
This pull is ready for review.

roycaihw:
@yliaog gentle ping

CHANGELOG.md
```diff
@@ -1,3 +1,43 @@
+# v11.0.0b1
```
yliaog: so you're releaseing a1 and b1 together?

roycaihw: no. This is copying the a1 release note from the release branch, since we didn't updated the release note in the master branch when we did a1 release.

PR #1007, opened 2019-11-16
Use ==/!= to compare str, bytes, and int literals

cclauss:
Identity is not the same thing as equality in Python so use ==/!= to compare str, bytes, and int literals. In Python >= 3.8, these instances will raise SyntaxWarnings so it is best to fix them now. https://docs.python.org/3.8/whatsnew/3.8.html#porting-to-python-3-8


roycaihw:
this looks correct. Could you add a testcase for this?

PR #1033, opened 2019-12-18
Updated string comparison as per python3.8 syntax

rishikant42:
- python3.8 raise SyntaxWarning if someone use `is` keyword for string comparison.

```
$ python
Python 3.8.0 (default, Oct 28 2019, 16:14:01) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "str1" is "str1"
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True
>>> "str2" is not "str3"
<stdin>:1: SyntaxWarning: "is not" with a literal. Did you mean "!="?
True
```
- Added test cases for `create_from_dict` util function.


micw523:
This is a duplicate of #1007.

We don't really need many test cases from ```create_from_dict``` due to their using the exact same backend code.

/close

PR #1049, opened 2020-01-08
Fix Custom Objects API example

jpbalarini:
I was trying to create a namespaced custom object using the provided example without success. I found that the resource that the example provides is missing the `spec` wrapper. 

I was able to create a custom object successfully after changing that. Some reference documentation regarding the custom resource definition:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#customresourcedefinition-v1beta1-apiextensions-k8s-io

Thanks!


micw523:
/lgtm
ref:
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/

PR #1050, opened 2020-01-09
Release 11.0.0 stable client

palnabarun:
Updates version constants and re-generates the client


palnabarun:
@roycaihw I had a doubt, should I have merged master to the release-11.0 branch so as to get the latest fixes from master?

micw523:
I think he usually uses fast forward merge?

micw523:
Also, if there are base repo changes that haven't been brought in yet, it could be a good idea to do that as well.

palnabarun:
> Also, if there are base repo changes that haven't been brought in yet, it could be a good idea to do that as well.

I checked out the base repo changes when I generated the client. I will check once more just to be extra sure.

palnabarun:
> I think he usually uses fast forward merge?

I also saw that for the earlier releases. Wasn't sure whether to do it.

roycaihw:
> should I have merged master to the release-11.0 branch so as to get the latest fixes from master?

If we detected bugs in the pre-releases (alpha & beta) and fixed them in master, you can update the pre-release branch by doing https://github.com/kubernetes-client/python/blob/master/devel/release.md#update-pre-release-branch.

palnabarun:
@roycaihw Thanks for pointing that out.

palnabarun:
One more doubt I had, if there are changes in the repo, where should the changelog's be committed? This PR or the PR which targets master?

roycaihw:
> palnabarun wants to merge 6 commits into kubernetes-client:master from palnabarun:release-11.0

It looks like this PR still targets the master branch. Perhaps you need to open a new PR to change that

> if there are changes in the repo, where should the changelog's be committed

If there are [API changes](https://github.com/kubernetes-client/python/blob/master/devel/release.md#sanity-check-generated-client), the client needs to be re-generated and release notes should be collected in the master branch first. Then the new client and the release notes will be [pulled](https://github.com/kubernetes-client/python/blob/master/devel/release.md#update-pre-release-branch) into the release branch. So they are mostly in sync.

palnabarun:
Closing this PR and filing a new one.

PR #1067, opened 2020-02-02
V1Taint value was listed as optional

rhristov:
The description itself says required, while the notes was confusingly saying [optional]. We remove the [optional] note


micw523:
/close
Please refer to here:
https://github.com/kubernetes-client/python/blob/master/CONTRIBUTING.md

I suggest this being filed as an issue instead.

PR #1069, opened 2020-02-04
Rest: Warn about swallowed _request_timeout

sbrandtb:
If `_request_timeout` is neither an int, nor a 2-tuple, it is swallowed without further notice which is a rather unfortunate because the level developers would have to look for this issue is pretty deep.

This actually leads to confusion already, see https://github.com/apache/airflow/pull/6643#issuecomment-575438090

While it would break backwards compatibility to raise an exception, we should at least warn the developer.

This is meant as kind of a discussion base. I am not fully aware of how you developers would like to deal with this. Feedback appreciated :slightly_smiling_face:

### DIFF
--- a/kubernetes/client/rest.py
+++ b/kubernetes/client/rest.py
@@ -17,6 +17,7 @@ import json
 import logging
 import re
 import ssl
+import warnings
 
 import certifi
 # python 2 and python 3 compatibility library
@@ -145,6 +146,11 @@ class RESTClientObject(object):
                   len(_request_timeout) == 2):
                 timeout = urllib3.Timeout(
                     connect=_request_timeout[0], read=_request_timeout[1])
+            else:
+                warnings.warn(
+                    "_request_timeout ignored because it is neither an "
+                    "integer, nor a 2-tuple. This will become an exception "
+                    "in the future.", DeprecationWarning)
 
         if 'Content-Type' not in headers:
             headers['Content-Type'] = 'application/json'


sbrandtb:
I am working on the CLA...

micw523:
This issue is probably better filed upstream, since this file is automatically generated.

roycaihw:
I agree with @micw523. This file is generated by https://github.com/openapitools/openapi-generator. The discussion and change fit better there.

/close

cjgibson:
For those that stumble across this issue in the near future, this root issue here is resolved within https://github.com/openapitools/openapi-generator and the fix will be released alongside their upcoming `5.0.0` release. In the interim, you can fix this (somewhat easily) by opting to pass a `(connection, read)` timeout tuple instead of a bare timeout integer. E.g.:

```python
>>> from kubernetes.client.configuration import Configuration
>>> from kubernetes.client.rest import RESTClientObject
>>> rest_client_object = RESTClientObject(Configuration())
>>> impossible_timeout = 1e-10  # seconds, light travels ~3 cm
>>>
>>> # will ignore the provided timeout
>>> rest_client_object.request('GET', 'http://neverssl.com', _request_timeout=impossible_timeout)
<kubernetes.client.rest.RESTResponse object at 0x7f9c03cfff90>
>>>
>>> # will respect the provided timeout
>>> rest_client_object.request('GET', 'http://neverssl.com', _request_timeout=(
...     0.1 * impossible_timeout, 0.9 * impossible_timeout))
socket.timeout: timed out

During handling of the above exception, another exception occurred:
urllib3.exceptions.ConnectTimeoutError: (<urllib3.connection.HTTPConnection object at 0x7f9c03e20410>, 'Connection to neverssl.com timed out. (connect timeout=1.0000000000000001e-11)')

During handling of the above exception, another exception occurred:
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='neverssl.com', port=80): Max retries exceeded with url: / (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x7f9c03e20410>, 'Connection to neverssl.com timed out. (connect timeout=1.0000000000000001e-11)'))
```

I actually ended up isolating the problem on my end before finding this issue here, so search keywords to aid the all mighty Googs:
Kubernetes doesn't timeout, ignores timeout, never times out, hangs on network, waits for network

PR #1073, opened 2020-02-06
Cleanup ThreadPool with atexit rather than __del__

### LINKED ISSUE #1037: GC hanging on API client cleanup
Python hangs during garbage collection of the API client, in the `__del__` method.

Garbage collection succeeds and the process exits

```python
import kubernetes

client = kubernetes.config.new_client_from_config()
thread = kubernetes.client.ApisApi(client).get_api_versions(async_req=True)
del client
```

The hang doesn't actually occur until python exits. IE, with the following script:

```python
import kubernetes

client = kubernetes.config.new_client_from_config()
thread = kubernetes.client.ApisApi(client).get_api_versions(async_req=True)
del client
print("done")
```
The output is as follows (I ctrl-ced twice after the process hung):
```
done



^CException ignored in: <bound method ApiClient.__del__ of <kubernetes.client.api_client.ApiClient object at 0x7f8f9ba67c50>>
  File "/home/fabian/Envs/test/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 81, in __del__
    self._pool.join()
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 546, in join
    self._worker_handler.join()
  File "/usr/lib64/python3.6/threading.py", line 1056, in join
    self._wait_for_tstate_lock()
  File "/usr/lib64/python3.6/threading.py", line 1072, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
^CError in atexit._run_exitfuncs:
  File "/usr/lib64/python3.6/multiprocessing/util.py", line 262, in _run_finalizers
    finalizer()
  File "/usr/lib64/python3.6/multiprocessing/util.py", line 186, in __call__
    res = self._callback(*self._args, **self._kwargs)
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 582, in _terminate_pool
    worker_handler.join()
  File "/usr/lib64/python3.6/threading.py", line 1056, in join
    self._wait_for_tstate_lock()
  File "/usr/lib64/python3.6/threading.py", line 1072, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
```

**Environment**:
- Kubernetes version (`kubectl version`): Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.6+2e5ed54", GitCommit:"2e5ed54", GitTreeState:"clean", BuildDate:"2019-10-10T22:04:13Z", GoVersion:"go1.12.8", Compiler:"gc", Platform:"linux/amd64"}

- OS (e.g., MacOS 10.13.6): Fedora 27 (Workstation Edition)
- Python version (`python --version`): 2.7.15 and 3.6.6
- Python client version (`pip list | grep kubernetes`): 10.0.1

fabianvf:
This removes the `__del__` function from the generated Python client,
and replaces it with a `cleanup` function. When a `ThreadPool` is created,
the `cleanup` function is registered with the `atexit` module.

This PR also allows the client to be used as a context manager, which
will automatically clean up after itself rather than having to wait til
process exit.

This fixes #1037, where the API client could hang indefinitely at
garbage collection.

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -10,6 +10,7 @@
 
 from __future__ import absolute_import
 
+import atexit
 import datetime
 import json
 import mimetypes
@@ -77,11 +78,19 @@ class ApiClient(object):
         # Set default User-Agent.
         self.user_agent = 'OpenAPI-Generator/11.0.0-snapshot/python'
 
-    def __del__(self):
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
+    def close(self):
         if self._pool:
             self._pool.close()
             self._pool.join()
             self._pool = None
+            if hasattr(atexit, 'unregister'):
+                atexit.unregister(self.close)
 
     @property
     def pool(self):
@@ -89,6 +98,7 @@ class ApiClient(object):
          avoids instantiating unused threadpool for blocking clients.
         """
         if self._pool is None:
+            atexit.register(self.close)
             self._pool = ThreadPool(self.pool_threads)
         return self._pool
 
--- a/kubernetes/test/test_api_client.py
+++ b/kubernetes/test/test_api_client.py
@@ -1,0 +1,25 @@
+# coding: utf-8
+
+
+import atexit
+import weakref
+import unittest
+
+import kubernetes
+
+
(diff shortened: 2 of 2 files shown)


yliaog:
could you add a test that can run successfully with this fix, and would fail without the fix?

PR #1079, opened 2020-02-11
cherry-pick of #1073 to release-10.0: Fix python3 hang

fabianvf:
Cleanup ThreadPool with atexit rather than __del__

(cherry picked from commit 0976d59d6ff206f2f428cabc7a6b7b1144843b2a)

Original PR: #1073


fabianvf:
@roycaihw I'm not totally sure why CI is failing, it doesn't look related to these changes

roycaihw:
@fabianvf I think those tests failed because the new minikube cluster (newer k8s version) we test against doesn't support those APIs any more. You could cherrypick https://github.com/kubernetes-client/python/pull/888 and https://github.com/kubernetes-client/python/pull/966 to fix those tests

roycaihw:
```
PyYAML requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*' but the running Python is 3.4.8
```

It looks like PyYAML has dropped 3.4 support. I think we can do the same for 10.0.2 https://github.com/kubernetes-client/python/pull/807

fabianvf:
@roycaihw ah, would you like me to revert that last commit then?

roycaihw:
@fabianvf I think yours is better.

/lgtm
/approve

Issue #1039, opened 2019-12-24
Can't update the image of CronJob with patch_namespaced_cron_job


ricardozd:
**What happened (please include outputs or screenshots)**:

I'm creating a deploy in AWS with lambda for set the images in Kubernetes Cluster.

I want set image in CronJob:

Cron:

```
NAME                                SCHEDULE      SUSPEND   ACTIVE   LAST SCHEDULE   AGE
cronjob.batch/analytics-staging-0   */5 * * * *   False     0        4m23s           144m
```

Describe cronjob:

```
Name:                          analytics-staging-0
Namespace:                     staging-app
Labels:                        app=analytics-staging-0
Annotations:                   <none>
Schedule:                      */5 * * * *
Concurrency Policy:            Replace
Suspend:                       False
Successful Job History Limit:  824641908604
Failed Job History Limit:      1
Starting Deadline Seconds:     <unset>
Selector:                      <unset>
Parallelism:                   1
Completions:                   1
Pod Template:
  Labels:  <none>
  Containers:
   analytics-staging-0:
    Image:      accountid.dkr.ecr.eu-west-1.amazonaws.com/app:latest
    Port:       <none>
    Host Port:  <none>
    Command:
      /opt/app/analytics
    Environment:
      AWS_REGION:    eu-west-1
    Mounts:          <none>
  Volumes:           <none>
Last Schedule Time:  Tue, 24 Dec 2019 15:20:00 +0100
Active Jobs:         analytics-staging-0-1577197200

```
Python Code:

```

def set_image_job(image, namespace, name, conn):

     api_instance = conn.BatchV1beta1Api()
 
     body = {
         "spec": {
             "template": {
                 "spec": {
                     "containers": [
                         {
                             "image": image
                         }
                     ]
                 }
             }
         }
     }
 
     try:
         response = api_instance.patch_namespaced_cron_job(name, namespace, body, pretty=True)
         print(response)
     except Exception as e:
         print("Exception when calling BatchV2alpha1Api->patch_namespaced_cron_job: %s\n" % e)
 
 if __name__ == "__main__":
     conn = get_connection_eks()
     eks_resource = EKSResources('staging-app', conn)
     jobs = eks_resource.get_cron_jobs()
     for item in jobs:
         set_image_job('accountid.dkr.ecr.eu-west-1.amazonaws.com/app:latest', 'staging-app',
                       item, conn) 

```
**Note: item = analytics-staging-0**

The response is ok, but don't change the image.

[...]

PR #1083, opened 2020-02-18
Add example code for enabling debugging

roycaihw:
ref https://github.com/kubernetes-client/python/issues/1039#issuecomment-582570422

setting `debug` to true allows the client to log HTTP request sent to the server.


micw523:
/lgtm
On the other hand, we should probably think about a master document on explaining what the examples do. People won’t necessarily expect this example to teach them how to do debug logging by looking at the file name.

roycaihw:
I agree. We have examples that cover similar areas. It would be better to de-dup the examples and document their focuses in a README

Issue #1081, opened 2020-02-18
__deserialize_date return type dependent on runtime environment


juliantaylor:
the __deserialize_date function behaves differently depending on which dependencies you have installed at runtime.
https://github.com/kubernetes-client/python/blob/master/kubernetes/client/api_client.py#L585

if dateutils is installed you get datetime objects for dates and if not you get strings.

This is very bad behaviour for a library as code will break depending on which other packages are currently installed.
The function should be changed to always return a single type regardless of current runtime environment.
One way to do that would be to make the dependency mandatory and throw an import error on library load or to remove the dependency by parsing the string with stdlib methods.

roycaihw:
ref https://github.com/kubernetes-client/python/pull/1084#discussion_r381526844. The issue should be fixed when we upgrade the code-generator version in the 12.0.0a1 release

PR #1084, opened 2020-02-19
make dateutil mandatory for api client

juliantaylor:
Avoids differing incompatible return types depending on what modules are
installed at runtime.
This is backward incompatible though likely the less breaking variation
as most environments should have dateutils installed due to the
requirements.txt.

Closes gh-1081

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -18,6 +18,7 @@ from multiprocessing.pool import ThreadPool
 import os
 import re
 import tempfile
+import dateutil.parser
 
 # python 2 and python 3 compatibility library
 import six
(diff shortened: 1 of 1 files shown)


kubernetes/client/api_client.py
```diff
@@ -18,6 +18,7 @@
 import os
 import re
 import tempfile
+import dateutil.parser
```
roycaihw: This file is auto-generated by openapi-generator. Therefore this change won't persist when we re-generate the client. 

This change has been made upstream https://github.com/OpenAPITools/openapi-generator/pull/4316 v4.2.1. The change will be picked up when we upgrade to use the newer version (currently we are at v3.3.4)

roycaihw:
/close

#1088 will include the upstream generator upgrade (https://github.com/kubernetes-client/python/pull/1084#discussion_r381526844)

PR #1087, opened 2020-02-22
add basic watch configmap e2e test

roycaihw:
in future we can add tests for
- watching custom objects
- using watch for pod logs

### DIFF
--- a/kubernetes/e2e_test/test_watch.py
+++ b/kubernetes/e2e_test/test_watch.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


kubernetes/e2e_test/test_watch.py
```diff
@@ -0,0 +1,94 @@
...
+        resp = api.create_namespaced_config_map(
+            body=configmap_b, namespace='default')
+
+        # patch configmap b
+        configmap_b['data']['config'] = "{}"
+        resp = api.patch_namespaced_config_map(
+            name=name_b, namespace='default', body=configmap_b)
+
+        # delete all configmaps
+        resp = api.delete_collection_namespaced_config_map(
```
micw523: Looks like the resp on line 56, 66, 71, 75 are unused. Is it possible to not assign the return value to a variable?

roycaihw: thanks. Updated

PR #1091, opened 2020-02-26
Release 10.1.0

fabianvf:
Pulls in #1073 to prevent hangs during garbage collection


kubernetes/client/apis/custom_objects_api.py
```diff
@@ -1657,7 +1657,7 @@ def patch_cluster_custom_object_with_http_info(self, group, version, plural, nam
 
         # HTTP header `Content-Type`
         header_params['Content-Type'] = self.api_client.\
-            select_header_content_type(['application/merge-patch+json', 'application/json-patch+json'])
+            select_header_content_type(['application/json-patch+json', 'application/merge-patch+json'])
```
roycaihw: This is a known issue that breaks existing clients. Could you restore the custom object patch behavior following https://github.com/kubernetes-client/python/blob/master/devel/release.md#hot-issues?

fabianvf: Every time I think I've finally learned how to read I prove myself wrong

roycaihw:
/lgtm
/approve
Thanks!

Issue #1062, opened 2020-01-27
ModuleNotFoundError: No module named 'kubernetes.dynamic'


jerukitumns:
**What happened (please include outputs or screenshots)**:

Cannot execute script from example documentation for list pod.

```
Traceback (most recent call last):
  File "kube.py", line 1, in <module>
    from kubernetes import client, config
  File "/home/myuga/.local/lib/python3.8/site-packages/kubernetes/__init__.py", line 21, in <module>
    import kubernetes.dynamic
ModuleNotFoundError: No module named 'kubernetes.dynamic'
```

**What you expected to happen**:
module cannot load properly

**How to reproduce it (as minimally and precisely as possible)**:
-

**Environment**:
- Kubernetes version (`kubectl version`): v1.17.0
- OS : Manjaro Linux
- Python version : 3.8.1
- Python client version (`pip list | grep kubernetes`)

WeiJieQiang:
I have the same error here as well.

WeiJieQiang:
The error was solved by uninstall kubernetes from pip3 and reinstall using `pip3 install kubernetes`. The version I am using is kubernetes 10.0.1.

roycaihw:
/assign @micw523 

ref https://github.com/kubernetes-client/python/pull/1035

micw523:
Because this adds a new symbolic link (and we did not support importing from base.dynamic directly), I think reinstalling the package is needed.

roycaihw:
I think it's a bug fix that hasn't been released yet. In [11.0.0b2](https://github.com/kubernetes-client/python/blob/v11.0.0b2/kubernetes/__init__.py) we still don't have https://github.com/kubernetes-client/python/pull/1035 yet. 

On a side note, @micw523 do you know why we didn't add kubernetes.dynamic to setup.py? https://github.com/kubernetes-client/python/blob/073cf4d89e532f92b57e8955b4efc3d5d5eb80cf/setup.py#L60-L63

micw523:
Ah... I think I simply forgot. That might be the reason why.

roycaihw:
The fixes will be included in the upcoming 11.0.0 release https://github.com/kubernetes-client/python/pull/1097

roycaihw:
Closing the issue as 11.0.0 has been released. Please reopen if you hit any issue

/close

PR #1096, opened 2020-03-02
Add kubernetes.dynamic to setup.py pkg list

micw523:
This address the concerns in #1062.


roycaihw:
/lgtm
/approve

We should make sure this is included in the 11.0.0 release. cc @palnabarun

palnabarun:
@roycaihw Noted. Thanks. :)

PR #1100, opened 2020-03-09
Redo pypi token encryption for travis ci deployment

roycaihw:
The pypi deployment was [failing](https://travis-ci.org/kubernetes-client/python/builds/660257148) with 
```
Uploading distributions to https://upload.pypi.org/legacy/
Uploading kubernetes-11.0.0_snapshot-py2-none-any.whl
100%|##########| 1.46M/1.46M [00:00<00:00, 3.70MB/s]
NOTE: Try --verbose to see response content.
HTTPError: 403 Client Error: Invalid or non-existent authentication information. See https://pypi.org/help/#invalid-auth for details for url: https://upload.pypi.org/legacy/
...
PyPI upload failed.
failed to deploy
```
even though it reported `Authenticated as kubernetes` in an earlier step. I couldn't find a way to enable `--verbose` through travis.

I issued a new pypi api token and re-encrypted it as a travis token. I also added some minor changes. Let's see if they help this time. 

/cc @fabianvf @palnabarun

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-# ref: https://docs.travis-ci.com/user/languages/python
 language: python
 dist: xenial
 services:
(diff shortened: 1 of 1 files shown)


micw523:
It does not seem like the deploy stage is run.
I have two questions:
- Will the deploy stage run on any pull request, saying, is it possible that unverified code will run its way to PyPI release without going through the PR process, and got released to PyPI?
- Since the token is visible, is it possible for someone to copy the token and do an HTTP request to upload a dummy/malicious release to PyPI to hijack the package?

roycaihw:
> Will the deploy stage run on any pull request

The deployment is triggered on [tag](https://docs.travis-ci.com/user/deployment/pypi#deploying-tags). I think tagging requires write access to this repo and cannot be achieved with a PR.

> Since the token is visible, is it possible for someone to copy the token

The pypi API token is encrypted through `travis encrypt your-api-token --add deploy.password` and can only be used by travis ci. See https://docs.travis-ci.com/user/deployment/pypi for more details.

micw523:
/lgtm
This sounds good to me!

roycaihw:
Tested again with tag `v0.0.0-travis-experiment2`. The server complained about the semver this time. Hopefully this means we've made through the authentication :) 

```
HTTPError: 400 Client Error: '11.0.0-snapshot' is an invalid value for Version. Error: Start and end with a letter or numeral containing only ASCII numeric and '.', '_' and '-'. See https://packaging.python.org/specifications/core-metadata for url: https://upload.pypi.org/legacy/
```

I'm going to create an experimental branch to test with a real semver generated (v0.0.0a1)

roycaihw:
I experimented in [this branch](https://github.com/kubernetes-client/python/commits/travis-pypi-experiment). v0.0.0a2 was deployed to pypi successfully in [this job](https://travis-ci.org/kubernetes-client/python/jobs/660495630). The [py34 job](https://travis-ci.org/kubernetes-client/python/jobs/660495633) failed because some installation requires py35+. The [remaining jobs](https://travis-ci.org/kubernetes-client/python/builds/660495627) in the build skipped pypi deploy because v0.0.0a2 existed already

palnabarun:
I have tested the package by running https://github.com/kubernetes-client/python/blob/master/examples/out_of_cluster_config.py against Python 2.7, 3.4, 3.5, 3.6, 3.7 and 3.8.

The build failure on Py3.4 will be handled when I push the multi stage build pipeline.

fabianvf:
For further improvements it may make sense to move the deploy step into a separate stage, so that it will only run if all prior tests pass. We could also add an early validation stage that will fail out if a non-semver tag is set, to catch a malformed tag

PR #1101, opened 2020-03-10
Travis configuration now does extra tag validation

fabianvf:
Separate build matrix into 3 stages - verify-tag, test, and deploy
- verify-tag compares the `kubernetes.__version__` to `$TRAVIS_TAG`. Only
runs on pushed tags
- test runs the actual tests
- deploy deploys the package to pypi. Only runs on pushed tags

PR #1102, opened 2020-03-11
Fix verify_tag build stage in Travis pipeline

palnabarun:
Ref: https://github.com/kubernetes-client/python/pull/1101#issuecomment-597479381


roycaihw:
We've [verified that the config works](https://github.com/kubernetes-client/python/tree/travis-pypi-experiment), and released 10.1.0 and 11.0.0. Thanks for all the contribution!

PR #1103, opened 2020-03-11
Update the hotfix patch list and add a script to automate the patching process

palnabarun:
This PR updates the list of hot issues in the [release management documentation]( https://github.com/kubernetes-client/python/blob/master/devel/release.md#hot-issues)

Also, the PR adds a script that cherry-picks the changes from the relevant commits and commits as a single commit.

**Please note:** This is supposed to be a temporary hack until the upstream code generator is fixed the relevant hot issues.


roycaihw:
/lgtm
/approve

Thanks!

PR #1106, opened 2020-03-12
Improve compatibility-matrix documentation in alignment with client-go

roycaihw:
The existing documentation was inaccurate about version skew. I copied the improved documentation from [client-go](https://github.com/kubernetes/client-go#compatibility-matrix). I also cut down the size of the matrix to only include the supported versions. 

/cc @alanjcastonguay

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,25 +82,24 @@ supported versions of Kubernetes clusters.
 
 #### Compatibility matrix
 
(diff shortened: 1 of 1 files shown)


alanjcastonguay:
Review: Diffs of markdown tables are pretty hard to read. :)

palnabarun:
Thanks for simplifying the table and removing the older versions. :)

/lgtm

PR #1122, opened 2020-03-25
Add Istio client

DrAuYueng:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**
Manage Istio resources in kubernetes as same as Deployment, Service, etc.
Resources list:
1. DestinationRule
2. EnvoyFilter
3. Gateway
4. ServiceEntry
5. Sidecar
6. VirtualService

Compatible with istio 1.4.0/1.4.3.

**Which issue(s) this PR fixes:**
None

**Special notes for your reviewer:**
None

Does this PR introduce a **user-facing** change?:
Yes


yliaog:
all the following resources are CRDs, right? you may use the existing CRD support, like the example here, https://github.com/kubernetes-client/python/blob/master/examples/custom_object.py

DestinationRule
EnvoyFilter
Gateway
ServiceEntry
Sidecar
VirtualService

micw523:
Also, istio has its own website and its own github repos. https://istio.io/about/
It's not a part of a standard kubernetes cluster.

DrAuYueng:
> all the following resources are CRDs, right? you may use the existing CRD support, like the example here, https://github.com/kubernetes-client/python/blob/master/examples/custom_object.py
> 
> DestinationRule
> EnvoyFilter
> Gateway
> ServiceEntry
> Sidecar
> VirtualService

something like this, tks.

yliaog:
/close

please reopen if you have other question

PR #1141, opened 2020-04-09
Snapshot the Kubernetes 1.16 release

palnabarun:
- [x] Update version constants
- [x] Generate client
- [x] Apply hotfixes


palnabarun:
I had a doubt whether the API changes be put in CHANGELOG along with this PR or the alpha release PR.

What do you think?

scripts/constants.py
```diff
@@ -15,16 +15,16 @@
...
 # client version for packaging and releasing.
-CLIENT_VERSION = "11.0.0-snapshot"
+CLIENT_VERSION = "12.0.0-snapshot"
 
 # Name of the release package
 PACKAGE_NAME = "kubernetes"
 
 # Stage of development, mainly used in setup.py's classifiers.
-DEVELOPMENT_STATUS = "4 - Beta"
+DEVELOPMENT_STATUS = "3 - Alpha"
```
roycaihw, line 27: nit: we have been using Beta in the master branch (x.0.0-snapshot) meaning a "dev" status. It doesn't have any real effect since we don't publish the snapshot client. Perhaps Alpha fits it better. I think we can stick with one and don't need to toggle between Alpha and Beta in the future

palnabarun: I was researching what should be the classifier for an upstream development branch, but couldn't find any concrete precedence for our use case.

I thought settling for Alpha would make more sense than Beta.

> I think we can stick with one and don't need to toggle between Alpha and Beta in the future
:+1:

palnabarun:
> Would you mind collecting the API changes in CHANGELOG, or is @scottilee going to collect them in the 12.0.0a1 release PR?

I was thinking that collecting them in the 12.0.0a1 would make more sense as one single commit can have all kinds of changes in the CHANGELOG and cherry-pick would thus become easier.

palnabarun:
@roycaihw I have updated this PR with the API changes collected from upstream.

Added them under a heading `v12.0.0a1-snapshot` since I thought putting them under `v12.0.0a1` would have been misleading as that would mean we have already released `v12.0.0a1` client.

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
+# v12.0.0a1-snapshot
+
+Kubernetes API Version: 1.16.9
+
+**API Change:**
```
roycaihw, line 5: Awesome. Thanks for collecting the API changes!

palnabarun:
> nit: I noticed that kubernetes-client/gen#155 was not included. It could be included in this PR or in a separate one

I cut this PR before that change went in. The next PR incorporating update of OpenAPI generator would have the https://github.com/kubernetes-client/gen/pull/155 as well as https://github.com/kubernetes-client/gen/pull/154

palnabarun:
Thanks for the approval. :slightly_smiling_face:

PR #1143, opened 2020-04-11
Add release 1.17.0

irvifa:


irvifa:
@ palnabarun hi :wave: anything that need to be fixed?

palnabarun:
Hi @irvifa, can you please have a look at the release documentation [here](https://github.com/kubernetes-client/python/blob/master/devel/release.md) ? There are some concrete steps needed to be taken in order to generate the client for a specific Kubernetes version.

Also, we are in the process of releasing Kubernetes 1.16 based client at the moment. After 1.16 based client goes into Stable release, then only we will start the process for Kubernetes 1.17 based client.

irvifa:
Oh okay noted, will close this PR then 🙇

PR #1187, opened 2020-06-22
Fix the issue with Github Actions

palnabarun:
The submodule needs to be synced when checking out the repository.

Ref: https://github.com/actions/checkout


palnabarun:
Thank you @yliaog :slightly_smiling_face:

PR #1211, opened 2020-07-23
Release 12.0.0a1

scottilee:
/assign @palnabarun @roycaihw


kubernetes/README.md
```diff
@@ -1,4 +1,4 @@
-# kubernetes.client
```
roycaihw, line 1: the drop of `kubernetes` weird. I couldn't reproduce it locally. Also I noticed that this commit includes a file `kubernetes/swagger.json` which we don't commit anymore. Could you double check and share the `kubernetes-client/gen` version, `openapi-generator` version and your OS version being used?

kubernetes/.gitlab-ci.yml
```diff
@@ -0,0 +1,33 @@
...
+  image: python:3.5-alpine
+nosetest-3.6:
+  extends: .nosetest
+  image: python:3.6-alpine
+nosetest-3.7:
+  extends: .nosetest
+  image: python:3.7-alpine
+nosetest-3.8:
+  extends: .nosetest
+  image: python:3.8-alpine
```
palnabarun, line 33: nit: You can remove this file too. I am guessing this is coming from the base submodule.

README.md
```diff
@@ -115,11 +115,13 @@ between client-python versions.
...
 | 8.0             | Kubernetes main repo, 1.12 branch    | ✗                             |
 | 9.0 Alpha/Beta  | Kubernetes main repo, 1.13 branch    | ✗                             |
-| 9.0             | Kubernetes main repo, 1.13 branch    | ✓                             |
+| 9.0             | Kubernetes main repo, 1.13 branch    | ✗                             |
 | 10.0 Alpha/Beta | Kubernetes main repo, 1.14 branch    | ✗                             |
 | 10.0            | Kubernetes main repo, 1.14 branch    | ✓                             |
 | 11.0 Alpha/Beta | Kubernetes main repo, 1.15 branch    | ✗                             |
 | 11.0            | Kubernetes main repo, 1.15 branch    | ✓                             |
+| 12.0 Alpha/Beta | Kubernetes main repo, 1.16 branch    | ✗                             |
+| 12.0            | Kubernetes main repo, 1.16 branch    | ✓                             |
```
palnabarun, line 124:
```suggestion
| 12.0 Alpha/Beta | Kubernetes main repo, 1.16 branch    | ✓                             |
```

Since, this release involves an Alpha client, we should add a tick against that row. When the stable client releases, we will add the last line and cross out the alpha/beta support.

kubernetes/README.md
palnabarun: The `kubernetes-client/gen` version and  `openapi-generator` version should be pinned since the update scripts takes it from upstream.

I am suspecting something which may cause that import issue, but not sure why the issue is occurring though. If you see https://github.com/kubernetes-client/gen/blob/master/openapi/python.sh#L63-L65, those lines find and replace all instances of `import client` and `from client` to `import kubernetes.client` and `from kubernetes.client` respectively. That line should run always when the generator script is executed.

pacoxu:
11.0.0 not work with kubernetes 1.16 with pod startupProbe.

12.0.0a1 will fix it.  Waiting for the pypi new version.

kubernetes/README.md
roycaihw: > If you see https://github.com/kubernetes-client/gen/blob/master/openapi/python.sh#L63-L65, those lines find and replace all instances of import client and from client to import kubernetes.client and from kubernetes.client respectively

Nice catch!

heisaman:
> 11.0.0 not work with kubernetes 1.16 with pod startupProbe.
> 
> 12.0.0a1 will fix it. Waiting for the pypi new version.

Tried v12.01, still not working with kubernetes 1.16 with pod startupProbe?

roycaihw:
@heisaman Could you try the latest GA release https://pypi.org/project/kubernetes/18.20.0/ to see if that fixes the issue. If not, could you open an issue in https://github.com/kubernetes-client/python/issues?

heisaman:
> @heisaman Could you try the latest GA release https://pypi.org/project/kubernetes/18.20.0/ to see if that fixes the issue. If not, could you open an issue in https://github.com/kubernetes-client/python/issues?

@roycaihw Thanks for the reply. I'll try the latest GA release when I am free.

PR #1224, opened 2020-08-03
Missing backslash in V1ExecAction.md

bankyadam:
By this, after the pipe, it was treated as a new cell.

### DIFF
--- a/kubernetes/docs/V1ExecAction.md
+++ b/kubernetes/docs/V1ExecAction.md
@@ -4,7 +4,7 @@ ExecAction describes a \"run in container\" action.
 ## Properties
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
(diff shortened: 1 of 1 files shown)


bankyadam:
I signed it

yliaog:
it is auto-generated, and will be overwritten in next release.

kubernetes/docs/V1ExecAction.md
```diff
@@ -4,7 +4,7 @@ ExecAction describes a \"run in container\" action.
 ## Properties
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
-**command** | **list[str]** | Command is the command line to execute inside the container, the working directory for the command  is root (&#39;/&#39;) in the container&#39;s filesystem. The command is simply exec&#39;d, it is not run inside a shell, so traditional shell instructions (&#39;|&#39;, etc) won&#39;t work. To use a shell, you need to explicitly call out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy. | [optional]
```
roycaihw, line 7: it's generated from the upstream [openapi spec](https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/swagger.json)

```
6613           "description": "Command is the command line to execute inside the container, the working directory for the command  is root ('/') in the container's filesys       tem. The command is simply exec'd, it is not run inside a shell, so traditional shell instructions ('|', etc) won't work. To use a shell, you need to explicitly call        out to that shell. Exit status of 0 is treated as live/healthy and non-zero is unhealthy.",
```

you can send a PR to k/k to escape it

Issue #166, opened 2017-03-31
Add missing supports for kubernetes client stable and gold


mbohlool:
according to new [kubernetes client guidelines](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/csi-new-client-library-procedure.md#client-capabilities) python client is beta support and has silver capabilities. We are almost there to have gold capabilities and stable support. missing parts are:

Gold:
- [ ] Support attach, port-forward calls
- [ ] Proto encoding
- [ ] Generated examples

Stable:
- [ ] Support level documented per-platform
- [ ] Deprecation policy (backwards compatibility guarantees documented)
- [ ] Release process well documented
- [ ] N documented users of the library

dims:
@mbohlool what does "proto encoding" mean?

mbohlool:
@lavalamp should know more about this, but I assume it means we encode the request body with proto instead of json.

lavalamp:
Correct, request & response bodies. Here's an example proto file: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/apps/v1beta1/generated.proto

You may not be able to determine the proto encoding from the OpenAPI spec--it'll be interesting to see you try to support these protos in a non-go language...

Also the levels were chosen so that Silver was a perfectly acceptable sort of client to have, so it's not bad to be Silver :)

It's just that if someone is going to write a high-bandwidth controller of some sort, proto is a lot more efficient.

brendandburns:
@lavalamp what does "Generated Examples" mean in that doc? Does that mean that you have example code for the library? or that it is generated from something?

pokoli:
Sorry if it's too obious, but does proto mean using this library: https://github.com/google/protobuf/tree/master/python ?

mbohlool:
I looked briefly. The library needs protoc to be installed using other
means than pip. If prefer other python only solutions but if that is only
option, we should take it. Basically we want to be able to talk proto with
api server instead of json. That could include using generated proto
messages in k8s main repo. This needs some investigation before
implementation.

BenElgar:
Just to clarify, does this mean port forwarding is not currently supported?

micw523:
That is correct. Port forwarding is not yet supported on this client.

mayu2010:
port-forward call still not supported in python-client now?

TheOriginalAlex:
Is there a record of documented users of the library?  i.e. A markdown file or something.  I'd like to add to it if so.

[...]

PR #1237, opened 2020-08-24
Unittests for portforwarding ability added in python-base.

iciclespider:
Looking for feed back if this looks on track as a way of implementing port forwarding.

These unittest additions provide usage examples for the port forwarding proposal here: https://github.com/kubernetes-client/python-base/pull/210

Also added a port forwarding example in the examples directory.

https://github.com/kubernetes-client/python/issues/166

### DIFF
--- a/examples/pod_portforward.py
+++ b/examples/pod_portforward.py
@@ -124,6 +124,7 @@ def portforward_commands(api_instance):
     # urllib.request. The same can be done with urllib3.util.connection.create_connection
     # if the "requests" package is used.
     socket_create_connection = socket.create_connection
+
     def kubernetes_create_connection(address, *args, **kwargs):
         dns_name = address[0]
         if isinstance(dns_name, bytes):
@@ -144,7 +145,8 @@ def portforward_commands(api_instance):
                         port = service_port.target_port
                         break
                 else:
-                    raise RuntimeError("Unable to find service port: %s" % port)
+                    raise RuntimeError(
+                        "Unable to find service port: %s" % port)
                 label_selector = []
                 for key, value in service.spec.selector.items():
                     label_selector.append("%s=%s" % (key, value))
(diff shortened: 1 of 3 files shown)


examples/pod_portforward.py
```diff
@@ -0,0 +1,123 @@
...
+    response.close()
+    print('Status:', response.status)
+    print(html)
+
+
+def main():
+    config.load_kube_config()
+    c = Configuration()
+    c.assert_hostname = False
+    #Configuration.set_default(c)
```
yliaog, line 116: instead of commenting it out, the following should work:

c = Configuration.get_default_copy()
c.assert_hostname = False
Configuration.set_default(c)

iciclespider:
@yliaog I submitted another commit for review.

kubernetes/e2e_test/test_client.py
```diff
@@ -182,39 +186,61 @@ def test_portforward_raw(self):
...
+                    break
+                data = sock.recv(1024)
+                if data:
+                    reply += data
+                else:
+                    assert False, 'Unexpected sock close'
+            self.assertEqual(reply, sent)
+            sock.close()
+        self.assertIsNone(pf.error(1234))
+        self.assertIsNone(pf.error(1235))
```
yliaog, line 243: it would be nice to have a case for error is not None. that does not need to be in this PR, it can be added later.

yliaog:
https://travis-ci.org/github/kubernetes-client/python/jobs/725011007 failure, please fix.

iciclespider:
> https://travis-ci.org/github/kubernetes-client/python/jobs/725011007 failure, please fix.

I got everything to pass, except for the unittest module that is attempting to import `portforward`, which does not yet exist in the `kubernetes/base` git submodule.

roycaihw:
@iciclespider Could you update the submodule to fix the import failure? https://github.com/kubernetes-client/python/blob/master/devel/submodules.md#update-submodule

PR #1253, opened 2020-09-09
updating base for file-like config load support

ryphon:
Base PR success, opening PR in main client to support file like config loading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,3 +174,4 @@ This will cause a failure in  non-exec/attach calls. If you reuse your api clien
 recreate it between api calls that use _stream_ and other api calls.
 
 See more at [exec example](examples/pod_exec.py).
+


ryphon:
No explanation for port forward test failure. The changes I made would have failed much higher up in the test than this, additionally they would have failed in MANY other places if something was wrong with the python-base code.

yliaog:
please fix Travis CI build failure

ryphon:
@yliaog Something else is wrong with the portforward_raw test that doesn't involve this PR. The Changelog PR has a failure with this test as well, with the only changes being purely README text.

ryphon:
Yeah... @yliaog I'm not sure which commit it was. But that portforward_raw test appears to be inconsistent.

PR #1287, opened 2020-10-16
Create Python Program to Check Leap Year

sameerraj362000:
To check leap year


palnabarun:
/close

@sameerraj362000 -- We are not participating in Hacktoberfest.

Issue #1284, opened 2020-10-15
host is set to localhost when loading from kube config file in v12


limonkufu:
**What happened (please include outputs or screenshots)**:
We are using the following code to set up:
```
from kubernetes import client, config, watch

config.load_kube_config(kubeconfig)
    host = config.kube_config.Configuration().host
    logging.info("HOST INFO: {}".format(host))
```

The `kubeconfig` file has `server` field set up correctly. This is working with with version 11.0.0 correctly but when we change the version to 12.0.0 then it returns:

`HOST INFO: http://localhost`

I tried to investigate it and according to [kube_config.py file](https://github.com/kubernetes-client/python-base/blob/fd322f70aa6c33782ab466afd4eaae271a27cbcb/config/kube_config.py) it should be set correctly in [`_load_cluster_info` method](https://github.com/kubernetes-client/python-base/blob/fd322f70aa6c33782ab466afd4eaae271a27cbcb/config/kube_config.py#L529) but it is not setting up correctly.

**What you expected to happen**:
Host url to be set correctly in compliance with kube-config file

**How to reproduce it (as minimally and precisely as possible)**:

- Install version 11.0.0 run the above code with a valid kube-config file
- See the host url set correctly
- Update the version to 12.0.0 and run the code again
- See the host url set incorrectly to `http://localhost`

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (`kubectl version`): v1.18.2
- OS (e.g., MacOS 10.13.6): Ubuntu 18.04
- Python version (`python --version`) 3.7
- Python client version (`pip list | grep kubernetes`) 12.0.0

Limonkufu:
@felixhuettner thanks this solves the issue but using `get_default_copy()` instead of direct call seems counter-intuitive after setting the configuration from file. (note that this change is also not documented in the readme/examples)

rasisuku:
I am also facing this issue.

chancez:
At the very least, this needs a mention in the changelog so that users know how to fix this.

ntavares:
load_incluster_config() broke too for 12.0.0. It's not clear to me yet how to retain the same mechanism...

DontWorry33:
My process for setting up my configuration when using load_incluster_config() was:

```
from kubernetes import client, config, utils
from kubernetes.client import Configuration

config.load_incluster_config()
c = Configuration()
c.assert_hostname = False
Configuration.set_default(c)
```

[...]

PR #1285, opened 2020-10-15
Release 12.0.1

palnabarun:
We missed removing the bug fix changelog for the 503 client discovery exception bug. Since this has already been fixed, updated the client with the latest submodule and updated the constants.

- [x] Update submodule
- [x] Update constants
- [x] Generate client
- [x] Update CHANGELOG

**Do Later:** Add information to the release hotfix documentation to cherry-pick the test for configuration behavior after removing the stubbed OpenAPI tests.

PR #1303, opened 2020-10-30
Document a breaking change in the default configuration

roycaihw:
xref: https://github.com/kubernetes-client/python/issues/1284, https://github.com/OpenAPITools/openapi-generator/pull/4485, https://github.com/OpenAPITools/openapi-generator/pull/5315

Document a breaking change of `kubernetes.config.Configuration()` that exists since v12.0.0a1. Add a unit test that asserts the behavior. 

Special notes: 
1. The upstream change fixed a bug in different configuration objects unexpectedly sharing a common dict, but it broke the use case where people use `Configuration()` to easily access one global configuration. 
2. Not sure if the test folder gets wiped when we regenerate the client. If so, we may need to carry a patch, or find another place to host it. 
3. I looked at python-base and found the only place we use `Configuration()` is [_refresh_oidc](https://github.com/kubernetes-client/python-base/blob/2da2b981ca806b25487ad92d01a2164815c18517/config/kube_config.py#L413). The client doesn't rely on the default configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Kubernetes API Version: 1.16.14
 
 Kubernetes API Version: 1.16.14
 
+**Breaking Change:**
+
+- `kubernetes.config.Configuration()` will now return the default "initial" configuration, `kubernetes.config.Configuration.get_default_copy()` will return the default configuration if there is a default set via `Configuration.set_default(c)`, otherwise, it will also return the default "initial" configuration. [OpenAPITools/openapi-generator#4485](https://github.com/OpenAPITools/openapi-generator/pull/4485), [OpenAPITools/openapi-generator#5315](https://github.com/OpenAPITools/openapi-generator/pull/5315)
+
 **API Change:**
 
 - Resolve regression in metadata.managedFields handling in update/patch requests submitted by older API clients ([#91748](https://github.com/kubernetes/kubernetes/pull/91748), [@apelisse](https://github.com/apelisse)) [SIG API Machinery and Testing]
--- a/kubernetes/test/test_configuration.py
+++ b/kubernetes/test/test_configuration.py
@@ -1,0 +1,39 @@
+# coding: utf-8
+
+import unittest
+
+from kubernetes.client import Configuration
+
+class TestConfiguration(unittest.TestCase):
+
+    def setUp(self):
+        pass
+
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ Kubernetes API Version: 1.16.14
 
 Kubernetes API Version: 1.16.14
 
+**Breaking Change:**
+- `kubernetes.config.Configuration()` no longer returns the default configuration. Use `kubernetes.config.Configuration.get_default_copy()` now to get the default configuration. [OpenAPITools/openapi-generator#4485](https://github.com/OpenAPITools/openapi-generator/pull/4485), [OpenAPITools/openapi-generator#5315](https://github.com/OpenAPITools/openapi-generator/pull/5315)
```
yliaog, line 28: it's a bit confusing ...

kubernetes.config.Configuration() still returns the default "initial" configuration,
kubernetes.config.Configuration.get_default_copy() will return the default configuration if there is a default set via Configuration.set_default(c), otherwise, it will also return the default "initial" configuration

roycaihw: Updated. I reworded it a little bit, since the change in `kubernetes.config.Configuration()` was the regression.

palnabarun:
I was wondering if we should do this as a minor release since changeling and releases are usually immutable.

/hold
(just for resolving the above doubt, the change looks good to me 👍🏽)

roycaihw:
@palnabarun Yes, I think that will help raise awareness. Would you mind fold the change to https://github.com/kubernetes-client/python/pull/1285, and add an action required note in 12.0.1?

palnabarun:
Yeah sure. Let me cherry-pick your commit and apply there.

palnabarun:
@roycaihw Can you please split your change to the CHANGELOG and tests into two commits? I want to preserve attribution when I edit those. :)

palnabarun:
The test addition also requires adding a note to the Hot Patch section to the release docs.

We need to specify that the commit which adds the test needs to be cherrypicked while performing the release.

roycaihw:
@palnabarun Thanks. I split the commit into two

palnabarun:
@roycaihw -- I have cherry picked both changes to the Release 12.0.1 PR. #1285

roycaihw:
/close

in favor of #1285

PR #1307, opened 2020-11-07
Release 17.0 snapshot

palnabarun:
- [x] Update constants
- [x] Generate client
- [x] Remove tests
- [x] Cherrypick hotfixes
- [x] Update CHANGELOG
- [x] Update README

**Note:**
The CHANGELOG also mentions changes in the versioning scheme of the client and about dropping Python 2 support in the next stable release (v18.0.0)

The CHANGELOG in master needs to be updated with changes from the v12 stable release (#1308). This branch already contains changes from #1308. Until then, putting a hold on this PR.

/hold


CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
+# v17.0.0-snapshot
+
+Kubernetes API Version: 1.17.13
+
+**Important Information:**
+
+- The Kubernetes Python client versioning scheme has changed. The version numbers used till Kubernetes Python Client v12.y.z lagged behind the actual Kubernetes minor version numbers. From this release, the client is moving a version format `vY.Z.P` where `Y` and `Z` are respectively from the Kubernetes version `v1.Y.Z` and `P` would incremented due to changes on the Python client side itself. Ref: https://github.com/kubernetes-client/python/issues/1244
```
roycaihw, line 7: We should add this to README. Maybe when we do the first v17 alpha release

CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
...
+- Python 2 had reached [End of Life](https://www.python.org/doc/sunset-python-2/) on January 1, 2020. The Kubernetes Python Client will drop support for Python 2 from the next release (v18.0.0) and will no longer provide support to older clients as per the [Kubernetes support policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions).
```
roycaihw, line 8: Do we plan to send notes to the kubernetes-client slack channel and kubernetes-dev mailing list?

roycaihw:
Have a question but it doesn't block this PR. 

/lgtm
/approve
/hold cancel

CHANGELOG.md
palnabarun: Yes. That's what my thoughts were. After we cut the first Alpha, we should put both the things in the Important Imformation block in the README.

CHANGELOG.md
palnabarun: Yes. We should be sending those to the mailing list and Slack channel.

Issue #1300, opened 2020-10-28
Test flake: kubernetes/e2e_test/test_client.py::TestClient::test_portforward_raw


roycaihw:
The presubmit `test_portforward_raw` is flaky. Example: https://travis-ci.org/github/kubernetes-client/python/jobs/735898772

=================================== FAILURES ===================================
_______________________ TestClient.test_portforward_raw ________________________
self = <kubernetes.e2e_test.test_client.TestClient testMethod=test_portforward_raw>
    def test_portforward_raw(self):
        client = api_client.ApiClient(configuration=self.config)
        api = core_v1_api.CoreV1Api(client)
    
        name = 'portforward-raw-' + short_uuid()
        pod_manifest = manifest_with_command(
            name,
            ' '.join((
                '((while true;do nc -l -p 1234 -e /bin/cat; done)&);',
                '((while true;do nc -l -p 1235 -e /bin/cat; done)&);',
                'sleep 60',
            ))
        )
        resp = api.create_namespaced_pod(body=pod_manifest,
                                         namespace='default')
        self.assertEqual(name, resp.metadata.name)
        self.assertTrue(resp.status.phase)
    
        while True:
            resp = api.read_namespaced_pod(name=name,
                                           namespace='default')
            self.assertEqual(name, resp.metadata.name)
            self.assertTrue(resp.status.phase)
            if resp.status.phase != 'Pending':
                break
            time.sleep(1)
    
        pf = portforward(api.connect_get_namespaced_pod_portforward,
                         name, 'default',
                         ports='1234,1235,1236')
        self.assertTrue(pf.connected)
        sock1234 = pf.socket(1234)
        sock1235 = pf.socket(1235)
        sock1234.setblocking(True)
        sock1235.setblocking(True)
        sent1234 = b'Test port 1234 forwarding...'
        sent1235 = b'Test port 1235 forwarding...'
        sock1234.sendall(sent1234)
        sock1235.sendall(sent1235)
        reply1234 = b''
        reply1235 = b''
        while True:
            rlist = []
            if sock1234.fileno() != -1:
                rlist.append(sock1234)
            if sock1235.fileno() != -1:
                rlist.append(sock1235)
            if not rlist:
                break
            r, _w, _x = select.select(rlist, [], [], 1)
            if not r:
                break
            if sock1234 in r:
                data = sock1234.recv(1024)
                self.assertNotEqual(data, b'', "Unexpected socket close")
                reply1234 += data
            if sock1235 in r:
[...]

PR #1329, opened 2020-11-22
Implement back end raw port test container using python.

iciclespider:
Attempts to fix: #1300

### DIFF
--- a/kubernetes/e2e_test/test_client.py
+++ b/kubernetes/e2e_test/test_client.py
@@ -13,6 +13,7 @@
 # under the License.
 
 import json
+import os
 import select
 import socket
 import time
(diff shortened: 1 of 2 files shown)


kubernetes/e2e_test/test_client.py
```diff
@@ -167,17 +168,64 @@ def test_portforward_raw(self):
...
+        with open(os.path.join(os.path.dirname(__file__), 'port_server.py')) as fh:
+            port_server_py = fh.read()
         name = 'portforward-raw-' + short_uuid()
-        pod_manifest = manifest_with_command(
-            name,
-            ' '.join((
-                '((while true;do nc -l -p 1234 -e /bin/cat; done)&);',
-                '((while true;do nc -l -p 1235 -e /bin/cat; done)&);',
-                'sleep 60',
-            ))
```
yliaog, line 177: what about using the k8s test image below, instead of writing your own echo server?
https://github.com/kubernetes/kubernetes/tree/master/test/images/agnhost#netexec

iciclespider: This is a test of raw socket usage, not http based. Granted, the code could send the needed http request using the raw socket. It then has to pull in what is being returned and compare against a hard coded assumption of the return value. If the agnhost image ever changes the return value, the test will fail.

This is also a test of raw socket usage against multiple ports. I do not know if agnhost can orchestrate that. I am not a go guru and would have to spend time understanding exactly what it is this agnhost container is doing.

My preference is to control exactly what is occurring on the test container and not create external dependencies. This is a python client library, and I find it simpler to understand a python based container, rather then a go based implemention.

yliaog: sounds good. thanks for the fix

PR #1347, opened 2021-01-04
[DO NOT MERGE] debug configmap e2e

roycaihw:
The test failed to delete the configmap object (e.g. https://github.com/kubernetes-client/python/pull/1345). Print out the API responses to debug. 

/assign
/hold

PR #1348, opened 2021-01-05
e2e: use labels for configmap api; wait for default service account creation

roycaihw:
The [Bound Service Account Tokens](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md) feature creates a configmap `kube-root-ca.crt` in every namespace. Use a label to select the test objects we create. 

Also wait for the default service account to be created before creating pods that need it. 

/hold
wait for the tests to pass

The submodule update contains two commits, which do not require CHANGELOG change:
1. add labels to `dynamic/test_client.py::TestDynamicClient::test_configmap_apis`
2. add Python 3.9 to build

ref https://github.com/kubernetes-client/python-base/pull/222, https://github.com/kubernetes-client/python/pull/1347

### DIFF
--- a/kubernetes/e2e_test/test_client.py
+++ b/kubernetes/e2e_test/test_client.py
@@ -19,14 +19,21 @@ import socket
 import time
 import unittest
 import uuid
+import six
 
 from kubernetes.client import api_client
 from kubernetes.client.api import core_v1_api
 from kubernetes.e2e_test import base
 from kubernetes.stream import stream, portforward
 from kubernetes.stream.ws_client import ERROR_CHANNEL
+from kubernetes.client.rest import ApiException
 
 import six.moves.urllib.request as urllib_request
+
+if six.PY3:
+    from http import HTTPStatus
+else:
+    import httplib
 
 def short_uuid():
     id = str(uuid.uuid4())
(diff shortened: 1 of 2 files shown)


roycaihw:
There are some new test failure regarding the `default` service account not getting created before pod creation. Thought that it was a flake but it seemed not. I will take another look soon

roycaihw:
(fixed some syntax error) The tests pass now https://travis-ci.org/github/kubernetes-client/python/builds/754195162. 
@yliaog Could you take another look? Thanks

/hold cancel

kubernetes/e2e_test/test_client.py
```diff
@@ -65,6 +72,23 @@ def test_pod_apis(self):
 
         name = 'busybox-test-' + short_uuid()
         pod_manifest = manifest_with_command(name, "while true;do date;sleep 5; done")
+
+        # wait for the default service account to be created
```
yliaog, line 76: better to give a timeout than wait forever

kubernetes/e2e_test/test_client.py
```diff
@@ -130,6 +154,23 @@ def test_exit_code(self):
...
```
yliaog, line 158: ditto

kubernetes/e2e_test/test_client.py
roycaihw: done

/hold
wait for test results

roycaihw: /hold cancel
The test is green

yliaog:
/lgtm

thanks for fixing it.

PR #1356, opened 2021-01-19
Update requirements.txt

keshav-creator:
Added the readme in the Requirements file....


roycaihw:
Sorry for the late reply. It's unclear to me why this change is needed. Please add more context on why should we make this change and reopen if necessary.

/close

PR #1363, opened 2021-01-28
Enable leaderelection

Invictus17:
Linking it to https://github.com/kubernetes-client/python/pull/1218


yliaog:
the module kubernetes/base in this repo needs to be updated to have 'leaderelection'

kubernetes/leaderelection
```diff
@@ -0,0 +1 @@
+base/leaderelection
```
roycaihw, line 1: looks like this is a text file, which caused the import failure. Instead this should be a symbolic link. See https://github.com/kubernetes-client/python/blob/master/kubernetes/dynamic as an example

Invictus17:
There are failures in the 'kubernetes/e2e_test/test_client.py' test which I think are unrelated to my PR.
https://travis-ci.org/github/kubernetes-client/python/builds/757635573

yliaog:
close and reopen to rerun the ci tests

Issue #1396, opened 2021-03-18
TypeError: __init__() got an unexpected keyword argument 'assert_hostname'


VariableDeclared:
**What happened (please include outputs or screenshots)**:

When running the `pod_exec.py` example a TypeError is thrown, this appears to come from one of the depencies that needs an arguement updated.

**What you expected to happen**:

The example would run successfully.

**How to reproduce it (as minimally and precisely as possible)**:

```
mkvirtualenv k8s_test_scripts_test --python python2
cd ..
pip install -r requirements.txt
cd examples/
ls
python pod_exec.py
pip install kubernetes
python pod_exec.py
pip install kubernetes==17.14.0a1
python pod_exec.py
```

**Anything else we need to know?**:

Tried on Python3 and Python2

**Environment**:
- Kubernetes version (`kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-20T02:22:41Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.12-gke.1210", GitCommit:"199a41188dc0ca5d6d95b1cc7e8ba96e05f9dd0a", GitTreeState:"clean", BuildDate:"2021-02-05T18:03:16Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}
```
- OS (e.g., MacOS 10.13.6):
Ubuntu 20.04 LTS
- Python version (`python --version`)
`Python 2.7.18`
- Python client version (`pip list | grep kubernetes`)
kubernetes        17.14.0a1

Akasurde:
@VariableDeclared Could you please check #1401 and let me know if it works for you? Thanks.

roycaihw:
/assign @Akasurde 

Fixed by #1401
/close

VariableDeclared:
> @VariableDeclared Could you please check #1401 and let me know if it works for you? Thanks.

That works, thanks!

PR #1401, opened 2021-03-23
examples/pod_exec.py: fix example

Akasurde:
Use get_default_copy() API with configuration.

### DIFF
--- a/examples/pod_exec.py
+++ b/examples/pod_exec.py
@@ -117,8 +117,11 @@ def exec_commands(api_instance):
 
 def main():
     config.load_kube_config()
-    c = Configuration()
-    c.assert_hostname = False
+    try:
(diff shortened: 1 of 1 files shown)


Akasurde:
Fixes: #1396

examples/pod_exec.py
```diff
@@ -117,8 +117,11 @@ def exec_commands(api_instance):
 
 def main():
     config.load_kube_config()
-    c = Configuration()
-    c.assert_hostname = False
+    try:
+        c = Configuration().get_default_copy()
```
roycaihw, line 121: nit: maybe add a comment that this is the right way of setting the config starting v12.0.0

roycaihw (commented):
Thanks!

/lgtm
/approve

PR #1403, opened 2021-03-26
Allow optional list of YAML objects as param to create_from_yaml util

mecampbellsoup:
## Description

Currently we are doing something like the following to use the `utils.create_from_yaml` helper:

```python
manifest_path = os.path.join('/tmp', name)
with open(template_path, 'r') as t, open(manifest_path, 'w') as outfile:
    template = Template(t.read())
    output = template.render(ns_name=ns_name)
    outfile.write(output)
await utils.create_from_yaml(k8s_client=api, yaml_file=manifest_path)
```
We have a YAML template file which we dynamically inject variables into, which is finally passed to the util function.

There is an extra read happening here that could/should be removed. Adding an optional param `yaml_objects` seemed like a safe, backwards-compatible approach.

### DIFF
--- a/kubernetes/e2e_test/test_utils.py
+++ b/kubernetes/e2e_test/test_utils.py
@@ -13,8 +13,10 @@
 # under the License.
 
 import unittest
+from os import path
 
 import yaml
+
 from kubernetes import utils, client
 from kubernetes.client.rest import ApiException
 from kubernetes.e2e_test import base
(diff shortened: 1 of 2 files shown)


mecampbellsoup:
@yliaog @roycaihw should I rebase this, are you guys interesting in merging it?

roycaihw:
/remove-lifecycle stale

Sorry for the late response. The change looks good to me. Please rebase and I will tag the PR

/approve

mecampbellsoup:
@roycaihw rebased & fixed conflicts!

PR #1405, opened 2021-04-04
API client shows error when metadata is ignored

zidhumenon:
"message": "HorizontalPodAutoscaler.autoscaling \"\" is invalid: metadata.name: Required value: name or generateName is required",


yliaog:
the file is generated, please fix upstream generator.

zidhumenon:
@yliaog Could you please share the link for upstream, where I have to change it?

yliaog:
https://github.com/openapitools/openapi-generator is the generator based on the openapi spec, which is in turn generated by the https://github.com/kubernetes/kubernetes/blob/master/hack/update-openapi-spec.sh

zidhumenon:
@yliaog since its looks like we need to fix it in https://github.com/openapitools/openapi-generator i think we can drop this issue on this repo. what do you say?

yliaog:
sounds good to me

zidhumenon:
@yliaog Thanks for the confirmation.  I am closing this now

Issue #1414, opened 2021-04-12
Document the release process for v18


roycaihw:
ref https://github.com/kubernetes-client/python/pull/1411#issuecomment-818216374

The current release process doesn't support multiple parallel releasing well. I need to document how to update the pre-release branches (release-18.0, release-19.0) to pull changes from master branch without hitting into merge conflicts.

PR #1415, opened 2021-04-13
update release doc

### LINKED ISSUE #1412: Document the workaround for LIBRARY env in client generation
ref https://github.com/kubernetes-client/python/pull/1411#discussion_r611772514

Document the temporary workaround in devel release doc. Cherrypick the workaround into master branch if needed (and add a TODO about removing it once we fix the issue in gen repo).

roycaihw:
fixes #1412, #1414

- add hot issue about required LIBRARY env in update-client.sh;
- add update python-base submodule step;
- clarify when we should collect release notes;
- mention a workaround for avoiding merge conflicts when pulling master;
- mark "making disctribution packages" as legacy and deprecated;
- clarify the write permission is required for creating a Github release page.

cc @scottilee

### DIFF
--- a/devel/release.md
+++ b/devel/release.md
@@ -1,8 +1,8 @@
 # Release process
 
 The release process for the python client involves creating (or updating) a
-release branch, updating release tags, and creating distribution packages and
-uploading them to pypi.
+release branch, updating python-base submodule, updating release tags, and
+creating distribution packages and uploading them to pypi.
 
 There are several releases per version:
 - snapshot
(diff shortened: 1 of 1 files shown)


devel/release.md
```diff
@@ -32,6 +32,21 @@ git pull upstream master
 You may need to fix some conflicts. For auto-generated files, you can commit
 either version. They will be updated to the current version in the next step.
 
+A workaround for merge conflicts is to manually cherrypick the change from the
+master branch to the release branch. See https://github.com/kubernetes-client/python/pull/1411
+for examples.
```
yliaog, line 37: some clarification on the reasoning for this (i.e. parallel multiple releases), also there is a step about committing snapshot to master, that should be optional initially, and when all in-flight releases are completed, the snapshot should be updated to the latest release.

PR #1418, opened 2021-04-18
Make job_crud example print the correct job status after job creation

steveprabha:
create_job function was not printing the current status of the job. Modified the create_job function to call read_namespaced_job_status in order to get the current job status

### DIFF
--- a/examples/job_crud.py
+++ b/examples/job_crud.py
@@ -20,6 +20,8 @@ from os import path
 
 import yaml
 
+from time import sleep
+
 from kubernetes import client, config
 
 JOB_NAME = "pi"
@@ -53,7 +55,16 @@ def create_job(api_instance, job):
     api_response = api_instance.create_namespaced_job(
         body=job,
         namespace="default")
-    print("Job created. status='%s'" % str(api_response.status))
+    # Need to wait for a second for the job status to update
+    sleep(1)
+    print("Job created. status='%s'" % str(get_job_status(api_instance)))
+
+
+def get_job_status(api_instance):
+    api_response = api_instance.read_namespaced_job_status(
+        name=JOB_NAME,
+        namespace="default")
+    return api_response.status
 
 
 def update_job(api_instance, job):
--- a/examples/job_crud.py
+++ b/examples/job_crud.py
@@ -55,7 +55,8 @@ def create_job(api_instance, job):
     api_response = api_instance.create_namespaced_job(
         body=job,
         namespace="default")
-    print("Job created. status='%s'" % str(get_job_status(api_instance)))
+    print("Job created. status='%s'" % str(api_response.status))
+    print("Job status='%s'" % str(get_job_status(api_instance)))
 
 
 def get_job_status(api_instance):
@@ -66,6 +67,7 @@ def get_job_status(api_instance):
             namespace="default")
         if api_response.status.succeeded is not None or api_response.status.failed is not None:
             job_completed = True
+        sleep(1)
     return api_response.status
 
 
--- a/examples/job_crud.py
+++ b/examples/job_crud.py
@@ -17,10 +17,9 @@ Creates, updates, and deletes a job object.
 """
 
 from os import path
-
-import yaml
-
 from time import sleep
+ 
+import yaml
 
 from kubernetes import client, config
 
--- a/examples/job_crud.py
+++ b/examples/job_crud.py
@@ -18,7 +18,7 @@ Creates, updates, and deletes a job object.
 
 from os import path
 from time import sleep
- 
+
 import yaml
 
 from kubernetes import client, config
--- a/examples/job_crud.py
+++ b/examples/job_crud.py
@@ -55,15 +55,17 @@ def create_job(api_instance, job):
     api_response = api_instance.create_namespaced_job(
         body=job,
         namespace="default")
-    # Need to wait for a second for the job status to update
-    sleep(1)
     print("Job created. status='%s'" % str(get_job_status(api_instance)))
 
 
 def get_job_status(api_instance):
-    api_response = api_instance.read_namespaced_job_status(
-        name=JOB_NAME,
-        namespace="default")
+    job_completed = False
+    while not job_completed:
+        api_response = api_instance.read_namespaced_job_status(
+            name=JOB_NAME,
+            namespace="default")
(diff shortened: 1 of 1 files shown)


steveprabha:
@yliaog Please check this and let me know your comments. Thanks

examples/job_crud.py
```diff
@@ -53,7 +55,16 @@ def create_job(api_instance, job):
     api_response = api_instance.create_namespaced_job(
         body=job,
         namespace="default")
-    print("Job created. status='%s'" % str(api_response.status))
+    # Need to wait for a second for the job status to update
+    sleep(1)
```
yliaog, line 59: api_response.status is the status immediately after creating 'job'. then controller will run to update the job status as the job is making progress. the time to take for job to run to completion is not determined, it could be short, could be long. the way to get status update is to Watch it.

steveprabha: @yliaog Do i need to write a function to watch the status and print? If yes can you please elaborate it?

yliaog: https://github.com/kubernetes-client/python/blob/master/examples/pod_namespace_watch.py has an example for 'watch', or, you may periodically poll for status update.

steveprabha: @yliaog Should we just print the status as soon as we get the JOB_NAME available in the watch code? like below
```v1 = client.CoreV1Api()
    w = watch.Watch()
    for event in w.stream(v1.list_pod_for_all_namespaces, timeout_seconds=10):
        if JOB_NAME in event['object'].metadata.name:
              api_response = api_instance.read_namespaced_job_status(
                name=JOB_NAME,
                namespace="default")
     return api_response.status```
     
Or should we wait for the Job to be in the RUNNING phase and then print the status?

Comment (roycaihw) [reply examples/job_crud.py:59]: It would be great if you can print the watch events until the job succeeded/failed. This can be put into a separate function.

Comment (steveprabha) [reply examples/job_crud.py]: @roycaihw @yliaog I have updated the code to Poll for status update until success/failure. Please check.

File: examples/job_crud.py [hunk 2]
```diff
@@ -53,7 +55,18 @@ def create_job(api_instance, job):
...
+
+
+def get_job_status(api_instance):
+    job_completed = False
+    while not job_completed:
+        api_response = api_instance.read_namespaced_job_status(
+            name=JOB_NAME,
+            namespace="default")
+        if api_response.status.succeeded is not None or api_response.status.failed is not None:
+            job_completed = True
```
yliaog, line 68: this is hot loop, please add sleep(...)

examples/job_crud.py
```diff
@@ -53,7 +55,18 @@ def create_job(api_instance, job):
     api_response = api_instance.create_namespaced_job(
         body=job,
         namespace="default")
-    print("Job created. status='%s'" % str(api_response.status))
+    print("Job created. status='%s'" % str(get_job_status(api_instance)))
```
yliaog, line 58: better to keep the previous print line 56 to show the Job is created, then poll for job running status until job is completed

steveprabha: @yliaog sleep for 1 sec is good?

yliaog: that depends on the how long it takes to complete the job, 1s is probably ok.

steveprabha: @yliaog Please check the code changes now. Thanks.

examples/job_crud.py
```diff
@@ -54,6 +56,19 @@ def create_job(api_instance, job):
         body=job,
         namespace="default")
     print("Job created. status='%s'" % str(api_response.status))
+    print("Job status='%s'" % str(get_job_status(api_instance)))
```
yliaog, line 59: better to move the print to the polling loop below, so that every 1s, the status is printed out

steveprabha: @yliaog Do you have any idea why the travis-ci build is failing?

yliaog:
/home/travis/build/kubernetes-client/python/kubernetes/../examples/job_crud.py:67:80: E501 line too long (95 > 79 characters)
diff --git a/examples/job_crud.py b/examples/job_crud.py
index 23282bd..eb66789 100644
--- a/examples/job_crud.py
+++ b/examples/job_crud.py
@@ -17,11 +17,10 @@ Creates, updates, and deletes a job object.
 """
 
 from os import path
+from time import sleep
 
 import yaml
 
-from time import sleep
-
 
 JOB_NAME = "pi"
@@ -59,7 +58,6 @@ def create_job(api_instance, job):
     get_job_status(api_instance)
 
 
-
 def get_job_status(api_instance):
     job_completed = False
     while not job_completed:
ERROR: InvocationError for command /home/travis/build/kubernetes-client/python/scripts/update-pycodestyle.sh (exited with code 1)

yliaog:
@steveprabha 
the output is from here: https://travis-ci.org/github/kubernetes-client/python/jobs/769884870

examples/job_crud.py
```diff
@@ -20,6 +20,8 @@
 
 import yaml
 
+from time import sleep
```
roycaihw, line 23: the [pycodestyle test](https://travis-ci.org/github/kubernetes-client/python/jobs/770682844) is picky about the import. Could you make it 
```
from os import path
from time import sleep
 
import yaml

```
?

steveprabha: @roycaihw I have made the above mentioned change. Please verify now. Thanks

steveprabha: @roycaihw Looks like the build is failing again. Can you please tell me what's happening?

yliaog: are there any special character on the line after " from time import sleep"?

==============================================
diff --git a/examples/job_crud.py b/examples/job_crud.py
index 2174ef7..ab02761 100644
--- a/examples/job_crud.py
+++ b/examples/job_crud.py
@@ -18,7 +18,7 @@ Creates, updates, and deletes a job object.
 
 from os import path
 from time import sleep
- 
+
 import yaml
 
yliaog: https://travis-ci.org/github/kubernetes-client/python/jobs/770874568 has the diff you can check out

steveprabha: No there is no special characters except newline @yliaog

steveprabha: @yliaog looks like there was an space along with the newline and i fixed it now. Please check and approve once the build is successful

PR #1439, opened 2021-04-28
Created a namespaced_custom_object.py and added patch example

Yashks1994:
#### What type of PR is this?
/kind documentation
/kind cleanup

#### What this PR does / why we need it:
Added the example to patch to namespace custom resource and renamed the file to namespaces_custom_object.py because we are providing example to the namespaces custom objects.
#### Which issue(s) this PR fixes:

Fixes #1438

```release-note

```

```docs

```

PR #1440, opened 2021-04-29
Removed custom_object.py and created a new file namespaced_custom_object.py which is more specific to usage of namespaced custom resource 

### LINKED ISSUE #1438: Improve the custom_object.py to be more specific to usage of namespaced custom resource
https://github.com/kubernetes-client/python/tree/master/examples

The [examples/custom_object.py](https://github.com/kubernetes-client/python/blob/master/examples/custom_object.py) example script can be improved by making it more specific to usage of  [kubernetes.client.CustomObjectsApi](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md) for a namespaced scope custom resource.

Following are the suggestion for improvement:
- rename the file to `namespaced_custom_object.py`
- include example usage of `patch_namespaced_custom_object` method

Yashks1994:
#### What type of PR is this?
/kind cleanup
/kind documentation

Removed the custom_object.py and created a new file as namespaced_custom_object.py which is more specific to usage of namespaced custom resource.

Fixes #1438

### DIFF
--- a/examples/namespaced_custom_object.py
+++ b/examples/namespaced_custom_object.py
@@ -17,7 +17,7 @@ Uses a Custom Resource Definition (CRD) to create a custom object, in this case
 a CronTab. This example use an example CRD from this tutorial:
 https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/
 
-The following yaml manifest has to be applied first:
+The following yaml manifest has to be applied first for namespaced scoped CRD:
 
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -72,6 +72,11 @@ def main():
         }
     }
 
+    # patch to update the `spec.cronSpec` field
+    patch_body = {
+        "spec": {"cronSpec": "* * * * */10", "image": "my-awesome-cron-image"}
+    }
+
     # create the resource
     api.create_namespaced_custom_object(
         group="stable.example.com",
@@ -92,6 +97,18 @@ def main():
     )
     print("Resource details:")
     pprint(resource)
+
+    # patch the namespaced custom object to update the `spec.cronSpec` field
+    patch_resource = api.patch_namespaced_custom_object(
+        group="stable.example.com",
+        version="v1",
+        name="my-new-cron-object",
+        namespace="default",
+        plural="crontabs",
+        body=patch_body,
+    )
+    print("Resource details:")
+    pprint(patch_resource)
 
     # delete it
     api.delete_namespaced_custom_object(


examples/namespaced_custom_object.py
```diff
@@ -16,9 +16,7 @@
 Uses a Custom Resource Definition (CRD) to create a custom object, in this case
 a CronTab. This example use an example CRD from this tutorial:
 https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/
-
 The following yaml manifest has to be applied first:
-
```
roycaihw, line 21: let's keep these empty lines. Otherwise LGTM

Yashks1994: Done

Priyankasaggu11929:
@Yashks1994, just a note.

Git takes renaming of the  file with name `custom_object.py` to `namespaced_custom_object.py` as two steps:
    - deleting the file `custom_object.py`
    - and creating a new file `namespaced_custom_object.py` having the same contents as above.

To reflect the deletion of file `custom_object.py`, you need to add the deleted file as well in your git staging & then commit both the deleted file & new file.

Doing that, in the `Files changed` section of this PR, both the deleted file (custom_object.py) & new file (namespaced_object.py) will reflect.

Yashks1994:
@roycaihw needs your /lgtm as CI build passed.

roycaihw:
> The file renaming would look something like this in the PR's Files Changed section.

@Priyankasaggu11929 I was surprised when I saw that the first time. I didn't expect Github to render it that way but it's really beautiful. 

@Yashks1994 LGTM. Please squash the commits as I suggested in https://github.com/kubernetes-client/python/pull/1439#issuecomment-828926719: 
- one commit for adding patch example to custom_object.py
- one commit for renaming the file to namespaced_custom_object.py

examples/namespaced_custom_object.py
```diff
@@ -99,30 +98,28 @@ def main():
...
     )
-    print("Resource deleted")
-
+    print("Resource details:")
+    pprint(patch_resource)
 
-    # patch the namespaced custom object to update the `spec.cronSpec` field
-    patch_resource = api.patch_namespaced_custom_object(
+    # delete it
+    api.delete_namespaced_custom_object(
```
roycaihw, line 114: why do we have a commit that reverse the order of `patch_namespaced_custom_object` and `delete_namespaced_custom_object`?

roycaihw: https://github.com/kubernetes-client/python/pull/1444/commits is what I was asking for, if that makes sense

Yashks1994: I kept the order like this,
we have create, get, patch, delete

roycaihw: currently the two PR have the same code change, but the commits are different

Commits should represent meaningful milestones or units of work. Currently the commits in this PR are: 
- your first commit renames the file and adds a patch method after the delete method, which means the example has a bug and will run into an error
- your second commit reverses the order of the patch method and the delete method to fix the bug in the first commit

please take a look at https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md#squash-commits. The second commit is a "Work in progress" (it fixes a bug in the first commit during development), so please squash it. 

again, I think having two commits like the following will be most reasonable and readable:
- one commit for adding patch example to custom_object.py
- one commit for renaming the file to namespaced_custom_object.py

at this point I'm also okay with having just one commit, which will be more reasonable than the current two commits in this PR

Yashks1994: I am keeping the one commit and with proper commit message.

roycaihw:
Thanks! 
/lgtm

PR #1441, opened 2021-04-29
Detect OS X sed and fail client generation script

roycaihw:
#### What type of PR is this?

/kind cleanup

In the client generation script, fail early if OS X `sed` is being used and suggest using GNU `sed` instead. Both the [python repo script](https://github.com/kubernetes-client/python/blob/0ea3542ae5301aec370406641827c25c2fa1dced/scripts/update-client.sh#L62-L65) and the [gen repo script](https://github.com/kubernetes-client/gen/blob/65a1e4739cf0dcc2b93f887d33fe0ad3772b6169/openapi/python.sh#L56-L65) use `sed` in format that OS X `sed` may not support: https://stackoverflow.com/a/4247319.

Address issues like:
```
--- Patching generated code...
sed: 1: "/Users/scott/Projects/k ...": bad flag in substitute command: 'l'
```

```release-note
NONE
```

/cc @scottilee @yliaog

### DIFF
--- a/scripts/update-client.sh
+++ b/scripts/update-client.sh
@@ -24,6 +24,22 @@ set -o pipefail
 # The openapi-generator version used by this client
 export OPENAPI_GENERATOR_COMMIT="v4.3.0"
 
+# OS X sed doesn't support "--version". This way we can tell if OS X sed is
+# used.
+if ! sed --version &>/dev/null; then
+  # OS X sed and GNU sed aren't compatible with backup flag "-i". Namely
+  # sed -i ... - does not work on OS X
+  # sed -i'' ... - does not work on certain OS X versions
+  # sed -i '' ... - does not work on GNU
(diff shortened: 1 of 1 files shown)


roycaihw:
For more context, here are my local test results:

OS X sed doesn't work with `-i` and `-i''`
```
$ sw_vers
ProductName:	macOS
ProductVersion:	11.2.3
BuildVersion:	20D91
$ /usr/bin/sed -i 's/a/b/g' test
sed: 1: "test": undefined label 'est'
$ /usr/bin/sed -i'' 's/a/b/g' test
sed: 1: "test": undefined label 'est'
```
GNU sed doesn't work with `-i ''`
```
$ sed --version
sed (GNU sed) 4.7
$ sed -i '' 's/a/b/g' test
sed: can't read s/a/b/g: No such file or directory
```

Issue #1378, opened 2021-02-15
How do we do rolling restart?


nsubrahm-aws:
How do we do `kubectl rollout restart deployment your_deployment_name` ? I couldn't locate appropriate documentation for this.

MoShitrit:
@nsubrahm-awsI believe the best way would be to use the [`patch deployment`](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/AppsV1Api.md#patch_namespaced_deployment) function and add a dummy annotation to the `spec.template.metadata.annotations` block. This should trigger the creation of a new replicaset and automatically handle the restart based on the `strategy` specified in the deployment spec.
That is also what `kubectl rollout restart deployment` does behind the scenes. 
When you run `kubectl rollout restart deployment`, it adds a `kubectl.kubernetes.io/restartedAt` annotation which contains a timestamp, like this one:
```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    metadata:
      annotations:
        kubectl.kubernetes.io/restartedAt: "2021-02-15T11:12:54-05:00"
```

roycaihw:
cc @roycaihw 
/assign @MoShitrit

roycaihw:
What @MoShitrit suggested sounds good, if adding the annotation is what kubectl does behind the scene. @nsubrahm-aws Could you try it out and see if it works?

MoShitrit:
For the record, I found it [here](https://github.com/kubernetes/kubectl/blob/c1df07341a15fb36264a018e130ee397332bdfc0/pkg/polymorphichelpers/objectrestarter.go#L51) in the `kubectl` codebase

mark8s:
nice

PR #1450, opened 2021-05-02
add example to demonstrate a rolling restart of the deployment

Priyankasaggu11929:
#### What type of PR is this?
/kind documentation

This PR add `examples/dynamic-client/deployment_rollout-restart.py` python script, to demonstrate how to perform a rolling restart on a deployment resource.

The output of the script looks like:

deployment_rolling_restart

Fixes https://github.com/kubernetes-client/python/issues/1378

```release-note
NONE
```

### DIFF
--- a/examples/dynamic-client/deployment_rolling_restart.py
+++ b/examples/dynamic-client/deployment_rolling_restart.py
@@ -1,0 +1,120 @@
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


examples/dynamic-client/deployment_rolling_restart.py
```diff
@@ -0,0 +1,120 @@
...
+    import datetime
+    import pytz
+
+    # Creating a dynamic client
+    client = dynamic.DynamicClient(
+        api_client.ApiClient(configuration=config.load_kube_config())
+    )
+
+    # fetching the deployment api
+    api = client.resources.get(api_version="apps/v1", kind="Deployment")
```
yliaog, line 35: any reason for using dynamic client, instead of using the typed client?

Priyankasaggu11929: HI @yliaog, thanks for the reveiw.

No, there is no specific reason for using `dynamic client`. I've just replicated the same example using the `typed client`. If required, I can replace this example with the one using `typed client`.

Thank you!

yliaog: ok. i think typed client would benefit from such an example too.

Priyankasaggu11929: Thanks @yliaog. I'll add the same example using the `typed client` too.

PR #1463, opened 2021-05-11
Fix loss of multi-value headers in request

### LINKED ISSUE #1462: Unable to pass multiple header values with client (ex: Impersonate-Group)

When calling the Kubernetes API with the python client it should be possible to pass multiple values for `Impersonate-Group`:

```
        (data, status, headers) = api_client.call_api(
            path,
            'GET',
            auth_settings = ['BearerToken'],
            collection_formats = {'Impersonate-Group': 'multi'},
            header_params = {
                'Impersonate-Group': ['foo', 'bar'],
            },
            response_type = 'object',
        )
```

The result is only the value `bar` is passed.

I would expect that two `Impersonate-Group` headers would be sent, the first with value `foo` and the second with `bar`.

See example code above.

The core issue is https://github.com/kubernetes-client/python/blob/master/kubernetes/client/api_client.py#L136-L137

```
            header_params = dict(self.parameters_to_tuples(header_params,
                                                           collection_formats))
```

The `parameters_to_tuples` call seems reasonable and produces the flattened list of key/value pairs, but then using `dict(...)` causes the second value to be lost. This can be simply fixed by using the urllib3 `HTTPHeaderDict` constructor instead of `dict`.

**Environment**:

Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0-4-g38212b5", GitCommit:"d038424d6d4f1cc39ad586ac0d36dac3a7a37ceb", GitTreeState:"clean
", BuildDate:"2020-06-21T17:49:43Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0+e49167a", GitCommit:"e49167aad6a08046be6ab21ff13029110c76951d", GitTreeState:"clean", 
BuildDate:"2021-01-28T07:35:27Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}

- OS (e.g., MacOS 10.13.6):

Fedora 33

Python 3.9.2

11.0.0

jkupferer:
#### What type of PR is this?

/kind bug

The api client library currently loses request header values when multiple values are passed.

Fixes #1462

This adds the capability to pass multiple values in request headers, such as passing multiple values for `Impersonate-Group`

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -19,6 +19,7 @@ from multiprocessing.pool import ThreadPool
 import os
 import re
 import tempfile
+from urllib3.connection import HTTPHeaderDict
 
 # python 2 and python 3 compatibility library
 import six
@@ -133,7 +134,7 @@ class ApiClient(object):
             header_params['Cookie'] = self.cookie
         if header_params:
             header_params = self.sanitize_for_serialization(header_params)
-            header_params = dict(self.parameters_to_tuples(header_params,
+            header_params = HTTPHeaderDict(self.parameters_to_tuples(header_params,
                                                            collection_formats))
 
         # path parameters


roycaihw:
Thanks for the PR! The change looks reasonable to me. 

`kubernetes/client/api_client.py` is generated code, so a change like this will get lost the next time we re-generate the code. The best way to change the file is to fix the upstream code generator. Would you mind sending a PR to [openapi-generator](https://github.com/OpenAPITools/openapi-generator) to fix this [code template](https://github.com/OpenAPITools/openapi-generator/blob/c966b5fe1f558ff71a7dab98004d33ff3dc4138e/modules/openapi-generator/src/main/resources/python/api_client.mustache#L144)? 

Assuming the openapi-generator maintainers LGTM'ed the change, we can maintain a patch ([like this](https://github.com/kubernetes-client/python/blob/master/scripts/rest_client_patch.diff)) in this repo to speed things up, so we don't have to wait for the next openapi-generator release.

yliaog:
Once OpenAPITools/openapi-generator#9457 is fixed, we should patch the 1.17 / 1.18 release

Issue #1402, opened 2021-03-23
Document client-side timeouts in watch


WouldYouKindly:
Hi!

From the watch example it is not clear that `timeout_seconds` is a server-side timeout, and that the client will hang indefinitely in case of a network failure. I found out about this from [this comment](https://github.com/kubernetes-client/python/issues/1148#issuecomment-626184613) after we experienced the issue in my team.

I suggest documenting this in [the watch example](https://github.com/kubernetes-client/python/blob/master/examples/pod_namespace_watch.py).

roycaihw:
cc @roycaihw

Priyankasaggu11929:
@roycaihw, can I work on updating the documentation around this? Taking reference information from this comment here https://github.com/kubernetes-client/python/issues/1148#issuecomment-626184613.

Thank you!

WouldYouKindly:
Thank you both very much!

Issue #1148, opened 2020-04-16
Watcher stuck forever in a case of network disconnect


NikPaushkin:
**What happened (please include outputs or screenshots)**:
If the network to Kubernetes API disconnects, watcher just gets stuck forever with no exception.

**What you expected to happen**:
Watcher should raise an exception in a case of any network issues.

**How to reproduce it (as minimally and precisely as possible)**:
1. Run watcher for any object list with some timeout (60 seconds, for example)
1. Turn off network adapter on your laptop
1. Wait for 60 seconds (timeout) and watcher still does not return the control: thread just stuck

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (`kubectl version`): 1.17.0
- OS (e.g., MacOS 10.13.6): Windows 10
- Python version (`python --version`): 3.7
- Python client version (`pip list | grep kubernetes`): 11.0.0

roycaihw:
This seems to be related to https://github.com/kubernetes-client/python/issues/868

NikPaushkin:
> This seems to be related to #868

@roycaihw well, it can be fixed like that, if we don't want to fix stream itself

roycaihw:
cc @alanjcastonguay

alanjcastonguay:
Yeah, this is one of the failure modes that prompted #868. But I still don't have a method to fix it, other than simply setting a client-side timeout. Did you have one set in this test?

NikPaushkin:
@alanjcastonguay It works like that always for me. The minimal code is trivial

```py
from kubernetes import client, watch
from datetime import datetime

watcher = watch.Watch()
func = client.CoreV1Api(client.ApiClient(client.Configuration())).list_namespaced_pod
kwargs = dict()
kwargs['timeout_seconds'] = 60
try:
    for event in watcher.stream(func, ** kwargs):
        watcher_last_time_alive = datetime.now()
except Exception as e:
    logging.error("API request failed", extra={
        "API": "Kubernetes",
        "method": func,
        "kwargsGiven": kwargs,
        "error": str(e)
    })

```

Cweiping:
This problem can be repeated in the case of unstable network environment. The more unstable the network, the more likely it is to have problems

alanjcastonguay:
`kwargs['timeout_seconds'] = 60` is a polite request to the server, asking it to cleanly close the connection after 60 seconds. If you have a network outage, this does nothing. You can set this number much higher, maybe to 3600 seconds (1h).

[...]

PR #1467, opened 2021-05-14
add documentation for the server & client side timeout

Priyankasaggu11929:
#### What type of PR is this?
/kind documentation

This PR documents information about `server-side` & `client-side` timeout in the [examples/pod_namespace_watch.py](https://github.com/kubernetes-client/python/blob/master/examples/pod_namespace_watch.py) example script.

Fixes https://github.com/kubernetes-client/python/issues/1402

```release-note
NONE
```
Signed-off-by: Priyanka Saggu <priyankasaggu11929@gmail.com>

### DIFF
--- a/examples/watch/timeout-settings.md
+++ b/examples/watch/timeout-settings.md
@@ -1,0 +1,71 @@
+
+**This documentation briefly provides information about the `server side` & `client side` connection timeout settings, in the watch request handler.**
+
+---
+
+There are two inputs available in the client, that could be used to set connection timeouts:
+
+- `timeout_seconds`
+- `_request_timeout`
+
+---
+
+#### Sever-side timeout (`kwargs['timeout_seconds'] = n`)
+
+- The value of the argument `timeout_seconds`, **n**, (which is time duration in seconds) is consumed at the server side. It is included in the request URL to the server. 
+  
+  *For eg.* ~ `https://localhost:6443/api/v1/namespaces/default/pods?labelSelector=app%3Ddemo&timeoutSeconds=100&watch=True`
+
+- In case, if the `timeout_seconds` value is set, the value `n` would determine the server-side connection timeout duration.
+
+  *For eg.* ~ if `kwargs['timeout_seconds'] = 3600`, then the server-side connection timeout will be equal to 1 hour.
+  
+  This timeout duration is determined by the expression ~ `timeout = time.Duration(3600) * time.seconds`, *i.e.* `timeout = 1 hour`
+
+  ***Refer:*** 
+  - *[https://github.com/kubernetes/apiserver/blob/release-1.20/pkg/endpoints/handlers/get.go#L254](https://github.com/kubernetes/apiserver/blob/release-1.20/pkg/endpoints/handlers/get.go#L254)*
+
+- In case, if the `timeout_seconds` value is not set, then the connection timeout will be a randomized value (in seconds) between `minRequestTimeout` and 2*`minRequestTimeout`, to spread out the load.
+
+  It is determined using the expression ~ `timeout = time.Duration(float64(minRequestTimeout) * (rand.Float64() + 1.0))`
+
+  Where `minRequestTimeout` indicates the minimum number of seconds a handler must keep a request open before timing it out.
+  
+  The default value of `minRequestTimeout` is 1800 seconds.
+
+  ***Refer:***
+  - *[https://github.com/kubernetes/apiserver/blob/release-1.20/pkg/endpoints/handlers/get.go#L257](https://github.com/kubernetes/apiserver/blob/release-1.20/pkg/endpoints/handlers/get.go#L257)*
+  - *[https://github.com/kubernetes/kubernetes/blob/release-1.20/staging/src/k8s.io/apiserver/pkg/server/config.go#L320](https://github.com/kubernetes/kubernetes/blob/release-1.20/staging/src/k8s.io/apiserver/pkg/server/config.go#L320)*
+
+- In case of a network outage, the server side timeout value will have no effect & the client will hang indefinitely without raising any exception. Note, that this is the case provided when there is no other client-side timeout (i.e., `_request_timeout`) value specified.
+
+  (*See the section below for information on `client side timeout`*)
+
+- It is recommended to set this timeout value to a higher number such as 3600 seconds (1 hour).
+
+---
+
+#### Client-side timeout (`kwargs['_request_timeout'] = n`)
+
+- The value of the argument `_request_timeout`, **n** (which is time duration in seconds) is set to the socket used for the connection.
+
+- In case, if the `_request_timeout` value is set, this argument can accept 2 types of input values ~
+    - float, 
+    - a tuple (with a length of 2)
+    
+  ***Refer***
+  - *[https://github.com/kubernetes-client/python/blob/v17.17.0/kubernetes/client/api_client.py#L336-L339](https://github.com/kubernetes-client/python/blob/v17.17.0/kubernetes/client/api_client.py#L336-L339)*
+
+- In case of network outage, leading to dropping all packets with no RST/FIN, the timeout value (in seconds) determined by the `request_timeout` argument, would be the time duration for how long the client will wait before dropping the connection.
+
+- When the timeout happens, an exception will be raised, for eg. ~
+  
+  `urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='localhost', port=6443): Read timed out.`
+  
+- In case, if the `_request_timeout` value is not set, then the default value is **`None`** & socket will have no timeout.
+
+  ***Refer:***
+  - *[https://docs.python.org/3/library/socket.html#socket.getdefaulttimeout](https://docs.python.org/3/library/socket.html#socket.getdefaulttimeout)*
+
+- It is recommended to set this timeout value to a lower number (for eg. ~ maybe 60 seconds).
+
--- a/examples/watch/pod_namespace_watch.py
+++ b/examples/watch/pod_namespace_watch.py
@@ -17,6 +17,12 @@ Uses watch to print the stream of events from list namespaces and list pods.
 The script will wait for 10 events related to namespaces to occur within
 the `timeout_seconds` threshold and then move on to wait for another 10 events
 related to pods to occur within the `timeout_seconds` threshold.
+
+
+Refer to the document below to understand the server-side & client-side
+timeout settings for the watch request handler: ~
+
+https://github.com/github.com/kubernetes-client/python/blob/master/examples/watch/timeout-settings.md
 """
 
 from kubernetes import client, config, watch


examples/pod_namespace_watch.py
```diff
@@ -31,6 +31,25 @@ def main():
     v1 = client.CoreV1Api()
     count = 10
     w = watch.Watch()
+    """
+    [INFO: `server-side` & `client-side` timeout]
+
+    `kwargs['timeout_seconds'] = n`:
+       - It is a server-side timeout, i.e. a polite request to the server,
+         mentioning it to cleanly close the connection after n seconds.
```
yliaog, line 39: https://github.com/kubernetes/apiserver/blob/92392ef22153d75b3645b0ae339f89c12767fb52/pkg/endpoints/handlers/get.go#L255

if it is not set, the default is a random timeout between 0 and minRequestTimeout
https://github.com/kubernetes/apiserver/blob/92392ef22153d75b3645b0ae339f89c12767fb52/pkg/endpoints/handlers/get.go#L258

examples/pod_namespace_watch.py
```diff
@@ -31,6 +31,25 @@ def main():
...
+       - If the user have a network outage, then this parameter will do nothing.
```
yliaog, line 40: watch is for server to push events to client side, so it is not accurate to say 'user have a network usage'

examples/pod_namespace_watch.py
```diff
@@ -31,6 +31,25 @@ def main():
...
+       - The client will hang indefinitely without raising an exception.
+       - You can set this number much higher, maybe to 3600 seconds (1h).
```
yliaog, line 42: the server side default is: faction * 1800

examples/pod_namespace_watch.py
```diff
@@ -31,6 +31,25 @@ def main():
...
+
+    `kwargs['_request_timeout'] = n`:
+       - It is a client-side timeout, configuring the user's local socket.
+       - If the user have a network outage, resulting into dropping all packets
```
yliaog, line 46: ditto, once watch connection is established, server push events to clients, 'user' does not send traffic to server over it

examples/pod_namespace_watch.py
```diff
@@ -31,6 +31,25 @@ def main():
...
+         with no RST/FIN, the value of n will specify, for how long
```
yliaog, line 47: 'how long' is in the number of seconds? better to specify

examples/pod_namespace_watch.py
```diff
@@ -31,6 +31,25 @@ def main():
...
+         the client would wait before realizing the network outage,
+         & dropping the connection.
```
yliaog, line 49: https://github.com/kubernetes-client/python-base/pull/36#discussion_r146445193
has some investigations on this, fyi

Priyankasaggu11929:
Thank you @yliaog for all the pointers. 

I'll read through the links to understand the concepts better, & will update the PR.

Priyankasaggu11929:
@yliaog, I've tried to redo the previous attempt of documentation, adding more information after reading & understanding through the links/pointers you provided above.

Please review. Thank you!

examples/watch/timeout-settings.md
```diff
@@ -0,0 +1,68 @@
...
+- In case, if the `timeout_seconds` value is set, the value `n` would determine the server-side connection timeout duration.
+
+  *For eg.* ~ if `kwargs['timeout_seconds'] = 3600`, then the server-side connection timeout will be equal to 1 hour.
+  
+  This timeout duration is determined by the expression ~ `timeout = time.Duration(3600) * time.seconds`, *i.e.* `timeout = 1 hour`
+
+  ***Refer:*** 
+  - *[kubernetes/apiserver/pkg/endpoints/handlers/get.go#L255](https://github.com/kubernetes/apiserver/blob/92392ef22153d75b3645b0ae339f89c12767fb52/pkg/endpoints/handlers/get.go#L255)*
+
+- In case, if the `timeout_seconds` value is not set, then the connection timeout will be a randomized value (in seconds) between 0 and `minRequestTimeout`, to spread out load.
```
yliaog, line 28: it is not between 0 and minRequestTimeout, it is between minRequestTimeout and 2* minRequestTimeout

examples/watch/timeout-settings.md
```diff
@@ -0,0 +1,68 @@
...
+
+  It is determined using the expression ~ `timeout = time.Duration(float64(minRequestTimeout) * (rand.Float64() + 1.0))`
+
+  Where `minRequestTimeout` indicates the minimum number of seconds a handler must keep a request open before timing it out.
+  
+  The default value of `minRequestTimeout` is 1800 seconds.
+
+  ***Refer:***
+  - *[kubernetes/apiserver/pkg/endpoints/handlers/get.go#L258](https://github.com/kubernetes/apiserver/blob/92392ef22153d75b3645b0ae339f89c12767fb52/pkg/endpoints/handlers/get.go#L258)*
+  - *[kubernetes/release-1.1/docs/admin/kube-apiserver.md](https://github.com/kubernetes/kubernetes/blob/release-1.1/docs/admin/kube-apiserver.md)*
```
yliaog, line 38: the doc is for release 1.1, which is very old, the latest release is 1.20, could you use the latest?

examples/watch/timeout-settings.md
```diff
@@ -0,0 +1,68 @@
...
+
+- In case of a network outage, this timeout value will have no effect & the client will hang indefinitely without raising any exception.
```
yliaog, line 40: what kind of network outage would cause this? it hangs when not accounting for the client side timeout described below, right?

examples/watch/timeout-settings.md
```diff
@@ -0,0 +1,68 @@
...
+- It is recommended to set this timeout value to a higher number such as 3600 seconds (1 hour).
+
+---
+
+#### Client-side timeout (`kwargs['_request_timeout'] = n`)
+
+- The value of the argument `_request_timeout`, **n** (which is time duration in seconds) is set to the socket used for the connection.
+
+- In case, if the `_request_timeout` value is set, this argument can accept 2 types of input values ~
+    - integer (int),
```
yliaog, line 51: according to the documentation, the timeout is float, not an int

examples/watch/timeout-settings.md
```diff
@@ -0,0 +1,68 @@
...
+
+---
+
+#### Client-side timeout (`kwargs['_request_timeout'] = n`)
+
+- The value of the argument `_request_timeout`, **n** (which is time duration in seconds) is set to the socket used for the connection.
+
+- In case, if the `_request_timeout` value is set, this argument can accept 2 types of input values ~
+    - integer (int), 
+    - a tuple (with a length of 2)
```
yliaog, line 52: i don't see the tuple in the documentation, could you provide a pointer?

examples/watch/timeout-settings.md
```diff
@@ -0,0 +1,68 @@
...
+- When the timeout happens, an exception will be raised, for eg. ~
+  
+  `urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='localhost', port=6443): Read timed out.`
+  
+- In case, if the `_request_timeout` value is not set, then the default value is **`None`** & socket will have no timeout.
+
+  ***Refer:***
+  - [https://docs.python.org/2/library/socket.html#socket.getdefaulttimeout](https://docs.python.org/2/library/socket.html#socket.getdefaulttimeout)
+
+- It is recommended to set this timeout value to a lower number (for eg. ~ maybe 60 seconds.
```
yliaog, line 67: add a right parenthesis at the end?

examples/watch/timeout-settings.md
```diff
@@ -0,0 +1,68 @@
...
+- In case of network outage, leading to dropping all packets with no RST/FIN, the timeout value (in seconds) determined by the `request_timeout` argument, would be the time duration for how long the client will wait before realizing & dropping the connection.
+
+- When the timeout happens, an exception will be raised, for eg. ~
+  
+  `urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='localhost', port=6443): Read timed out.`
+  
+- In case, if the `_request_timeout` value is not set, then the default value is **`None`** & socket will have no timeout.
+
+  ***Refer:***
+  - [https://docs.python.org/2/library/socket.html#socket.getdefaulttimeout](https://docs.python.org/2/library/socket.html#socket.getdefaulttimeout)
```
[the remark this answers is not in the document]
Priyankasaggu11929: noted. Thank you!

examples/watch/timeout-settings.md
Priyankasaggu11929: The similar `kube-apiserver` doc nolonger exist now in the recent release tags. 

I'm replacing this reference by this new link ~ https://github.com/kubernetes/kubernetes/blob/release-1.20/staging/src/k8s.io/apiserver/pkg/server/config.go#L320

examples/watch/timeout-settings.md
Priyankasaggu11929: Noted. Thank you!

examples/watch/timeout-settings.md
Priyankasaggu11929: ack. Thank you!

examples/watch/timeout-settings.md
Priyankasaggu11929: I saw the first reference in this comment here ~ https://github.com/kubernetes-client/python-base/pull/36#discussion_r146445193

Here's the reference from the Python client API ~ https://github.com/kubernetes-client/python/blob/master/kubernetes/client/api_client.py#L336-L339

Priyankasaggu11929: I'll make it more specific in the timeout_settings.md documentation (in the PR)~

the second input format could be a pair (tuple) of (connection, read) timeouts.

examples/watch/timeout-settings.md
Priyankasaggu11929: noted. Thanks for pointing.

examples/watch/timeout-settings.md
Priyankasaggu11929: Re: 
> what kind of network outage would cause this?

I'm not sure how to define or specify the  **what kind of network outage** part. Could you please phrase it or point me somewhere that talks more about the kind of network outage, I'll read more & try to add a proper definition thee.

Thank you! 

Re:
> It hangs when not accounting for the client side timeout described below, right?

I might be wrong with my understanding, but so far, I understand that it is the server side timeout (i.e. value of argument `timeout_seconds`, that would have no effect in case of the said network outage)

Just a pointer from this comment here ~ https://github.com/kubernetes-client/python/issues/1148#issuecomment-626184613

examples/watch/timeout-settings.md
```diff
@@ -0,0 +1,71 @@
...
+- In case, if the `_request_timeout` value is set, this argument can accept 2 types of input values ~
+    - float, 
+    - a tuple (with a length of 2)
+    
+  ***Refer***
+  - *[https://github.com/kubernetes-client/python/blob/v17.17.0/kubernetes/client/api_client.py#L336-L339](https://github.com/kubernetes-client/python/blob/v17.17.0/kubernetes/client/api_client.py#L336-L339)*
+ 
+  If it is tuple input type i.e. a pair of values like (connection, read), then the first value will be ignored. 
+
+- In case of network outage, leading to dropping all packets with no RST/FIN, the timeout value (in seconds) determined by the `request_timeout` argument, would be the time duration for how long the client will wait before realizing & dropping the connection.
```
yliaog, line 59: i think 'realizing' can be removed

examples/watch/timeout-settings.md
Priyankasaggu11929: @yliaog, noted.

So, I'll remove the following line from the documentation ~ 

>   If it is tuple input type i.e. a pair of values like (connection, read), then the first value will be ignored.

examples/watch/timeout-settings.md
Priyankasaggu11929: ack. Removing the word `realizing` & leaving the sentence as below:

> In case of network outage, leading to dropping all packets with no RST/FIN, the timeout value (in seconds) determined by the `request_timeout` argument, would be the time duration for how long the client will wait before dropping the connection.

examples/watch/timeout-settings.md
Priyankasaggu11929: @yliaog, no hurries at all. Just pointing in case, it was missed!

could you please take a look at this review comment too, as what more could be done to improve this section here?

Thank you so much!

yliaog: i think probably add the condition that client side timeout is not specified, something like below:

this timeout value will have no effect & the client will hang indefinitely without raising any exception if client side timeout (see below) is not specified

Priyankasaggu11929: @yliaog, Thank you for the pointers.

I'm rephrasing it as following: ~

> In case of a network outage, the server side timeout value will have no effect & the client will hang indefinitely without raising any exception. Note, that this is the case provided when there is no other client-side timeout (i.e., `_request_timeout`) value specified.
>
> See the section below for information on `client side timeout`

Priyankasaggu11929:
@yliaog, I've updated the PR with the suggested review changes.

yliaog:
thanks for the pr
/lgtm
/approve

roycaihw:
travis-ci.org is in read-only mode

PR #1306, opened 2020-11-05
Update pyyaml from 3.1.2 to 5.4.1

palnabarun:
5.3.1 fixed partially vulnerabilities disclosed in CVE-2020-1747.
A complete fix was debated at yaml/pyyaml#420 and eventually got patched in 5.4.1

Changeset: https://github.com/yaml/pyyaml/compare/3.12...5.4.1

I skimmed through the changeset and didn't see any apparent breaking changes.
But I will go through that in more detail. Putting a hold due to the same.

/hold

PR #1480, opened 2021-05-28
Drop python 3.5

palnabarun:
#### What type of PR is this?

/kind cleanup

Python 3.5 was EOL'ed on 13 Sep 2020. We shouldn't support Python versions that don't get security updates anymore. ([Ref 1][1], [Ref 2][2])

While going through updating PyYAML due to a CVE fix (#1306), I realized that our dependencies may have dropped support for out-of-support Python minor versions. PyYAML doesn't support Python 3.5 anymore with their latest release.

If we are to fix the CVE, we need to drop Python 3.5. I think we can drop as part of releasing client version v18.y.0b1 as Python doesn't break backward compatibility within minor versions.

[1]: https://devguide.python.org/#status-of-python-branches
[2]: https://endoflife.date/python

None

```release-note
NONE
```

None

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


roycaihw:
The test is constantly failing due to #1485

palnabarun:
Ummmm, this is weird. Force pushing to a branch where the PR is lgtm'ed and approved merged the PR.

I think this isn't expected. Is it?

palnabarun:
(Although the changes proposed in the PR are not very significant)

yliaog:
no, it is not expected, lgtm label was removed, how could it get merged?

PR #1520, opened 2021-08-06
Update release docs with more info

scottilee:
#### What type of PR is this?

/kind documentation

Adds more detailed information to the release process.

```release-note
NONE
```

### DIFF
--- a/devel/release.md
+++ b/devel/release.md
@@ -15,7 +15,11 @@ report issues. Typically, there is a single alpha or beta release, but if there
 are a higher than expected number of issues there can be multiple releases
 (e.g, a2 or b2).
 
-## Create or update release branch
+## 1. Update submodules
+
+Update submodules by referring to this [link](https://github.com/kubernetes-client/python/blob/master/devel/submodules.md#update-submodule). Commit the changes and open a pull request.
+
+## 2. Create or update release branch
 
 The release branch name should have release-x.x format. All minor and pre-releases
 should be on the same branch. To update an existing branch with master (only for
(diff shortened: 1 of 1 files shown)


devel/release.md
```diff
@@ -15,7 +15,17 @@ report issues. Typically, there is a single alpha or beta release, but if there
 are a higher than expected number of issues there can be multiple releases
 (e.g, a2 or b2).
 
-## Create or update release branch
+## 1. Update submodules
+
+Update submodules by running the following script:
```
roycaihw, line 20: could you link to https://github.com/kubernetes-client/python/blob/master/devel/submodules.md#update-submodule instead? In that way we make sure the instructions don't diverge in the future

devel/release.md
```diff
@@ -55,17 +65,24 @@ command:
...
 apply the manual fixes.***
 
 ```bash
 git push upstream $RELEASE_BRANCH
 ```
 
-## Hot issues
+## 4. Hot issues
+
+You can use the `scripts/apply-hotfixes.sh` script to apply the fixes below in
```
roycaihw, line 84: please add a note that the manual fix steps below are deprecated and exist only for documentation purpose.

devel/release.md
scottilee: Fixed.

devel/release.md
scottilee: Fixed.

scottilee:
Addressed comments.

roycaihw:
/lgtm
/approve

Thanks!

PR #1556, opened 2021-09-25
Create list_and_delete_pvc.py

MederD:
This python script lists and deletes persistent volume claims depending how old are they.


MederD:
```release-note
NONE
```

yliaog:
there are other examples covering 'list' of objects api call, this particular example is not very helpful, IMO.

PR #1563, opened 2021-10-05
Added info about change in versioning scheme

### LINKED ISSUE #1244: [RFC] Syncing Kubernetes Client versions with upstream Kubernetes versions
## Current Scenario

The Kubernetes Python client follows a versioning schema `x.y.z{a|b}N` where x is `Kubernetes Minor Version - 4`. For example, the Kubernetes Python client 12.0.0 is based on Kubernetes 1.16.

## What other clients do

- Go: client-go used to use a versioning scheme like us, but they eventually moved to semver compatible versions - v0.y.z where y and z are from Kubernetes versions 1.y.z. [Ref1](https://github.com/kubernetes/kubernetes/issues/84608) [Ref2](https://github.com/kubernetes/enhancements/pull/1350)
- Java: A similar schema like but the Java client major version number is Kubernetes minor version number - 8. For example, Java client 9.0.0 is based on Kubernetes 1.17. 

## Proposed Versioning Schemes

We can have a versioning scheme similar to client-go. Kubernetes 1.x.y would correspond to Python client 0.x.y.
Or, the versions can be 1.x.y exactly equal to the Kubernetes versions.

## Reasons

To make the numbering coherent and reduce confusion.

/assign

LevonTostig:
According to https://github.com/kubernetes-client/python/issues/1244 the information about the change in the versioning scheme should be added one last time in the version 19.0.0. Did that. That way issue 1244 could also be closed.

#### What type of PR is this?

/kind documentation

The PR adds the information about the changed versioning scheme to the change log one last time as specified in issue #1244 .

Fixes #1244

```release-note
NONE
```

```docs

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Kubernetes API Version: v1.19.15
 
+**Important Information:**
+
+- The Kubernetes Python client versioning scheme has changed. The version numbers used till Kubernetes Python Client v12.y.z lagged behind the actual Kubernetes minor version numbers. From this release, the client is moving a version format `vY.Z.P` where `Y` and `Z` are respectively from the Kubernetes version `v1.Y.Z` and `P` would incremented due to changes on the Python client side itself. Ref: https://github.com/kubernetes-client/python/issues/1244
+
 ### Bug Fix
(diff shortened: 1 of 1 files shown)


dkpeall:
Hi Sorry this is my first attempt here, please let me know what else I need to do.  I've signed the CLA but its not updating here?

vijaysr4 (approved):

PR #1624, opened 2021-11-16
add param _response_type for deserialized object to methods of Custom…

wingzzzz:
…ObjectsApi. add register_custom_model method

#### What type of PR is this?
/kind design
/kind api-change

There is no way in CustomObjectsApi for deserializing custom resource with custom models like k8s models(such as V1Deployment). User has to use dict type to handle custom resource. And I think it's not a good practise. So I add a method called register_custom_model for register custom model to kubernetes.client.models. (kubernetes\client\api_client.py   line 292: klass = getattr(kubernetes.client.models, klass)   get the class from kubernetes.client.models) And I add the param _response_type for methods of CustomObjectsApi, so user can pass custom model to deserialize the custom object.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/kubernetes/client/api/custom_objects_api.py
+++ b/kubernetes/client/api/custom_objects_api.py
@@ -93,6 +93,7 @@ class CustomObjectsApi(object):
                                  number provided, it will be total request
                                  timeout. It can also be a pair (tuple) of
                                  (connection, read) timeouts.
+        :param _response_type: class for deserialized object.
         :return: tuple(object, status_code(int), headers(HTTPHeaderDict))
                  If the method is called asynchronously,
                  returns the request thread.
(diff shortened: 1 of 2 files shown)


kubernetes/client/api/custom_objects_api.py
```diff
@@ -93,6 +93,7 @@ def create_cluster_custom_object_with_http_info(self, group, version, plural, bo
                                  number provided, it will be total request
                                  timeout. It can also be a pair (tuple) of
                                  (connection, read) timeouts.
+        :param _response_type: class for deserialized object.
```
yliaog, line 96: kubernetes/client/api/custom_objects_api.py:  the file is generated, any change here would be overwritten by the  generation of the file next time. please make changes in the generator, https://github.com/openapitools/openapi-generator

roycaihw:
> There is no way in CustomObjectsApi for deserializing custom resource with custom models like k8s models(such as V1Deployment)

Trying to understand the use case here. Why not use the typed client?

roycaihw:
Closing due to obsolete. Please feel free to reopen if needed!
/close

PR #1672, opened 2022-01-08
removed submodule

yliaog:
#### What type of PR is this?
/kind bug

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "kubernetes/base"]
-	path = kubernetes/base
-	url = https://github.com/kubernetes-client/python-base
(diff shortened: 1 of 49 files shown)


roycaihw:
Why do we have these many (321) commits?

yliaog:
the commits are old, they are added to preserve the commit history

leoskyrocker:
@yliaog @roycaihw 
it's not easy to see the actual changes here.
The wrapper function is great, but it seems like there is not actual documentation around it.
I would use it if we're saying we can commit to supporting it.

I'm wondering if it is going to be likely for the wrapper function to suddenly disappear/change drastically in the upcoming versions.
Alternatively, we could update the documentation if we think it is here to stay (and stay consistent) for the foreseeable future?

roycaihw:
@leoskyrocker From the slack thread I think you mean the behavior of `load_config` loading out-of-cluster or in-cluster kubeconfig.

The behavior was inspired by client-go. We don't plan to drop this behavior. It would be nice if we document the behavior. cc @MoShitrit

Issue #1626, opened 2021-11-17
async creation of custom resources (CR) using the CRD is failing


venukarnati92:
Async creation of custom resources (CR) using the CRD is failing for dynamic Client create call.

Synchronous Implementation with this example is working fine https://github.com/kubernetes-client/python/blob/master/examples/dynamic-client/namespaced_custom_resource.py

Now to make creation process as async update line number 192 on the above example to the following

`ingressroute_api.create(body=ingressroute_manifest_first, namespace=namespace_first, async_req=True)`

Error Message

`(ansible) 21-11-16 2:45:14 venu.karnati@ABOUTPIANOORIGIN:~/Desktop/gitProject/python/examples/dynamic-client$ python namespaced_custom_resource.py Traceback (most recent call last): File "namespaced_custom_resource.py", line 250, in <module> main() File "namespaced_custom_resource.py", line 192, in main ingressroute_api.create(body=ingressroute_manifest_first, namespace=namespace_first, async_req=True) File "/Users/venu.karnati/.pyenv/versions/ansible/lib/python3.6/site-packages/kubernetes/dynamic/client.py", line 119, in create return self.request('post', path, body=body, **kwargs) File "/Users/venu.karnati/.pyenv/versions/ansible/lib/python3.6/site-packages/kubernetes/dynamic/client.py", line 62, in inner return serializer(self, json.loads(resp.data.decode('utf8'))) AttributeError: 'ApplyResult' object has no attribute 'data'`

**Environment**:
- Kubernetes version 1.21.2
- MacOS 11.6):
- Python 3.6.8
- kubernetes          18.20.0

roycaihw:
cc @fabianvf @roycaihw

roycaihw:
In our typed clients, the syntax for using async_req is:
```
  thread = api.method(body, async_req=True)
  result = thread.get()
```
For example, the [CRD client](https://github.com/kubernetes-client/python/blob/549482d8842a47c8b51122422e879e7cc497bf88/kubernetes/client/api/apiextensions_v1_api.py#L43-L46).

Could you try if the same works with the dynamic client?

venukarnati92:
@roycaihw 
On further debugging I was able to find that when you make an asynchronous call response doesn't have data attribute.I can see request is launched but client.py is unable to get the response details.Looks we need to have separate logic for response parsing when the call is asynchronous .

@roycaihw The above suggestion will still fail because the error is happening when we make a create call itself.

[...]

PR #1697, opened 2022-02-09
Async fix for Dynamic Client

venukarnati92:
/kind bug

```release-note
Notable feature additions for async creation of Custom resources using dynamic Client
```

What this PR does / why we need it?

async creation of custom resources (CR) using the Custom Resource Definition is failing with the following error

AttributeError: 'ApplyResult' object has no attribute 'data'

Fix for the issue:

For Async call meta_request(func) is failing because response object is different for async vs sync call.The following fix will make sure that response object is same for synchronous and asynchronous calls.

Which issue(s) this PR fixes:
https://github.com/kubernetes-client/python/issues/1626

Special notes for reviewer:

Usually for all async calls we get the response when we make the following call thread.get() but incase of dynamic client async calls I don't see that behavior because of the meta_request(func) is already parsing the response.

NONE

### DIFF
--- a/kubernetes/base/dynamic/client.py
+++ b/kubernetes/base/dynamic/client.py
@@ -267,7 +267,7 @@ class DynamicClient(object):
         # Authentication setting
         auth_settings = ['BearerToken']
 
-        return self.client.call_api(
+        api_response = self.client.call_api(
             path,
             method.upper(),
             path_params,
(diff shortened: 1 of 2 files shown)


venukarnati92:
As suggested by @yliaog , moving the following PR https://github.com/kubernetes-client/python-base/pull/269 to this repo.

PR #1732, opened 2022-03-01
Support setting the _request_timeout parameter for the dynamic client

### LINKED ISSUE #1726: Support request timeout in dynamic client
I would like to use the dynamic client to interact with the Kubernetes API. But at the moment it looks like it is not possible to configure a request timeout for it.

For the normal api client it is possible to configure a timeout via the `_request_timeout` parameter as can be seen here:
https://github.com/kubernetes-client/python/blob/dd7417a0b889371b4a855f27ffffcfbfa09c2ad3/kubernetes/client/api_client.py#L305-L310

But when I looked at the `request` method of the dynamic client I saw that this option can not be passed to its api client
https://github.com/kubernetes-client/python/blob/dd7417a0b889371b4a855f27ffffcfbfa09c2ad3/kubernetes/base/dynamic/client.py#L270-L283

I would like to pass the `_request_timeout` parameter as an extra `kwargs` to the request methods of the dynamic client so that it is then pass on to the actual api client.

If I missed anything please let me know otherwise I would go ahead and create a PR for this.

philipp-sontag-by:
#### What type of PR is this?
/feature

This PR adds support for the `_request_timeout` parameter to the dynamic client. The parameter can be passed to the dynamic clients requests method and is then passed on the api client.

Fixes #1726 

I am not sure how to properly test this change since I am new to the code base. Any hints would be appreciated.

```release-note
The dynamic client now support the `_request_timeout` parameter to configure connection and request timeouts.
```

### DIFF
--- a/kubernetes/base/dynamic/client.py
+++ b/kubernetes/base/dynamic/client.py
@@ -279,7 +279,8 @@ class DynamicClient(object):
             files=local_var_files,
             auth_settings=auth_settings,
             _preload_content=False,
-            _return_http_data_only=params.get('_return_http_data_only', True)
+            _return_http_data_only=params.get('_return_http_data_only', True),
+            _request_timeout=params.get('_request_timeout')
         )
         if params.get('async_req'):
             return api_response.get()


roycaihw:
Is this still in draft state. Please let us know when it's ready for review. Thanks!

philipp-sontag-by:
Hey @yliaog and @fabianvf,
I would appreciate a review here when you have some time.

Issue #1715, opened 2022-02-16
'self' word in CRD fails with DynamicApiError


Akasurde:
**What happened (please include outputs or screenshots)**:
When trying to create a CRD with `self` keyword in `spec.properties` raises DynamicApiError

```
__init__() got multiple values for argument 'self'
```

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: stars.example.crd.com
spec:
  group: example.crd.com
  scope: Namespaced
  names:
    kind: Star
    listKind: StarList
    plural: stars
    singular: star
  subresources:
    status: {}
  validation:
    openAPIV3Schema:
      required: ["spec"]
      properties:
        spec:
          required: ["type","location","self"]
          properties:
            type:
              type: "string"
              minimum: 1
            location:
              type: "string"
              minimum: 1
            self:
              type: "string"
              minimum: 1
  versions:
    - name: v1alpha1
      served: true
      storage: true

```

**What you expected to happen**:
Python library should not fail and create the CRD specified.

**How to reproduce it (as minimally and precisely as possible)**:
Use Python code 

```python
k8s_obj = resource.create(definition, namespace=namespace, **params).to_dict()
```

or use 

Ansible Playbook - 

```yaml
---
- hosts: localhost
  gather_facts: no
  tasks:
    - name: k8s apply bugged crds
      kubernetes.core.k8s:
         state: present
         definition: "{{ lookup('file', 'crd.yml') }}"
         namespace: keyloak
```

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (`kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.3", GitCommit:"816c97ab8cff8a1c72eccca1026f7820e93e0d25", GitTreeState:"clean", BuildDate:"2022-01-25T21:17:57Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:53:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
```

- OS (e.g., MacOS 10.13.6):
macOS 11.6.1

- Python version (`python --version`)
3.9.10

[...]

PR #1738, opened 2022-03-04
fix issue when trying to build ResourceField with argument name 'self'

abikouo:
#### What type of PR is this?

Fixes #

#1715 

```release-note

```

```docs

```

### DIFF
--- a/kubernetes/base/dynamic/resource.py
+++ b/kubernetes/base/dynamic/resource.py
@@ -298,7 +298,7 @@ class ResourceInstance(object):
 
     def __deserialize(self, field):
         if isinstance(field, dict):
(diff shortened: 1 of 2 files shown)


yliaog:
could you please add a test?

abikouo:
> could you please add a test?

done!!

fabianvf (approved):
very technically breaking but only on fairly internal code (if this was go it would not be exported), so I have no issue with this at all. Thanks for the fix @abikouo !

geoberle:
thanks @abikouo for this fix!

@yliaog, @fabianvf this fix is not part of a release yet. can i help with anything?

PR #1754, opened 2022-03-17
Update select poll logic to support eventlet.

### LINKED ISSUE #1753: WSClient.update() fails on select.poll() when running in eventlet on Linux

WSClient.update() was modified by PR https://github.com/kubernetes-client/python-base/pull/268 to use select.poll() instead of select.select, which breaks when running under eventlet on Linux:

  File "/usr/local/lib/python3.8/site-packages/kubernetes/stream/ws_client.py", line 182, in update
    poll = select.poll()
AttributeError: module 'select' has no attribute 'poll'
module 'select' has no attribute 'poll'

eventlet monkey_patches the select module and does not support the poll() method

Adding a check on getattr(select, "poll", None) is None to the if statement and reverting to select.select when there is no poll attribute will fix the problem.

I'll push a PR

No exception should occur

**Environment**:
- Kubernetes version (`kubectl version`): 1.19, 1.21
- OS (e.g., MacOS 10.13.6): CentOS 7
- Python version (`python --version`)  3.8
- Python client version (`pip list | grep kubernetes`)  23.3.0

bobh66:
#### What type of PR is this?
/kind bug
/kind regression

Fixes exception in WSClient.update() caused by select not having a poll() attribute when running under eventlet. Introduced by PR https://github.com/kubernetes-client/python-base/pull/268

Fixes #1753 

```release-note
NONE
```

```docs
```

### DIFF
--- a/kubernetes/base/stream/ws_client.py
+++ b/kubernetes/base/stream/ws_client.py
@@ -179,7 +179,9 @@ class WSClient:
         #                   efficient as epoll. Will work for fd numbers above 1024.
         # select.epoll()  - newest and most efficient way of polling.
         #                   However, only works on linux.
-        if sys.platform.startswith('linux') or sys.platform in ['darwin']:
+        #                   Will not work on Linux when eventlet is being used.
+        if ((sys.platform.startswith('linux') and getattr(select, "poll", None) is not None) or
+                sys.platform in ['darwin']):
             poll = select.poll()
             poll.register(self.sock.sock, select.POLLIN)
             r = poll.poll(timeout)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v23.3.0
+
+Kubernetes API Version: v1.23.4
+
+
 # v23.3.0b1
 
 Kubernetes API Version: v1.23.4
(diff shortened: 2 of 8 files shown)


kubernetes/base/stream/ws_client.py
```diff
@@ -179,7 +179,9 @@ def update(self, timeout=0):
         #                   efficient as epoll. Will work for fd numbers above 1024.
         # select.epoll()  - newest and most efficient way of polling.
         #                   However, only works on linux.
-        if sys.platform.startswith('linux') or sys.platform in ['darwin']:
+        #                   Will not work on Linux when eventlet is being used.
+        if ((sys.platform.startswith('linux') and getattr(select, "poll", None) is not None) or
```
yliaog, line 183: eventlet supports epoll, does it make sense to use epoll

bobh66: If that's a better solution I'm happy to rework the PR.  I'm not very familiar with select and poll but if epoll is equivalent I'll update the change.

Akasurde, line 183: You can take inspiration from https://github.com/python/cpython/pull/21257 and https://github.com/ansible/ansible/pull/70352

roycaihw:
Closing due to inactivity. 
/close

PR #1781, opened 2022-04-18
Set cwd to None if empty string for ExecProvider

### LINKED ISSUE #1774: ExecPlugins are broken when using KubeConfigs loaded from dict

We use https://github.com/kubernetes-client/python/blob/master/kubernetes/base/config/kube_config.py#L820 to load Kubeconfigs from memory in Python, these Kubeconfigs have a configuration to use an Exec plugin to get an auth token (specifically we are trying to auth with an EKS cluster.)

https://github.com/kubernetes-client/python-base/pull/275/files#diff-a994e720c2c925ca342b251f2d057ab3a730a1bdc166e5766249a30bdf170a47R73 this PR to python-base attempts to pass the cwd of the config file to the exec plugin, however the cwd of a config file loaded from a dict is empty string (https://github.com/kubernetes-client/python/blob/master/kubernetes/base/config/kube_config.py#L542). When empty string is passed to subprocess.Popen, subprocess.Popen errors with "ERROR:root:[Errno 2] No such file or directory: ''" and does not run the auth plugin. 

Commenting out https://github.com/kubernetes-client/python-base/pull/275/files#diff-a994e720c2c925ca342b251f2d057ab3a730a1bdc166e5766249a30bdf170a47R73 fixed our auth issues completely, but the longterm fix is probably checking if the base_path is empty string and if so, pass None as the cwd to Popen. 

The issue means that anyone using EKS with load_kube_config_from_dict has been broken since this PR was merged. 

Run the ExecProvider and return a token
Use load_kube_config_from_dict with an Authentication configuration using "command" for the user

**Environment**:
- OS (e.g., MacOS 10.13.6):
Linux
3.8.10
23.3.0

jeid64:
#### What type of PR is this?

/kind bug

If using ExecProvisioner with an inmemory kubeconfig, set cwd to None so that subprocess.Popen works, instead of empty string which breaks. 

Fixes #1774 

NONE

None

### DIFF
--- a/kubernetes/base/config/exec_provider.py
+++ b/kubernetes/base/config/exec_provider.py
@@ -53,7 +53,8 @@ class ExecProvider(object):
                 value = item['value']
                 additional_vars[name] = value
             self.env.update(additional_vars)
-        self.cwd = cwd
+        
+        self.cwd = cwd if cwd else None
 
     def run(self, previous_response=None):
         kubernetes_exec_info = {
--- a/kubernetes/base/config/exec_provider.py
+++ b/kubernetes/base/config/exec_provider.py
@@ -54,7 +54,7 @@ class ExecProvider(object):
                 additional_vars[name] = value
             self.env.update(additional_vars)
         
-        self.cwd = cwd if cwd else None
+        self.cwd = cwd or None
 
     def run(self, previous_response=None):
         kubernetes_exec_info = {


yliaog:
could you please add a test?

peng19940915:
i found this problem too

kubernetes/base/config/exec_provider.py
```diff
@@ -53,7 +53,8 @@ def __init__(self, exec_config, cwd):
                 value = item['value']
                 additional_vars[name] = value
             self.env.update(additional_vars)
-        self.cwd = cwd
+        
+        self.cwd = cwd if cwd else None
```
iciclespider, line 57:
```suggestion
        self.cwd = cwd or None
```

jeid64:
@yliaog what would the test be? just checking for nullability?

yliaog:
add a test case in exec_provider_test.py to check it is set as intended.

PR #1785, opened 2022-04-30
Fix kubernetes port forwarding from hanging in certain edge cases.

iciclespider:
It turns out SSL sockets can buffer data such that the select method is not aware of it. See:

  https://docs.python.org/3/library/ssl.html#notes-on-non-blocking-sockets
  https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.pending

/kind bug
/kind failing-test


kubernetes/e2e_test/test_client.py
```diff
@@ -230,10 +230,6 @@ def test_exit_code(self):
         resp = api.delete_namespaced_pod(name=name, body={},
                                          namespace='default')
 
-    # Skipping this test as this flakes a lot
-    # See: https://github.com/kubernetes-client/python/issues/1300
-    # Re-enable the test once the flakiness is investigated
-    @unittest.skip("skipping due to extreme flakiness")
```
yliaog, line 236: i have not read the test carefully, but figured it's better to just ask: is there a bug in the this test you're fixing, if there is, could you please describe the bug in the test a bit?

kubernetes/base/stream/ws_client.py
```diff
@@ -353,69 +353,76 @@ def _proxy(self):
...
+                                port_number = six.byte2int(frame.data[1:2]) + (six.byte2int(frame.data[2:3]) * 256)
+                                if port_number != port.port_number:
+                                    raise RuntimeError(
+                                        "Unexpected port number in initial channel frame: %s" % port_number
+                                    )
+                                channel_initialized[channel] = True
+                        elif opcode not in (ABNF.OPCODE_PING, ABNF.OPCODE_PONG, ABNF.OPCODE_CLOSE):
+                            raise RuntimeError("Unexpected websocket opcode: %s" % opcode)
+                        if not (isinstance(self.websocket.sock, ssl.SSLSocket) and self.websocket.sock.pending()):
+                            pending = False
```
yliaog, line 405: looks good to me. thanks for fixing it.

kubernetes/e2e_test/test_client.py
iciclespider: From the PR description: #1300

iciclespider: The "bug" was that SSL sockets can have a "pending" state where data is available, but the select.select method return that the socket has data.

yliaog: I get the "bug" you're fixing in the code [ws_client.py], the fix looks good.

The question is about the changes you're making in this test file, are these changes required to make the test PASS?

iciclespider: @yliaog , I am sorry, now I see what you are asking.

The test as I originally wrote it was poorly implemented, especially the loops obtaining data. They made assumptions about data arriving within 1 second. I reworked the socket select logic to have an overall timeout value and to not make assumptions about what is the maximum amount of time data will arrive on the sockets.

I suspect that such assumptions could have been contributing to the flakey'ness of the test, but I am not 100% sure.

yliaog:
Thanks for the explanation and fixing the issue.

/lgtm
/approve

PR #1796, opened 2022-05-08
FIX:  Update watch resource_version on BOOKMARK events. 

### LINKED ISSUE #1729: Add support for bookmark events

In kubernetes-client/python-base#234 the parsing of BOOKMARK events has been bypassed ([and still is as of today](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/watch/watch.py#L103)). But those events exist for a reason and can be useful in some circumstances. A motivational example is this Airflow issue: https://github.com/apache/airflow/issues/21087.

As Airflows scheduler can be stopped and restarted, while many events occur on k8s, it would need to process BOOKMARK to recover its state. Sadly because those are ignored by the Python client, it cannot. Which manifests itself in varied ways (see the different issues referenced in the ticket above)

We'd like for the bookmark event to be parsed and yield by the watcher, so it is possible for clients to consume them.

(copied from kubernetes-client/python-base#286)

snjypl:
The BOOKMARK feature was not getting used properly. 

it is expected that on the BOOKMARK event, the watch should update it's resource_version to the one in the BOOKMARK event. 

Currently the BOOKMARK event was considered same as ERROR, because of which the watch resource_version was not updated. 

we are now decoding BOOKMARK event as a object/dict, and also updating the watch.resource_version  

/kind bug

Fixes #1729 
Related: 
[#23578](https://github.com/apache/airflow/pull/23578)
[#21087](https://github.com/apache/airflow/issues/21087)

```release-note
NONE
```

### DIFF
--- a/kubernetes/base/watch/watch.py
+++ b/kubernetes/base/watch/watch.py
@@ -96,11 +96,11 @@ class Watch(object):
     def unmarshal_event(self, data, return_type):
         js = json.loads(data)
         js['raw_object'] = js['object']
-        # BOOKMARK event is treated the same as ERROR for a quick fix of
-        # decoding exception
-        # TODO: make use of the resource_version in BOOKMARK event for more
-        # efficient WATCH
-        if return_type and js['type'] != 'ERROR' and js['type'] != 'BOOKMARK':
+        if return_type and js['type'] != 'ERROR':
+            if js['type'] == 'BOOKMARK':
+                # treat BOOKMARK as a custom object, so that resource_version
+                # is updated
+                return_type = object
             obj = SimpleNamespace(data=json.dumps(js['raw_object']))
             js['object'] = self._api_client.deserialize(obj, return_type)
             if hasattr(js['object'], 'metadata'):
--- a/kubernetes/base/watch/watch_test.py
+++ b/kubernetes/base/watch/watch_test.py
@@ -167,6 +167,42 @@ class WatchTests(unittest.TestCase):
         # more strict test with worse error message
         self.assertEqual(fake_api.get_namespaces.mock_calls, calls)
 
+    def test_watch_resource_version_with_bookmark(self):
+        """
+        test the resource_version get updated for bookmark events
+
+        """
+        fake_resp = Mock()
+        fake_resp.close = Mock()
+        fake_resp.release_conn = Mock()
+        values = [
+            '{"type": "ADDED", "object": {"metadata": {"name": "test1",'
+            '"resourceVersion": "1"}, "spec": {}, "status": {}}}\n',
+            '{ "type": "BOOKMARK", "object": {"kind": "Pod", "apiVersion": "v1",'
+            '"metadata": {"resourceVersion": "3"} }}\n',
+            '{"type": "ADDED", "object": {"metadata": {"name": "test3",'
+            '"resourceVersion": "4"}, "spec": {}, "status": {}}}\n']
+
+        def get_values(*args, **kwargs):
+            return values
+
+        fake_resp.stream = Mock(
+            side_effect=get_values)
+
+        fake_api = Mock()
+        fake_api.get_namespaces = Mock(return_value=fake_resp)
+        fake_api.get_namespaces.__doc__ = ':return: V1NamespaceList'
+
+        w = Watch()
+        iterations = 3
+
+        for c, e in enumerate(w.stream(fake_api.get_namespaces,
+                                       resource_version="0")):
+            self.assertEqual(w.resource_version,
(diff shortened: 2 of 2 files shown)


snjypl:
since the resoureVersion is the only thing that matters in an BOOKMARK event. 
i don't think, it is required or expected to convert the bookmark event to the `return_type`  since bookmark event will not have all the required fields. 

when we treat it as a ERROR, we skip updating the watcher.resource_version to the one in bookmark event, which defeats the purpose of having the BOOKMARK feature. 

and as far as de-serializing the bookmark event is concerned, the client should be fine with a dict object. the client is expected to check the  event['TYPE'] and deal with the bookmark event as  a dict

ecerulm:
The current behaviour when `allow_watch_bookmarks=True` and `resource_version=<alreadyexpiredresourceversion` is to raise an `ApiException (410) `, will this PR change that behaviour?

snjypl:
> The current behaviour when `allow_watch_bookmarks=True` and `resource_version=<alreadyexpiredresourceversion` is to raise an `ApiException (410) `, will this PR change that behaviour?

no, it won't. it is the expected behavior.

ecerulm:
> /assign @ecerulm

what do you want me to do here?

roycaihw:
It looks like you've been commenting on this PR already. Would you mind doing a first round of review? @ecerulm

ecerulm:
@roycaihw I'm not familiar with the setup of this project and how contributions are made, this PR is based on master which seems 23.0.1 , but the latest release is 23.3.0 (tag v23.3.0) , are PR supposed to be based on the master, on the tag v23.3.0 or on the tag 23.6.0?

ecerulm:
@roycaihw I'm not familiar with the setup of this project and how contributions are made, this PR is based on master which seems 23.0.1 , but the latest release is 23.3.0 (tag v23.3.0) , are PR supposed to be based on the master, on the tag v23.3.0 or on the tag 23.6.0?

ecerulm:
If I understand right , before with `allow_watch_bookmarks=True' you could have a loop like this

```
for event in  watcher.stream(v1.list_namespaced_pod, namespace, allow_watch_bookmarks=True):
  print(event['object'].metadata.resource_version
```

after the introduction of this PR the following will raise an `AttributeError: 'dict' object has not attribute 'metadata'` when it reaches a bookmark event, `event['object'].metadata` is of type `V1ObjectMeta` in the other events but it's of type `dict` for bookmark event. 

That can potentially break the code of people already using `allow_watch_bookmarks=True` today when they upgrade to a version containing this PR code. 

I'm not sure about the rules about backwards compatibility that this projects follows, it may be ok. I think somebody else should review this,  @roycaihw , I'm not that familiar with the code base.

ecerulm:
If I understand right , before with `allow_watch_bookmarks=True' you could have a loop like this

```
for event in  watcher.stream(v1.list_namespaced_pod, namespace, allow_watch_bookmarks=True):
  print(event['object'].metadata.resource_version
```

after the introduction of this PR the following will raise an `AttributeError: 'dict' object has not attribute 'metadata'` when it reaches a bookmark event, `event['object'].metadata` is of type `V1ObjectMeta` in the other events but it's of type `dict` for bookmark event. 

That can potentially break the code of people already using `allow_watch_bookmarks=True` today when they upgrade to a version containing this PR code. 

I'm not sure about the rules about backwards compatibility that this projects follows, it may be ok. I think somebody else should review this,  @roycaihw , I'm not that familiar with the code base.

snjypl:
@ecerulm 

> If I understand right , before with `allow_watch_bookmarks=True' you could have a loop like this
> 
> ```
> for event in  watcher.stream(v1.list_namespaced_pod, namespace, allow_watch_bookmarks=True):
>   print(event['object'].metadata.resource_version
> ```
> 
No, even with the latest release  you will get the same error.  even now bookmark event is returned as dict. 

this PR does not change that. so it is not backward incompatible.

snjypl:
if the user had opted for the bookmark feature, the this would be the recommenced way of doing it:  

_Note: this is the existing behavior. this PR  does not change it._
```
 for event in  watcher.stream(v1.list_namespaced_pod, namespace, allow_watch_bookmarks=True):
           if event['TYPE'] == 'ERROR':
                 # handle error 
           if event['TYPE']  == 'BOOKMARK':
                 # handle bookmark 
           
           print(event['object'].metadata.resource_version

```

snjypl:
@roycaihw @fabianvf  @yliaog 

is there anything i could do to help with the review?  do i need to add any notes? 

it will be super great to have this feature.

snjypl:
@roycaihw  ping ! :D

PR #1833, opened 2022-06-16
Add interactive configuration to exec provider.

### LINKED ISSUE #1787: ExecProvider swallows interactive messages
When using [kubelogin](https://github.com/Azure/kubelogin) as an ExecProvider in kube config, upon initial use, kubelogin prints a message to stderr of the form:
```
To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code <device-code> to authenticate.
```
The current ExecProvider implementation swallows this message and blocks until the code times out and the kubelogin command returns (an exit code of 1).

The ExecProvider implementation to provide some method to access the message such that it is either accessible to the python application to pass to a user, or passed to the user to action on some standard bus (stderr?).

Have kubelogin installed
Have a kube config setup with exec using kubelogin:
```yaml
<other config values>
users:
- name: aad-user
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      command: kubelogin
      args: <kubelogin args depending on your azure auth setup>
```
*Don't authenticate through another means first (ie using kubectl)*
Attempt to auth using kubernetes library initially:
```python
from kubernetes import config

config.load_kube_config()
```

I don't believe this is kubelogin specific, but rather any exec auth provider that prompts for user action will have the same issue. Let me know if I'm off basis and this should be a kubelogin issue instead tho :)

**Environment**:
kubernetes (python): 23.3.0
kubelogin: v0.0.10
Beyond that not environment specific

brendandburns:
#### What type of PR is this?

/kind bug

Fixes #1787 

```release-note
NONE
```

### DIFF
--- a/kubernetes/base/config/exec_provider.py
+++ b/kubernetes/base/config/exec_provider.py
@@ -57,11 +57,12 @@ class ExecProvider(object):
         self.cwd = cwd or None
 
     def run(self, previous_response=None):
+        is_interactive = sys.stdout.isatty()
         kubernetes_exec_info = {
             'apiVersion': self.api_version,
             'kind': 'ExecCredential',
             'spec': {
-                'interactive': sys.stdout.isatty()
+                'interactive': is_interactive
             }
         }
         if previous_response:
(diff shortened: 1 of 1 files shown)


brendandburns:
Friendly ping on this PR cc @yliaog

yliaog:
sorry, missed this.

/lgtm
/approve

PR #1850, opened 2022-07-05
Fix DynamicClient.server_side_apply

kariya-mitsuru:
#### What type of PR is this?

/kind bug

`DynamicClient.server_side_apply` is obviously designed to accept a `dict` or a `ResourceInstance` as `body` like other methods.
(This is why the `name` and `namespace` parameters are treated as optional.)
However, if a `dict` or a `ResourceInstance` is passed actually, an error occurs because `RESTClientObject.rest` cannot interpret the `Content-Type` `application/apply-patch+yaml`.

So, modify `RESTClientObject.rest` to treat `application/apply-patch+yaml` as other json-based `Content-Type`s.

Yes.

```release-note
You can no longer specify an argument of type str or bytes for the body parameter of DynamicClient.server_side_apply.
Pass an argument of type dict or ResourceInstance instead.
Accordingly, the name and namespace parameters can now be omitted if the body argument contains appropriate metadata.
```

### DIFF
--- a/kubernetes/base/dynamic/test_client.py
+++ b/kubernetes/base/dynamic/test_client.py
@@ -15,7 +15,6 @@
 import time
 import unittest
 import uuid
-import json
 
 from kubernetes.e2e_test import base
 from kubernetes.client import api_client
@@ -527,9 +526,8 @@ class TestDynamicClient(unittest.TestCase):
                             'ports': [{'containerPort': 80,
                                             'protocol': 'TCP'}]}]}}
 
-        body = json.dumps(pod_manifest).encode()
         resp = api.server_side_apply(
-            name=name, namespace='default', body=body,
+            namespace='default', body=pod_manifest,
             field_manager='kubernetes-unittests', dry_run="All")
         self.assertEqual('kubernetes-unittests', resp.metadata.managedFields[0].manager)
 
--- a/kubernetes/client/rest.py
+++ b/kubernetes/client/rest.py
@@ -157,7 +157,8 @@ class RESTClientObject(object):
             if method in ['POST', 'PUT', 'PATCH', 'OPTIONS', 'DELETE']:
                 if query_params:
                     url += '?' + urlencode(query_params)
-                if re.search('json', headers['Content-Type'], re.IGNORECASE):
+                if (re.search('json', headers['Content-Type'], re.IGNORECASE) or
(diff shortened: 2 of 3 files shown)


kubernetes/client/rest.py
```diff
@@ -157,7 +157,8 @@ def request(self, method, url, query_params=None, headers=None,
             if method in ['POST', 'PUT', 'PATCH', 'OPTIONS', 'DELETE']:
                 if query_params:
                     url += '?' + urlencode(query_params)
-                if re.search('json', headers['Content-Type'], re.IGNORECASE):
+                if (re.search('json', headers['Content-Type'], re.IGNORECASE) or
```
fabianvf, line 160: Unfortunately this file is generated by the OpenAPI generator, I'm not sure what the process is for carrying patches to the generated files but usually these need to be fixed in the generator itself

@roycaihw what should we do here?

fabianvf (approved):
/lgtm

fabianvf:
@abikouo any thoughts?

kubernetes/base/dynamic/test_client.py
```diff
@@ -527,9 +526,8 @@ def test_server_side_apply_api(self):
                             'ports': [{'containerPort': 80,
                                             'protocol': 'TCP'}]}]}}
 
-        body = json.dumps(pod_manifest).encode()
         resp = api.server_side_apply(
-            name=name, namespace='default', body=body,
```
roycaihw, line 532: > body=body

Does this mean users can no longer pass JSON as the body? Is this considered a breaking change?

fabianvf: It would be a breaking change, though I'm not sure that we intended for this to accept a string in the first place. Could add a little bit of logic to server_side_apply to attempt to yaml load the content in order to preserve that behavior if it's what we want

roycaihw: > I'm not sure that we intended for this to accept a string in the first place

I agree. I'm okay with calling it out as a breaking change in the release notes. We could provide instruction/util to help users switch/keep the behavior, but the default behavior doesn't have to accept a string.

roycaihw:
/lgtm
/approve

/hold
please feel free to unhold https://github.com/kubernetes-client/python/pull/1850#issuecomment-1175362904

abikouo:
> @abikouo any thoughts?

lgtm, it makes sense!!

PR #1860, opened 2022-07-19
Deploy docs as wiki

### LINKED ISSUE #1557: Github frequently refuses to render CoreV1Api documentation
Right now all attempts to view the `CoreV1Api` documentation result in this:

kubernetes-client-docs

As the error message states, the reason for this is that the markdown file is too large.  Unfortunately, this makes it difficult enough to be effectively impossible to read the documentation, as you can no longer use the tag links to find what you are looking for.

Rishit-dagli:
#### What type of PR is this?

/kind documentation
/kind feature

This PR introduces a new GitHub action that pushes `kubernetes/docs/` to GitHub wiki. The current markdown files are often quite big and end up in errors when trying to open them on GitHub.

From https://github.com/kubernetes-client/python/issues/1557#issuecomment-955112451

> The Wiki page sometimes worked for me and sometimes didn't. I really like it though. Regardless of the rendering issue, I think it helps if we have a centralized place for installation, examples, API doc, contribution, etc.

More information on why this is needed can be found at #1557.

Fixes #1557

This PR first requires the use of Wikis to be enabled on this repository as well as create a home page for the wiki to allow pulling the wiki in the first place.

/cc @roycaihw 

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/.github/workflows/deploy-wiki.yaml
+++ b/.github/workflows/deploy-wiki.yaml
@@ -1,0 +1,37 @@
+name: Deploy Wiki
+
+on:
+  push:
+    paths:
+      - 'kubernetes/docs/**'
+    branches:
+      - master
+
+jobs:
+  deploy-wiki:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
+    - name: Install rsync
+      run: |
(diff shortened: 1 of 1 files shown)


roycaihw (commented):
/lgtm
/approve

PR #1861, opened 2022-07-21
fix issue #1782   Resolve load_kube_config TMP file issue

### LINKED ISSUE #1782: `certificate-authority-data` from kubeconfig fails for long-running process due to tmpfs cleanup
Long-running applications using a kubeconfig with `certificate-authority-data` encounter errors like:
```
Max retries exceeded with url: /api/v1/pods?fieldSelector=spec.nodeName%3Dip-172-18-66-203.ec2.internal (Caused by SSLError(FileNotFoundError(2, 'No such file or directory'),))
```

Client should not expect tempfiles to live indefinitely. It is extremely common for servers to reap tempfiles.

If you create a client you can easily see that it is using a temp file that cannot go away for the duration of the client:
```
>>> import kubernetes.config
>>> kubernetes.config.load_kube_config()
>>> api_client = kubernetes.client.ApiClient()
>>> api_client.rest_client.pool_manager.connection_pool_kw["ca_certs"]
'/tmp/tmpqkht2v2g'
```
You can reproduce the issue by deleting that temp file and attempting to make a request.

Code is here: https://github.com/kubernetes-client/python/blob/1271465acdb80bf174c50564a384fd6898635ea6/kubernetes/base/config/kube_config.py#L63-L78

In order for this to work reliably for long-running processes on standard linux systems, the temp file really needs to be created for each request rather than a single time at startup.

That said, on linux systems, a potential hack would be to use `/proc/self/fd/<fileno>` instead of the temfile path since that would share the lifecycle of the process.

**Environment**:
- Python version (`python --version`): 3.6
- Python client version (`pip list | grep kubernetes`): 21.7.0

tobewithyou1996:
#### What type of PR is this?
/kind bug

setting  load_kube_config()   temp_file_path ,  Avoid being deleted in the /tmp  directory

Fixes #1782 

```release-note
NONE
```

```docs

```

### DIFF
--- a/kubernetes/base/config/kube_config.py
+++ b/kubernetes/base/config/kube_config.py
@@ -789,7 +789,8 @@ def list_kube_config_contexts(config_file=None):
 
 def load_kube_config(config_file=None, context=None,
                      client_configuration=None,
-                     persist_config=True):
+                     persist_config=True,
+                     temp_file_path=None):
     """Loads authentication and cluster information from kube-config file
     and stores them in kubernetes.client.configuration.
 
@@ -800,6 +801,7 @@ def load_kube_config(config_file=None, context=None,
         set configs to.
     :param persist_config: If True, config file will be updated when changed
         (e.g GCP token refresh).
+    :param temp_file_path: store temp files path.
     """
 
     if config_file is None:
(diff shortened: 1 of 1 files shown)


roycaihw:
Please sign the CLA @tobewithyou1996

tobewithyou1996:
hi，My CLA has been signed. But I don't know why my status hasn't been updated, or it doesn't show up here. @roycaihw roycaihw
image

PR #1863, opened 2022-07-21
load proxy-url from config

### LINKED ISSUE #1862: utilize --proxy-url when added to the ~/.kube/config
Proxy is not set when the configuration contains the proxy.
https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-context-and-configuration
If you update your config by running.
kubectl config set-cluster my-cluster-name --proxy-url=my-proxy-url
kubectl obeys this behavior and this library does not.

This library should utilize a proxy when the configuration is updated as above.

Update your kubeconfig file.
kubectl config set-cluster my-cluster-name --proxy-url=https://foo.bar

running kubectl get pod will return an error. (if you have a need for a proxy you can also specify a proxy.)
This library ignores the setting.

The go library does seem to have behavior that utilizes proxy-url

jareddjenkins:
#### What type of PR is this?
/kind bug

Fixes #1862

Proxy is not set when the configuration contains the proxy-url.
https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-context-and-configuration
See the context and configuration for the setting that should be utilized.

behavioral change

when the kubeconfig for the cluster in the current context contains --proxy-url for the cluster. The client will set the proxy to that of proxy-url
see https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-context-and-configuration

### DIFF
--- a/kubernetes/client/configuration.py
+++ b/kubernetes/client/configuration.py
@@ -75,7 +75,7 @@ class Configuration(object):
     def __init__(self, host="http://localhost",
                  api_key=None, api_key_prefix=None,
                  username=None, password=None,
-                 discard_unknown_keys=False,
+                 proxy=None, discard_unknown_keys=False,
                  ):
         """Constructor
         """
@@ -153,7 +153,7 @@ class Configuration(object):
            cpu_count * 5 is used as default value to increase performance.
         """
 
-        self.proxy = None
+        self.proxy = proxy
         """Proxy URL
         """
         self.no_proxy = None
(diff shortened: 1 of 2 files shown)


epitstickpan:
@roycaihw, anything I can do to help this get approval?

jlonganecker:
This feature would be very helpful! What is the status on this?

jlonganecker:
How would I go about reviewing this?

dorukozturk:
Any progress with this issue?

kubernetes/client/configuration.py
```diff
@@ -75,7 +75,7 @@ class Configuration(object):
     def __init__(self, host="http://localhost",
                  api_key=None, api_key_prefix=None,
                  username=None, password=None,
-                 discard_unknown_keys=False,
+                 proxy=None, discard_unknown_keys=False,
```
fabianvf, line 78: Sorry for the delay. Unfortunately I think this file is generated, any changes here would be overwritten on a version bump. In order to make permanent changes to this file the PR would need to go to the generator repo, then we'd need to generate from the new version.

dorukozturk: oh I see, that makes sense. Thank you for the information

PR #1865, opened 2022-07-25
fix: account for async_req=True in create_from_yaml_single_item

### LINKED ISSUE #1864: create_from_yaml not raising an exception on failure
[create_from_yaml ](https://github.com/kubernetes-client/python/blob/master/kubernetes/utils/create_from_yaml.py) is not raising a "FailToCreateError" exception when the contents of the yaml fail to apply.

The result from create_from_yaml clearly shows a failure, for the right reasons (my example is attempting to create a job in a non-existent namespace. But the failure is not caught by the "FailToCreateError" exception as suggested by the documentation.

image

I expect a "FailToCreateError" exception to be raised to allow handling of errors applying the contents yaml files.

Using local Kind cluster the following script and yaml file will report "Test job created" when in fact the job is never created.

```
from loguru import logger
from kubernetes.utils import create_from_yaml, FailToCreateError

def run_job(k8s_client, job_definition_file):
    try:
        status = create_from_yaml(k8s_client, job_definition_file, verbose=False, async_req=True)
    except FailToCreateError:
        logger.error(f"This should fail but never does")
        return False
    except ValueError as e:
        logger.error(str(e))
        return False
    except FileNotFoundError as e:
        logger.error(str(e))
        return False
    else:
        logger.info(f"Test job created")
        return status

config.load_kube_config()
k8s_client = client.ApiClient()

job_file = "path/to/job.yaml"
run_job(k8s_client, job_file)
```
Example job yaml file contents:

```
---
apiVersion: batch/v1
kind: Job
metadata:
  namespace: non-existant-namespace
  name: test-job
spec:
  template:
    metadata:
      labels:
        app: test-job
    spec:
      containers:
        - name: test-job
          image: some-image:latest
          imagePullPolicy: IfNotPresent
```

Believe it to be caused by [create_from_yaml_single_item](https://github.com/kubernetes-client/python/blob/0553bba32dd5f2029a086131c893f0b295ac4297/kubernetes/utils/create_from_yaml.py#L229) not raising an exception up to the higher functions.

**Environment**:
- Kubernetes version: 1.24.2
- OS: Fedora 36
- Python version: 3.10
- Python client version: 24.2.0

Edit: missing words.

antonblr:
#### What type of PR is this?

/kind bug

In cases when `utils.create_from_yaml_single_item()` is called with `async_req = True` we need to retrieve the actual result of the call from the queue.

Fixes #1864

```release-note

```

```docs

```

### DIFF
--- a/kubernetes/utils/create_from_yaml.py
+++ b/kubernetes/utils/create_from_yaml.py
@@ -257,6 +257,10 @@ def create_from_yaml_single_item(
         kwargs.pop('namespace', None)
         resp = getattr(k8s_api, "create_{0}".format(kind))(
             body=yml_object, **kwargs)
+
+    if kwargs.get('async_req'):
+        resp = resp.get()
+
     if verbose:
         msg = "{0} created.".format(kind)
         if hasattr(resp, 'status'):
--- a/kubernetes/e2e_test/test_utils.py
+++ b/kubernetes/e2e_test/test_utils.py
@@ -460,3 +460,24 @@ class TestUtils(unittest.TestCase):
             name="mock-pod-1", namespace=self.test_namespace, body={})
         app_api.delete_namespaced_deployment(
             name="mock", namespace=self.test_namespace, body={})
+
+    # Error test case scenarios
+
+    # Issue #1864.
+    def test_error_creating_resource_in_non_existent_namespace(self):
+        """
+        Should return a valid error when trying to create a resource
+        in non-existent namespace.
+        """
+        resource = self.path_prefix + "job-non-existent-namespace.yaml"
+        k8s_client = client.ApiClient(configuration=self.config)
+        with self.assertRaises(utils.FailToCreateError):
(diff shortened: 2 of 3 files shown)


kubernetes/utils/create_from_yaml.py
```diff
@@ -257,6 +257,10 @@ def create_from_yaml_single_item(
         kwargs.pop('namespace', None)
         resp = getattr(k8s_api, "create_{0}".format(kind))(
             body=yml_object, **kwargs)
+
+    if kwargs.get('async_req'):
```
roycaihw, line 261: If `async_req=True` is specified, isn't it the callers responsibility to examine the result? 
Here we are turning an async request to sync, is it what the caller would expect? 
cc @ChrisOwenBoring

antonblr: It will require more changes then, I believe (i.e. handling `if verbose:` down below, callers return type and mention that they won't raise `FailToCreateError` for `async_req`)

Issue #1846, opened 2022-06-28
sanitize_for_serialization returns empty Dict when all values are None


Uranium2:
**What happened**:
I'm trying to monitor my Jobs with `read_namespaced_job_status` and return the `status` in Dict/Json format. When I create the Job and get it's status just after (under 5/10 seconds before real creation on Cluster) I get a status of Dict where all values are `None`, which is fine. But when I try to `sanitize_for_serialization` it, I get an empty Dict.

```python
> from kubernetes import client
> from kubernetes.client import ApiClient

> batch_v1 = client.BatchV1Api()
> api_response = batch_v1.read_namespaced_job_status(name="job-1", namespace="my-namespace")
> print(api_response)
{'active': None,
 'completed_indexes': None,
 'completion_time': None,
 'conditions': None,
 'failed': None,
 'ready': None,
 'start_time': None,
 'succeeded': None,
 'uncounted_terminated_pods': None}
> api = ApiClient()
> api_response_status_dict = api.sanitize_for_serialization(api_response.status)
> print(api_response_status_dict)
{}
```
**What you expected to happen**:
I expected `api_response_status_dict` to be a non empty Dict, with all keys and with values to `None`.

**Environment**:
- Kubernetes version: v1.23.5
- OS : Ubuntu 20.04.3 LTS (WSL1)
- Python version : Python 3.8.10
- Python client version: 23.3.0

Issue #1421, opened 2021-04-21
Bump openapi-generator version in a future pre-release


roycaihw:
ref https://github.com/kubernetes-client/python/issues/1353#issuecomment-824230648

There is some bug fixes in openapi-generator that we want to pick up. We should bump the generator version in an upcoming pre-release. 

Currently our latest releases are v17.14.0a1 (published) and v18.17.0a1 (cut but not published yet). I don't think we want to bump the version in v17.x.0b1 because that will cause more merge conflict, plus Kubernetes 1.17 is no longer supported. Our options are either v18.x.0b1 or v19.x.0a1.

/assign
cc @yliaog

roycaihw:
I'm working on adding some scripts to automate the release process, e.g. https://github.com/kubernetes-client/python/pull/1419. 

I want to use the scripts to automate the client generation and release note collection for 18.0.0-snapshot in master (today it's 17.0.0-snapshot), and then we can upgrade the client generator.

roycaihw:
/reopen

to track upgrading the generator version in this client.

roycaihw:
/remove-lifecycle stale
/unassign
/help

This is needed but I don't have the bandwidth to get to it at the moment https://github.com/kubernetes-client/python/pull/1743

Issue #1032, opened 2019-12-13
exec command is parsing output if output looks like json


tkinz27:
**What happened (please include outputs or screenshots)**:

Working on creating a scale test where the test orchestrator will exec in each pod and run a curl command to verify the pod is in the expected state. The output of the curl is JSON.

The problem seems to be that the stream client will parse the json as a python dictionary if the output is only json, but return it to the caller as a stringified dictionary. Expecting json, `json.loads` is failing.

```
[11]: cmd = ["curl", "http://localhost:8081/v1/devices/self"]
In [12]:     resp = kstream.stream(api_instance.connect_get_namespaced_pod_exec,
    ...:                   name=pod,
    ...:                   namespace=ns,
    ...:                   container=container,
    ...:                   command=cmd,
    ...:                   stderr=True,
    ...:                   stdin=False,
    ...:                   tty=False,
    ...:                   stdout=True)
    ...:
In [13]:
In [13]:
In [13]:
In [13]: resp
Out[13]: '  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r100  1838  100  1838    0     0   897k  {"id":"01DW0BBVBF4G4G5BND6HB5AMBX","config":"{}","deviceType":"LBCM","locale":null,"serialNumber":"ce057580-1ddd-11ea-92ca-067144143630","tags":null}    0 --:--:-- --:--:-- --:--:--  897k\n'
[11:27 AM]
In [16]: cmd = ["curl", "-s", "http://localhost:8081/v1/devices/self"]
In [17]:     resp = kstream.stream(api_instance.connect_get_namespaced_pod_exec,
    ...:                   name=pod,
    ...:                   namespace=ns,
    ...:                   container=container,
    ...:                   command=cmd,
    ...:                   stderr=True,
    ...:                   stdin=False,
    ...:                   tty=False,
    ...:                   stdout=True)
    ...:
In [18]: resp
Out[18]: "{'id': '01DW0BBVBF4G4G5BND6HB5AMBX', 'config': '{}', 'deviceType': 'LBCM', 'locale': None,  'serialNumber': 'ce057580-1ddd-11ea-92ca-067144143630', 'tags': None}"
```

**What you expected to happen**:

I really did not expect the response to be different depending on the output of the command being run.

[...]

PR #1887, opened 2022-09-12
Upgrades the openapi python client to v6.2.1

spacether:
#### What type of PR is this?

/kind feature
/kind api-change

This PR switches to using the python generator from openapi-generator v6.2.1
Please do not review this PR into I move it out of WIP/Draft State
This PR updates the python client which now has:
- type hints everywhere
- more robust composition(oneOf/anyOf/allOf/not) support
- payload someProp case used rather than python some_prop used to ensure compliance when imposing json schema keywords like properties/additionalProperties/composition. This allows a payload to contain the separate keys someProp and some_prop
- supports newer openapi features like handling of multiple request body content types and response body content types

I am seeking feedback from your project on what you think about the python client in the latest openapi-generator and if you prefer to use it going forward.

Fixes #

- https://github.com/kubernetes-client/python/issues/1846
- https://github.com/kubernetes-client/python/issues/1421
- https://github.com/kubernetes-client/python/issues/1032

I am seeking feedback from your project on how you feel about using the code generated from the 6.2.1 python generator.

```
Action required. This is a breaking change as the new python client is structured differently.
Python min version changed from 3.6 to 3.7

### Changes
1. This generator uses spec case for all (object) property names and parameter names.
    - So if the spec has a property name like camelCase, it will use camelCase rather than camel_case
    - So you will need to update how you input and read properties to use spec case
2. Endpoint parameters are stored in dictionaries to prevent collisions (explanation below)
    - So you will need to update all of your endpoint calls that have query/path/header/cookie params 
      to pass that parameter data in query_params, header_params, path_params, cookie_params 
      named argument inputs with values of type dict
3. Endpoint responses now include the original response, the deserialized response body, and (todo)the deserialized headers
    - So you will need to update your code to use response.body to access deserialized data
4. All validated data is instantiated in an instance that subclasses all validated Schema classes and Decimal/str/list/tuple/frozendict/NoneClass/BoolClass/bytes/io.FileIO
    - This means that you can use isinstance to check if a payload validated against a schema class
    - This means that no data will be of type None/True/False
        - ingested None will subclass NoneClass
        - ingested True will subclass BoolClass
        - ingested False will subclass BoolClass
        - So if you need to check is True/False/None, instead use instance.is_true_oapg()/.is_false_oapg()/.is_none_oapg()
5. All validated class instances are immutable except for ones based on io.File
    - This is because if properties were changed after validation, that validation would no longer apply
    - So no changing values or property values after a class has been instantiated
6. String + Number types with formats
    - String type data is stored as a string and if you need to access types based on its format like date,
    date-time, uuid, number etc then you will need to use accessor functions on the instance
    - type string + format: See .as_date_oapg, .as_datetime_oapg, .as_decimal_oapg, .as_uuid_oapg
    - type number + format: See .as_float_oapg, .as_int_oapg
    - this was done because openapi/json-schema defines constraints. string data may be type string with no format
    keyword in one schema, and include a format constraint in another schema
    - So if you need to access a string format based type, use as_date_oapg/as_datetime_oapg/as_decimal_oapg/as_uuid_oapg
    - So if you need to access a number format based type, use as_int_oapg/as_float_oapg
7. Property access on AnyType(type unset) or object(dict) schemas
    - Only required keys with valid python names are properties like .someProp and have type hints
    - All optional keys may not exist, so properties are not defined for them
    - One can access optional values with dict_instance['optionalProp'] and KeyError will be raised if it does not exist
    - Use get_item_oapg if you need a way to always get a value whether or not the key exists
        - If the key does not exist, schemas.unset is returned from calling dict_instance.get_item_oapg('optionalProp')
        - All required and optional keys have type hints for this method, and @typing.overload is used
        - A type hint is also generated for additionalProperties accessed using this method
    - So you will need to update you code to use some_instance['optionalProp'] to access optional property
    and additionalProperty values
8. The location of the api classes has changed
    - Api classes are located in your_package.apis.tags.some_api
    - This change was made to eliminate redundant code generation
    - Legacy generators generated the same endpoint twice if it had > 1 tag on it
    - This generator defines an endpoint in one class, then inherits that class to generate
      apis by tags and by paths
    - This change reduces code and allows quicker run time if you use the path apis
        - path apis are at your_package.apis.paths.some_path
    - Those apis will only load their needed models, which is less to load than all of the resources needed in a tag api
    - So you will need to update your import paths to the api classes

This is an excerpt from the included migration guide which shows what needs to be updated: https://github.com/spacether/kube_python/tree/feat_switches_to_python_exp_client/kubernetes#migration-from-other-generators-like-python-and-python-legacy
```

```docs

```


yliaog:
i think we should not be using the experimental generator. please file a PR to use the official generator release.

spacether:
@yliaog I just updated this PR to use the 6.2.0 python generator

spacether:
@yliaog can you or others review this PR? Do I need to do anything else?

yliaog:
could you please squash the commits? I'll take a look afterwards. Thanks.

spacether:
@yliaog just squashed them into one commit now

yliaog:
https://github.com/spacether/kube_python/tree/feat_switches_to_python_exp_client/kubernetes#migration-from-other-generators-like-python-and-python-legacy

so this PR introduces breaking change? for breaking change, we need to solicit community feedback.

also it's better to split the PR into at least two commits, one commit for generated files, the other commit is for the manual edited files.

spacether:
> https://github.com/spacether/kube_python/tree/feat_switches_to_python_exp_client/kubernetes#migration-from-other-generators-like-python-and-python-legacy
> 
> so this PR introduces breaking change? for breaking change, we need to solicit community feedback.
> 
> also it's better to split the PR into at least two commits, one commit for generated files, the other commit is for the manual edited files.

Yes the updated client has breaking changes versus the old ones. I have not manually edited files.
Which files need to be manually edited/where in this code base uses the client?
How do we solicit community feedback?

yliaog:
https://github.com/kubernetes-client/gen/tree/master/openapi needs manual update?

please send email to:
kubernetes-dev@googlegroups.com

file an issue similar to: https://github.com/kubernetes-client/python/issues/974

spacether:
My plan is to wait until 6.2.1 is released to get community feedback

yliaog:
sounds good. thanks.

spacether:
Just posted an email in `https://groups.google.com/a/kubernetes.io/g/dev` which looks like the newest google group to use

tomplus:
@spacether Could you update examples/ and e2e tests?

spacether:
@tomplus can you help me with my tox requirement installation?
I have added the needed dependencies to requirements.txt, but they are not being picked up when installing with tox. What am I missing?
Note: `pip install -e .` correctly installs the requirements

atti92:
Any update on this? openapi-generator v6.3.0 is released already.

spacether:
I need help getting tox working. Until someone helps this is stalled. Also openapi-generator has announced that the current python generator will be replaced with the python-nextgen generator in v 7.0.0

Issue #1911, opened 2022-10-07
load_kube_config fails when decode a not safe url token


uelei:
**What happened (please include outputs or screenshots)**:
load_kube_config fails when decode a not safe url token

```    
config.load_kube_config()
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/dev-helpers-ydB7Fkd--py3.10/lib/python3.10/site-packages/kubernetes/config/kube_config.py", line 820, in load_kube_config
    loader.load_and_set(config)
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/dev-helpers-ydB7Fkd--py3.10/lib/python3.10/site-packages/kubernetes/config/kube_config.py", line 590, in load_and_set
    self._load_authentication()
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/dev-helpers-ydB7Fkd--py3.10/lib/python3.10/site-packages/kubernetes/config/kube_config.py", line 288, in _load_authentication
    if self._load_auth_provider_token():
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/dev-helpers-ydB7Fkd--py3.10/lib/python3.10/site-packages/kubernetes/config/kube_config.py", line 307, in _load_auth_provider_token
    return self._load_oid_token(provider)
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/dev-helpers-ydB7Fkd--py3.10/lib/python3.10/site-packages/kubernetes/config/kube_config.py", line 399, in _load_oid_token
    base64.b64decode(parts[1] + padding).decode('utf-8')
  File "/usr/local/Cellar/python@3.10/3.10.6_2/Frameworks/Python.framework/Versions/3.10/lib/python3.10/base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
binascii.Error: Invalid base64-encoded string: number of data characters (1509) cannot be 1 more than a multiple of 4
```
**What you expected to happen**:
load correctly the token

**How to reproduce it (as minimally and precisely as possible)**:
change your ~/.kube/config id-token to 

```
[...]

PR #1912, opened 2022-10-10
decoding a not safe url in load_kube_config

ramvikrams:
for issue #1911 fixing load_kube_config for decoding unsafe url token


uelei (approved):
LGTM

PR #1933, opened 2022-10-21
Adds support for custom Server Name Indication (SNI)

### LINKED ISSUE #1889: Client doesn't honor tls-server-name setting in kubeconfig

Tried to connect to my cluster with a utility that uses your library. As far as I can tell, the utility's code is correct and the k8s-client python code does not honor the tls-server-name and propagate it down the stack to urllib3.

Received this exception:

```
  File "/home/nicholas/code/k8spurger-venv/lib/python3.9/site-packages/urllib3/connectionpool.
py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/home/nicholas/code/k8spurger-venv/lib/python3.9/site-packages/urllib3/connectionpool.
py", line 386, in _make_request
    self._validate_conn(conn)
  File "/home/nicholas/code/k8spurger-venv/lib/python3.9/site-packages/urllib3/connectionpool.
py", line 1042, in _validate_conn
    conn.connect()
  File "/home/nicholas/code/k8spurger-venv/lib/python3.9/site-packages/urllib3/connection.py",
 line 467, in connect
    _match_hostname(cert, self.assert_hostname or server_hostname)
  File "/home/nicholas/code/k8spurger-venv/lib/python3.9/site-packages/urllib3/connection.py",
 line 540, in _match_hostname
    match_hostname(cert, asserted_hostname)
  File "/home/nicholas/code/k8spurger-venv/lib/python3.9/site-packages/urllib3/util/ssl_match_
hostname.py", line 150, in match_hostname
    raise CertificateError(
urllib3.util.ssl_match_hostname.CertificateError: hostname 'CNAME-THAT-I-USE' doesn
't match either of 'LIST', 'OF', 'OTHER', 'NAMES'
```

This is explicitly why I set tls-server-name in my kubeconfig.

TLS connection to my cluster, with the proper SNI sent.

Set your cluster.server property to some IP/CNAME not in the cert. Set tls-server-name in kubeconfig correctly.

Looks like it is pretty easy to support. Just need to change https://github.com/kubernetes-client/python/blob/master/kubernetes/base/config/kube_config.py#L544 to read the tls-server-name, and make sure that gets propagated to https://github.com/kubernetes-client/python/blob/master/kubernetes/client/rest.py#L73 as `assert_hostname`

tigrato:
#### What type of PR is this?
/kind bug

This PR adds support for custom SNI values. If `tls-server-name` is set in kubeconfig, the client did not use it resulting in SSL/TLS errors if the property is required to connect to the server.

Fixes #1889

```release-note
Fixed `tls-server-name` (SNI)  property usage when explicitly set in kubeconfig.
```

### DIFF
--- a/kubernetes/client/rest.py
+++ b/kubernetes/client/rest.py
@@ -77,6 +77,9 @@ class RESTClientObject(object):
         if configuration.retries is not None:
             addition_pool_args['retries'] = configuration.retries
 
+        if configuration.tls_server_name:
+            addition_pool_args['server_hostname'] = configuration.tls_server_name
+
         if maxsize is None:
             if configuration.connection_pool_maxsize is not None:
                 maxsize = configuration.connection_pool_maxsize
--- a/kubernetes/base/config/kube_config.py
+++ b/kubernetes/base/config/kube_config.py
@@ -567,6 +567,8 @@ class KubeConfigLoader(object):
                         temp_file_path=self._temp_file_path).as_file()
         if 'insecure-skip-tls-verify' in self._cluster:
             self.verify_ssl = not self._cluster['insecure-skip-tls-verify']
+        if 'tls-server-name' in self._cluster:
+            self.tls_server_name = self._cluster['tls-server-name']
 
     def _set_config(self, client_configuration):
         if 'token' in self.__dict__:
@@ -578,7 +580,7 @@ class KubeConfigLoader(object):
                 self._set_config(client_configuration)
             client_configuration.refresh_api_key_hook = _refresh_api_key
         # copy these keys directly from self to configuration object
-        keys = ['host', 'ssl_ca_cert', 'cert_file', 'key_file', 'verify_ssl']
+        keys = ['host', 'ssl_ca_cert', 'cert_file', 'key_file', 'verify_ssl','tls_server_name']
         for key in keys:
             if key in self.__dict__:
                 setattr(client_configuration, key, getattr(self, key))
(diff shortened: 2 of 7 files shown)


tigrato:
@yliaog @roycaihw 

Is there any possibility to include this PR in the next release?

kubernetes/client/rest.py
```diff
@@ -77,6 +77,9 @@ def __init__(self, configuration, pools_size=4, maxsize=None):
         if configuration.retries is not None:
             addition_pool_args['retries'] = configuration.retries
 
+        if configuration.tls_server_name:
```
fabianvf, line 80: Unfortunately I think this file is generated, in order to make permanent changes here I believe the PR would need to go to the generator repo

goodspark: Can you provide some more information on what to do? Because the gen repo doesn't really say nor point to what needs to be updated. The contributing guide doesn't go into detail either.

This file seems to ref https://github.com/kubernetes-client/gen/blob/master/proto/dependencies.sh even more generated files from the apimachinery repo.

programmerq: I did a bit of digging, and the rest.py file appears to come from the openapi-generator upstream code:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/python/rest.handlebars

Would it be possible to implement a custom template per the upstream openapi-generator docs? https://openapi-generator.tech/docs/templating

tigrato: @fabianvf @yliaog 

The auto-generated code is available for `openapi-generator` version `>=6.6.0` ([PR](https://github.com/OpenAPITools/openapi-generator/pull/15283)).

Since this repo uses version `v4.3.0`, I decided to apply a patch - similar to the current one that exists - to apply the necessary changes to support SNI. Once we start using a version that includes the upstream fix, the patch can be safely removed.

Can you please check it again?

goodspark (approved):
LGTM

kubernetes/client/rest.py
yliaog: thanks for the PR. it looks good to me, @fabianvf do you have any question or concern?

fabianvf (approved):
/lgtm

PR #1935, opened 2022-10-24
Fixed double module imports

Ark-kun:
In Python, when you write `import foo.bar.baz` this means that the modules would be imported and the name `foo` will be bound locally and becomes available in the module. https://docs.python.org/3/reference/simple_stmts.html#import

So, doing `import kubernetes.client` leads to name `kubernetes` (not `client`) being added to the `kubernetes` module leading to a weird infinite deep nesting. See:
```
>>> import kubernetes
>>> kubernetes
<module 'kubernetes' from 'C:\\Users\\Ark\\AppData\\Local\\Programs\\Python\\Python310\\lib\\site-packages\\kubernetes\\__init__.py'>
>>> kubernetes.kubernetes.kubernetes
<module 'kubernetes' from 'C:\\Users\\Ark\\AppData\\Local\\Programs\\Python\\Python310\\lib\\site-packages\\kubernetes\\__init__.py'>
```

Another issue is that while `import kubernetes.client` in `kubernetes.__init__.py` technically makes `kubernetes.client` available in user code that does `import kubernetes`, this a availability is a somewhat brittle side-effect of importing. Some linters and code tools do not see this effect and do not understand that the `kubernetes.client` attribute would be set at runtime. Using explicit import (`from kubernetes import client`) in `kubernetes.__init__.py` solves this issue.

We can solve this issues by using the `import ... from ...` syntax: Replace `import kubernetes.client` with `from kubernetes import client`.

I see that most modules already use relative imports, so I'm using relative imports here as well: `from . import client`.

#### What type of PR is this?

/kind cleanup

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/kubernetes/__init__.py
+++ b/kubernetes/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2016 The Kubernetes Authors.
+# Copyright 2022 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


yliaog:
[do-not-merge/work-in-progress](https://github.com/kubernetes-client/python/labels/do-not-merge%2Fwork-in-progress)

i see this label in this PR. do you need more work on it?

Ark-kun:
> i see this label in this PR. do you need more work on it?

No. Sorry. Just wanted to see all checks pass.

Ark-kun:
Looks like all the tests pass.

PR #1949, opened 2022-11-10
Improve the refreshing of tokens from inside the cluster

### LINKED ISSUE #1946: Requests from outside the cluster may have stale tokens and fail with status code `401`
### What happened

A request to Kubernetes **from outside the cluster** failed with status code `401`, even though the configuration had been loaded and previous requests had been successful.

The traceback is

```python3
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/kubernetes/client/api/core_v1_api.py", line 14721, in list_namespace
    return self.list_namespace_with_http_info(**kwargs)  # noqa: E501
  File "/usr/lib/python3/dist-packages/kubernetes/client/api/core_v1_api.py", line 14842, in list_namespace_with_http_info
  File "/usr/lib/python3/dist-packages/kubernetes/client/api_client.py", line 353, in call_api
    _preload_content, _request_timeout, _host)
  File "/usr/lib/python3/dist-packages/kubernetes/client/api_client.py", line 184, in __call_api
    _request_timeout=_request_timeout)
  File "/usr/lib/python3/dist-packages/kubernetes/client/api_client.py", line 377, in request
    headers=headers)
  File "/usr/lib/python3/dist-packages/kubernetes/client/rest.py", line 244, in GET
    query_params=query_params)
  File "/usr/lib/python3/dist-packages/kubernetes/client/rest.py", line 234, in request
    raise ApiException(http_resp=r)
kubernetes.client.exceptions.ApiException: (401)
Reason: Unauthorized
HTTP response headers: HTTPHeaderDict({'Audit-Id': '7dd1383a-497d-4d3c-85cc-26a13dc9406f', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Date': 'Tue, 08 Nov 2022 10:07:04 GMT', 'Content-Length': '129'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}
```

### What you expected to happen

Given the `410 Unauthorized` response, we can safely assume that the problem lies with tokens and more specifically the way tokens are refreshed.

The client has had an issue with tokens, namely issue #741, but it has been closed and (should have been) fixed by the kubernetes-client/python-base/pull/250 PR. Therefore no requests should be made with stale tokens and fail with `410 Unauthorized`. 

### How to reproduce it

A minimal example is

```python3
import time

from kubernetes import config
from kubernetes.client import CoreV1Api

config.load_config()

api_1 = CoreV1Api()
api_2 = CoreV1Api()

_ = api_1.list_namespace()  # succeeds
_ = api_2.list_namespace()  # succeeds

time.sleep(15 * 60)

_ = api_1.list_namespace()  # succeeds
_ = api_2.list_namespace()  # fails
```

That is, we

1. load the configuration,
2. create two clients,
3. make one request with each client,
4. sleep for 15 minutes, and
5. make one more request with each client.

The tokens we receive from EKS have a TTL of 14 minutes (see also https://github.com/aws/aws-cli/blob/2.8.9/awscli/customizations/eks/get_token.py#L62). Therefore we sleep for 15 minutes waiting for the token to expire.

_Before_ waiting both requests will succeed, but _after_ waiting the first request will succeed and the second one will fail.

Note that what matters is which client makes the request first _after_ the wait (not which client was created or made a request first _before_ the wait). For example, if `api_2` makes the request first after the wait, then it will succeed and `api_1` will fail.

Also note that we receive `410 Unauthorized` responses only when we make requests

1. with _multiple clients_. If we create a single client, all the requests we make with it will be successful. This suggests that different clients have different tokens they make requests with, and some of these tokens are stale.
2. from _outside the cluster_. Requests from inside the cluster succeed. This means the problem lies in `kube_config`.

### Anything else we need to know?

A closer look at the example shows that (under the hood) we

1. a. instantiate a loader (`KubeConfigLoader()`),
    b. get a token and update the token (`token`) and expiration timestamp (`expiry`) of the loader, and
    c. update the token (`api_key['authorization']`) of the default configuration (`Configuration._default`),
2. a. instantiate two clients (`CoreV1Api()`), and
    b. set copies of the default configuration as their configuration,
3. make a request with each client,
4. wait for the token to expire, and
5. a. get a fresh token and update the token and expiration timestamp of the loader,
    b. update the token of the first client,
    c. make a request with the first client (with the updated token), and
    d. make a request with the second client (with the expired token).
 
Notice that we do **not** update the token of the second client.

When we make a request with a client, **if the token of the loader is expired or expiring in 5 minutes or less**, we
1. get a fresh token and update the token and expiration timestamp of the loader, and
2. update the token of the client.

In the example, _after_ the token expires,
1. the first client makes a request which triggers the refresh hook. That is, the loader both refreshes its token **and** updates the token of the client.
2. the second client makes a request which does **not** trigger the refresh hook. That is, the loader neither refreshes its token (as it should not) nor updates the token of the client (as it should).

To fix this issue, when a client makes a request, the loader should
1. refresh its token, only if the token is expired or expiring, **but**
2. update the token of the client **every time**.

The diff is

```diff
diff --git a/kubernetes/base/config/kube_config.py b/kubernetes/base/config/kube_config.py
index b95955448..f72890982 100644
--- a/kubernetes/base/config/kube_config.py
+++ b/kubernetes/base/config/kube_config.py
@@ -575,7 +575,7 @@ class KubeConfigLoader(object):
             def _refresh_api_key(client_configuration):
                 if ('expiry' in self.__dict__ and _is_expired(self.expiry)):
                     self._load_authentication()
-                    self._set_config(client_configuration)
+                self._set_config(client_configuration)
             client_configuration.refresh_api_key_hook = _refresh_api_key
         # copy these keys directly from self to configuration object
         keys = ['host', 'ssl_ca_cert', 'cert_file', 'key_file', 'verify_ssl']
```

### Environment

- Kubernetes version: v1.25.3
- OS: Ubuntu 22.04.1 LTS
- Python version: Python 3.10.6
- Python client version: 25.3.0

### LINKED ISSUE #1948: Requests from inside the cluster misuse the API to refresh tokens
## Problem statement

The OpenAPITools/openapi-generator/pull/3594 PR added the `refresh_api_key_hook()` hook to `Configuration()` objects to refresh expired (or expiring) tokens. This hook is undefined by default. If defined, the `get_api_key_with_prefix()` method will call it to refresh the token _before_ returning it.

### Out-of-cluster

The kubernetes-client/python-base/pull/250 PR defined the `__refresh_api_key()` function and used it to override the `refresh_api_key_hook()` hook to refresh tokens from outside the cluster.

This is the proper usage of the API.

### In-cluster

The kubernetes-client/python-base/pull/191 and kubernetes-client/python-base/pull/193 PRs defined the `load_token_from_file()` function and used it to override the `get_api_key_with_prefix()` method to refresh tokens from inside the cluster.

That is, a well-defined higher-level function is overridden, instead of the undefined lower-level hook, which is designed to be overridden. Put simply, **the API is misused**.

A side-effect of this is that **the token of the client is never updated**. I.e., the client always has the (possibly stale) token that was loaded by the last call to `load_config()` at the time it was created. This works just because the token of the client is _not_ used for the requests to Kubernetes (even though it should be).

Note that by 'client' we refer to an `XXXApi` object (e.g. `CoreV1Api`), which we use to make requests to Kubernetes.

## Proposed fix

To use the API properly, update the tokens of clients making requests to Kubernetes, and use these tokens for requests, we can mirror the way tokens are refreshed from outside the cluster, i.e.,

* define a `_refresh_api_key` function (instead of `load_token_from_file`), which
* updates the token of the client (instead of simply returning a token), and
* override the `refresh_api_key_hook` hook (instead of `get_api_key_with_prefix`).

The diff is

```diff
diff --git a/kubernetes/base/config/incluster_config.py b/kubernetes/base/config/incluster_config.py
index 5dabd4b7c..86070df43 100644
--- a/kubernetes/base/config/incluster_config.py
+++ b/kubernetes/base/config/incluster_config.py
@@ -92,12 +92,12 @@ class InClusterConfigLoader(object):
         if not self._try_refresh_token:
             return
 
-        def load_token_from_file(*args):
+        def _refresh_api_key(client_configuration):
             if self.token_expires_at <= datetime.datetime.now():
                 self._read_token_file()
-            return self.token
+            self._set_config(client_configuration)
 
-        client_configuration.get_api_key_with_prefix = load_token_from_file
+        client_configuration.refresh_api_key_hook = _refresh_api_key
 
     def _read_token_file(self):
         with open(self._token_filename) as f:
```

AlexisZam:
#### What type of PR is this?

kind/design

Requests from inside the cluster misuse the API to refresh tokens.

Fixes #1948

```release-note
NONE
```

### DIFF
--- a/kubernetes/base/config/incluster_config.py
+++ b/kubernetes/base/config/incluster_config.py
@@ -92,12 +92,12 @@ class InClusterConfigLoader(object):
         if not self._try_refresh_token:
             return
 
-        def load_token_from_file(*args):
+        def _refresh_api_key(client_configuration):
             if self.token_expires_at <= datetime.datetime.now():
                 self._read_token_file()
-            return self.token
+            self._set_config(client_configuration)
 
-        client_configuration.get_api_key_with_prefix = load_token_from_file
+        client_configuration.refresh_api_key_hook = _refresh_api_key
 
     def _read_token_file(self):
         with open(self._token_filename) as f:


yliaog:
thanks for the pr. is there a test to show the pr fixes the problem stated in the issue: https://github.com/kubernetes-client/python/issues/1948

AlexisZam:
@yliaog This PR fixes a **design problem**, rather than a bug. As explained in #1948, the code works, but _the API is misused_. The proposed solution properly uses the API, and bring the in-cluster code in line with the out-of-cluster code.

Compare the _in-cluster_ code:

https://github.com/kubernetes-client/python/blob/392a8c1d0767ce534b121b3b0553e5b1297e430e/kubernetes/base/config/incluster_config.py#L95-L100

and the _out-of-cluster_ code:

https://github.com/kubernetes-client/python/blob/8f80cf444ed9750651a0745a360ddef845b1ede9/kubernetes/base/config/kube_config.py#L575-L579

yliaog:
thanks.

/lgtm
/approve

PR #1966, opened 2022-12-12
Fix load_kube_config unexpected `kube_config_path` keyword argument

### LINKED ISSUE #1893: load_kube_config() got an unexpected key word argument 'kube_config_path'
When using load_config() and "kube_config_path" in kwargs:
https://github.com/kubernetes-client/python/blob/16935036252be0fdf164f8ce624bd71eb8d35e89/kubernetes/base/config/__init__.py#L24
load_config(kube_config_path="/")

Received this error:
load_kube_config() got an unexpected key word argument 'kube_config_path'.

because:
image

image

blackstorm:
#### What type of PR is this?
/kind bug

Rename `kube_config_path` to the correct `config_file`

Fixes #1893 

Compatible `kube_config_path` argument, no break change.

```release-note
NONE
```

```docs

```

### DIFF
--- a/kubernetes/base/config/__init__.py
+++ b/kubernetes/base/config/__init__.py
@@ -34,6 +34,7 @@ def load_config(**kwargs):
     load_incluster_config functions.
     """
     if "kube_config_path" in kwargs.keys() or exists(expanduser(KUBE_CONFIG_DEFAULT_LOCATION)):
+        kwargs["config_file"] = kwargs.pop("kube_config_path", None)
         load_kube_config(**kwargs)
     else:
         print(
--- a/kubernetes/base/config/__init__.py
+++ b/kubernetes/base/config/__init__.py
@@ -33,7 +33,9 @@ def load_config(**kwargs):
     can be passed to either load_kube_config or
     load_incluster_config functions.
     """
-    if "kube_config_path" in kwargs.keys():
+    if "config_file" in kwargs.keys():
+        load_kube_config(**kwargs)
+    elif "kube_config_path" in kwargs.keys():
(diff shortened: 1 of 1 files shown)


kubernetes/base/config/__init__.py
```diff
@@ -33,7 +33,10 @@ def load_config(**kwargs):
     can be passed to either load_kube_config or
     load_incluster_config functions.
     """
-    if "kube_config_path" in kwargs.keys() or exists(expanduser(KUBE_CONFIG_DEFAULT_LOCATION)):
+    if "kube_config_path" in kwargs.keys():
```
roycaihw, line 36: it seems `kube_config_path` never worked. I wonder if we have any user setting it

kubernetes/base/config/__init__.py
```diff
@@ -33,7 +33,10 @@ def load_config(**kwargs):
...
+        kwargs["config_file"] = kwargs.pop("kube_config_path", None)
```
roycaihw, line 37: if both `kube_config_path` and `config_file` are set, we should keep the value of `config_file`

kubernetes/base/config/__init__.py
blackstorm: I think it might be due to the `load_kube_config` API change, `kube_config_path` is a ghost value now.

kubernetes/base/config/__init__.py
blackstorm: I added a check for the `config_file` value to ensure that it is used first

roycaihw:
/lgtm
/approve
Thanks!

PR #1973, opened 2022-12-19
redundant alias exports

### LINKED ISSUE #1970: Add __all__ or redundant alias to top level kubernetes

Currently pyright/pylance (popular Python language server/type-checker) give an error for e.g. this:

```
import kubernetes
kubernetes.config  # error!

# ... "config" is not a known member of module (reportGeneralTypeIssues)
```

The reason for this is the lack of `__all__` and/or rendundant aliasing in `kubernetes/__init__.py`-- effectively `kubernetes` has no public interface recognized by pyright/pylance.

This is easily fixable by defining `__all__` or with redundant aliasing:

```python
import kubernetes.client as client
import kubernetes.config as config
import kubernetes.dynamic as dynamic
import kubernetes.watch as watch
import kubernetes.stream as stream
import kubernetes.utils as utils
import kubernetes.leaderelection as leaderelection
```

smackesey:
#### What type of PR is this?

/kind bug

"Bug" is not really accurate but I had to pick a kind and nothing else fit. Rather, this PR marks symbols intended to be re-exported in `__init__.py` files as public by redundantly aliasing them (an alternative to defining `__all__`). This prevents false positive lint errors from type checkers like pyright.

Fixes #1970 

```release-note
This package now properly marks exported symbols as public, eliminating "private access" errors from some static analysis tools (pyright/pylance).
```

### DIFF
--- a/kubernetes/__init__.py
+++ b/kubernetes/__init__.py
@@ -16,10 +16,11 @@ __project__ = 'kubernetes'
 # The version is auto-updated. Please do not edit.
 __version__ = "25.0.0-snapshot"
 
-import kubernetes.client
-import kubernetes.config
-import kubernetes.dynamic
-import kubernetes.watch
-import kubernetes.stream
-import kubernetes.utils
-import kubernetes.leaderelection
+# Redundant aliasing marks submodules as "exported" symbols for static analyzers.
+import kubernetes.client as client
+import kubernetes.config as config
+import kubernetes.dynamic as dynamic
+import kubernetes.watch as watch
+import kubernetes.stream as stream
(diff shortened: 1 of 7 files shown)


kubernetes/__init__.py
```diff
@@ -16,10 +16,11 @@
 # The version is auto-updated. Please do not edit.
```
roycaihw, line 16: this file is auto generated. Please fix it in https://github.com/OpenAPITools/openapi-generator

rafrafek: I made a simple report here https://github.com/OpenAPITools/openapi-generator/issues/17152

Issue #1870, opened 2022-08-03
Error 400 Bad handshake when trying to execute command in a pod. 


ctheisingfl:
I have a simple script that is just trying to print out hello world in the pod's shell. It is unable to make a successful connection. 
```from click import command
from kubernetes import client, config
from kubernetes.stream import stream
import os
from kubernetes.client import configuration 


os.system("kubectl config use-context eap-dev")

def pod_exec(name, namespace, command, api_instance):
    contexts, active_context = config.list_kube_config_contexts()
    if not contexts:
        print("Cannot find any context in kube-config file.")
        return
    exec_command = ["/bin/sh", "-c", command]

    resp = stream(api_instance.connect_get_namespaced_pod_exec,
                  name,
                  namespace,
                  command=exec_command,
                  stderr=True, stdin=False,
                  stdout=True, tty=False,
                  _preload_content=True)

    while resp.is_open():
        resp.update(timeout=1)
        if resp.peek_stdout():
            print(f"STDOUT: \n{resp.read_stdout()}")
        if resp.peek_stderr():
            print(f"STDERR: \n{resp.read_stderr()}")
    resp.close()
    if resp.returncode != 0:
        raise Exception("Script failed")

command = """echo hello world"""
print (command)
pod = "foobarpodname"
config.load_kube_config()
v1 = client.CoreV1Api()
client1 = client.CoreV1Api(
        api_client=config.new_client_from_config(context="foobarcontext"))
pod_exec(pod, "foobarnamespace", command, client1)
```

I keep getting 
```kubernetes.client.exceptions.ApiException: (0)
Reason: Handshake status 400 Bad Request

and 

raise WebSocketBadStatusException("Handshake status %d %s", status, status_message, resp_headers)
websocket._exceptions.WebSocketBadStatusException: Handshake status 400 Bad Request```

I am not sure where the issue lies.

ctheisingfl:
I have fixed the issue.

michzimny:
@ctheisingfl, please share how have you fixed it.

ctheisingfl:
In 
resp = stream(api_instance.connect_get_namespaced_pod_exec,
                  name,
                  namespace,
                  command=exec_command,
                  stderr=True, stdin=False,
                  stdout=True, tty=False,
                  container = <container_name>,
                  _preload_content=True)

michzimny:
Thank you!

roycaihw:
Thank you @ctheisingfl for sharing!
/close

PR #1980, opened 2023-01-01
doc: comment pod_exec.py for multiple containers

ghinks:
The behavior in kubectl for an exec is to run the exec command against the default container if there are multiple
containers running in a pod (ie sidecars). However, the container needs to be specified with this python client call or
an error of kind **WebSocketBadStatusException** will be raised. Add a comment in the example to explain this.

#### What type of PR is this?

/kind documentation

**Document Side Car Container exec** 

Firstly a big thank you to the maintainers of the python kubernetes-client I hope to offer help going forwards but am 
revisiting python for the first time in a decade.

**Problem Description**

During the normal use of the [pod_exec.py](https://github.com/kubernetes-client/python/blob/master/examples/pod_exec.py) example it came to my attention that when executing against a pod with multiple containers (ie sidecars) a WebSocketBadStatusException was raised. To fix this a container , often annotated as *kubectl.kubernetes.io/default-container* would need to be specified in the call. (when using kubectl the behavior when in an exec is to target the default container in the pod)

```python
    resp = stream(api.connect_get_namespaced_pod_exec,
                  name=pod,
                  namespace='default',
                  command=exec_command,
                  container=container,
                  stdout=True, tty=False)
```
I found this out via the issue [1870](https://github.com/kubernetes-client/python/issues/1870). I want to help others so that they can find this fix without the pain I had. I am suggesting a comment in the existing example. An alternative would be an example of pod_exec_for_side_cars.py which would give an example of how to use this with a pod that has sidecars.

I would like to investigate the possibility of raising a better exception or targeting the default container in an exec call.

Reproduction Notes

I created a simple flask application and created an image from that tiny application. I Deployed this within a minikube 
cluster and followed the pod_exec.py example. I then installed [istio](https://istio.io/latest/docs/setup/getting-started/) 
onto my cluster and repeated the process. This time the pod_exec.py raised the exception.

- [this issue was tested with a simple flask application](https://github.com/ghinks/python_container_for_cloud)
- [flask application docker hub location](https://hub.docker.com/repository/docker/ghinks/python-for-cloud)
- [minikube service deployment to re-produce exception when no container is specified](https://github.com/ghinks/python-minikube-deployments)

Documents case described in issue [1870](https://github.com/kubernetes-client/python/issues/1870) in the example to
avoid others falling into the same issue

I have been out of the Python ecosystem for a long while but have come back as I see advantages for the work I do with
GitOps applications. I would like to follow up on this code comment with possible changes via another PR that would raise a
different error or use the default container if possible. I'm still reading the code and am not ready for that yet.

no

### DIFF
--- a/examples/pod_exec.py
+++ b/examples/pod_exec.py
@@ -71,6 +71,8 @@ def exec_commands(api_instance):
         '/bin/sh',
         '-c',
         'echo This message goes to stderr; echo This message goes to stdout']
+    # When calling a pod with multiple containers running the target container
+    # has to be specified with a keyword argument container=<name>.
     resp = stream(api_instance.connect_get_namespaced_pod_exec,
                   name,
                   'default',


cy-lee:
Anyone review this problem? It will be very useful for new developers.

ghinks:
Many thanks, I'm up for digging further to provide an example for use with sidecars or altering the code to use the default-container if sidecars are present.

PR #1989, opened 2023-01-22
Fix an OverflowError in readline_channel and enhance timeout handling

Masterchen09:
#### What type of PR is this?

/kind bug

This PR fixes an OverflowError caused by the readline_channel method in WSClient, when None is provided as a timeout value. Currently when None is provided, the timeout is internally set to infinity. The poll method and the select function in the update method cannot work with infinite timeouts, instead None or a negative value should be provided (see [here](https://docs.python.org/3/library/select.html#select.select) and [here](https://docs.python.org/3/library/select.html#select.poll.poll)). Additionally the overall timeout handling is enhanced to be more consistent with the peek_channel method. There was also an inconsistency between the usage of the select function and the poll method in the update method when negative timeouts are provided, because the select function does not work with negative timeouts, this is also fixed.

N/A

```release-note
Fix an OverflowError caused by WSClient.readline_channel and improved the overall handling of timeouts
```

```docs

```


yliaog:
please add unit test cases in ws_client_test.py

Masterchen09:
@yliaog I am not sure how I could test the websocket in an unit test (there aren't any test cases related to the websocket in ws_client_test.py yet)...however I have noticed that the websocket is more or less already tested in e2e_test/test_client.py, therefore I have added some more checks there to test different timeout values when using readline_channel/readline_stdout.

PR #2018, opened 2023-02-26
Fix snapshot version to be compliant with PEP440.

### LINKED ISSUE #2014: Problem with tests - Invalid version: '26.0.0-snapshot'

The latest version of setuptools refuses to install the library because the "master" version is not compliant with PEP 440. You can see it in this log: https://github.com/kubernetes-client/python/actions/runs/4217586381/jobs/7321474234

I suggest using ''26.0.0+snapshot" instead of "26.0.0-snapshot". What do you think?
I'll prepare PR if you agree. Thanks.

Related issue: https://github.com/pypa/setuptools/issues/3772

All tests pass.

`tox -e py`

```
.......
   File "/home/runner/work/python/python/.tox/.pkg/lib/python3.11/site-packages/setuptools/command/egg_info.py", line 219, in finalize_options
    parsed_version = parse_version(self.egg_version)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/python/python/.tox/.pkg/lib/python3.11/site-packages/pkg_resources/_vendor/packaging/version.py", line 197, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: '26.0.0-snapshot'
Backend: run command get_requires_for_build_editable with args {'config_settings': None}
running egg_info
Backend: Wrote response {'code': 1, 'exc_type': 'InvalidVersion', 'exc_msg': "Invalid version: '26.0.0-snapshot'"} to /tmp/pep517_get_requires_for_build_editable-p92ljork.json
.pkg: _exit> python /opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  py: FAIL code 1 (12.62 seconds)
  evaluation failed :( (13.14 seconds)
```

(n/a)

**Environment**:

(n/a)

tomplus:
#### What type of PR is this?

The latest version of setuptools enforces PEP440, where version "X.Y-snapshot" is not allowed. The plus sign can be used to separate the local version identifier.

/kind bug
/kind failing-test

Fixes #2014 

NONE

NONE


setup.py
```diff
@@ -16,7 +16,7 @@
 
 # Do not edit these constants. They will be updated automatically
 # by scripts/update-client.sh.
```
yliaog, line 18: i think scripts/update-client.sh needs to be updated too

tomplus: It looks like it uses constants.py to get the client version and update the setup.py. Am I missing something?

yliaog: ok. that's right.

yliaog:
thanks for fixing it.

/lgtm
/approve

PR #2023, opened 2023-03-02
Fix pod_namespace_watch example

attilapiros:
#### What type of PR is this?

/kind bug

The documentation of this example code says:
> The script will wait for 10 events related to namespaces to occur within
the `timeout_seconds` threshold and then move on to **wait for another 10 events**
related to pods to occur within the `timeout_seconds` threshold.

To wait for another 10 events this counter must be set again.

NONE

### DIFF
--- a/examples/watch/pod_namespace_watch.py
+++ b/examples/watch/pod_namespace_watch.py
@@ -44,6 +44,7 @@ def main():
             w.stop()
     print("Finished namespace stream.")
 
+    count = 10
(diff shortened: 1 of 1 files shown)


examples/watch/pod_namespace_watch.py
```diff
@@ -44,6 +44,7 @@ def main():
             w.stop()
```
roycaihw, line 44: if 10 namespace events were printed and `w.stop()` got called. Does the example continue to print the pod events? Could you verify that?

Issue #2024, opened 2023-03-03
Deprecation warning about urllib3 HTTPResponse.getheaders()


jaypipes:
**How to reproduce it (as minimally and precisely as possible)**:

We use the fabulous Kubernetes python client in the AWS Controllers for Kubernetes e2e test framework.

I've noticed a deprecation warning around the use of urllib3's `HTTPResponse.getheaders()` method:

```
DeprecationWarning: HTTPResponse.getheaders() is deprecated and will be removed in urllib3 v2.1.0. Instead access HTTPResponse.headers directly.
```

Easy enough fix so I submitted a pull request to address it.

jaypipes:
Hmmm, after doing some more investigation, looks like upstream urllib3's v2.0 release is significantly delayed (was supposed to be released in 2021). I think at least for the immediate future, it would be better for us to just silence the specific deprecation warning on our (ACK's) side.

sergiitk:
Hey @roycaihw. Openapi fixed this deprecation warning: https://github.com/OpenAPITools/openapi-generator/pull/14743, https://github.com/OpenAPITools/openapi-generator/blame/0bcf9d8bde878b7122478483148c40dd322e0092/modules/openapi-generator/src/main/resources/python/rest.mustache#L30.

Would it be possible to regenerate the docs, and make a patch release for the latest version?

PR #2025, opened 2023-03-03
fix deprecation warn for HTTPResponse.getheaders()

jaypipes:
Running the Kubernetes python client sometimes results in deprecation warnings being shown like this:

```
DeprecationWarning: HTTPResponse.getheaders() is deprecated and will be removed in urllib3 v2.1.0. Instead access HTTPResponse.headers directly.
```

This patch addresses this deprecation warning by accessing the HTTPResponse.headers field (which is a dict) directly.

Issue #2024

/kind cleanup
/kind deprecation

### DIFF
--- a/kubernetes/client/rest.py
+++ b/kubernetes/client/rest.py
@@ -41,11 +41,11 @@ class RESTResponse(io.IOBase):
 
     def getheaders(self):
         """Returns a dictionary of the response headers."""
-        return self.urllib3_response.getheaders()
+        return self.urllib3_response.headers
 
     def getheader(self, name, default=None):
         """Returns a given response header."""
-        return self.urllib3_response.getheader(name, default)
+        return self.urllib3_response.headers.getdefault(name, default)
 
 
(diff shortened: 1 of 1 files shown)


kubernetes/client/rest.py
```diff
@@ -41,11 +41,11 @@ def __init__(self, resp):
 
     def getheaders(self):
         """Returns a dictionary of the response headers."""
-        return self.urllib3_response.getheaders()
```
roycaihw, line 44: this file is generated by openapi-generator. Could you try sending a fix to the upstream? Otherwise we need to carry a patch to make sure this change doesn't get reverted during new releases

jaypipes: Hmmm, after doing some more investigation, looks like upstream urllib3's v2.0 release is significantly delayed (was supposed to be released in 2021). I think at least for the immediate future, it would be better for us to just silence the specific deprecation warning on our (ACK's) side. I'll close out the PR and issue.

PR #2059, opened 2023-05-22
fix: use dl.k8s.io, not kubernetes-release bucket

rjsadow:
#### What type of PR is this?
/kind cleanup
/priority important-soon

There are still references to https://storage.googleapis.com/kubernetes-release instead of https://dl.k8s.io/

One reference was missed from https://github.com/kubernetes-client/python/pull/2055

dl.k8s.io is the correct advertised download host and will eventually move to be fastly shielding a fully community-owned bucket

ref: https://github.com/kubernetes/k8s.io/issues/2396

```release-note
none
```

```docs

```

### DIFF
--- a/scripts/kube-init.sh
+++ b/scripts/kube-init.sh
@@ -71,7 +71,7 @@ echo "Checking docker service"
 sudo docker ps
 
 echo "Download Kubernetes CLI"
-wget -q -O kubectl "http://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/linux/amd64/kubectl"
+wget -q -O kubectl "http://dl.k8s.io/release/${K8S_VERSION}/bin/linux/amd64/kubectl"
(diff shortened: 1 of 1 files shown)


scripts/kube-init.sh
```diff
@@ -71,7 +71,7 @@ echo "Checking docker service"
 sudo docker ps
 
 echo "Download Kubernetes CLI"
-wget -q -O kubectl "http://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/linux/amd64/kubectl"
```
tomplus, line 74: What about minkube? It's 5 lines bellow...

rjsadow: That's in a different bucket that hasn't been migrated behind the reverse proxy yet.

tomplus: Thanks!

PR #2088, opened 2023-07-25
fix socks proxy

### LINKED ISSUE #1064: proxy SOCKS5 schema support
It would be nice if the client supports socks5 proxy to be on par with the kubectl which respects and uses properly env. proxy var of the sort of http_proxy=socks5://..... 

it would be sufficient if the code checks for the proxy schema in configuration.proxy and use urllib3.contrib.socks.SOCKProxyManager instead of urllib3.ProxyManager accordingly.

light4:
#### What type of PR is this?

/kind bug
/kind flake

Fixes #1064 

```release-note
support socks proxy
```

```docs

```

### DIFF
--- a/kubernetes/client/rest.py
+++ b/kubernetes/client/rest.py
@@ -19,13 +19,21 @@ import re
 import ssl
 
 import certifi
+
 # python 2 and python 3 compatibility library
 import six
-from six.moves.urllib.parse import urlencode
 import urllib3
+from requests.utils import should_bypass_proxies
+from six.moves.urllib.parse import urlencode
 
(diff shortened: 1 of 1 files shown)


kubernetes/client/rest.py
```diff
@@ -19,13 +19,21 @@
 import ssl
 
 import certifi
+
```
roycaihw, line 22: This file is generated using openapi-generator (see the top of this file). Could you send the fix to the upstream?

light4: ok

Issue #2089, opened 2023-07-26
TypeError: Object of type 'datetime' is not JSON serializable


younver:
**What happened (please include outputs or screenshots)**:
I've faced `TypeError: Object of type 'datetime' is not JSON serializable (with serialize function)` while passing `V1Pod.to_dict()` into `JSONResponse` in `fastapi.responses`

**What you expected to happen**:
I expected to not to face any serialisation error.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create `V1Pod` object
2. Convert it to python dict using `to_dict()` function
3. Pass converted dict into `JSONResponse` inside `fastapi` router
4. Trigger the router after starting `fastapi` application

**Anything else we need to know?**:
I'm opening a PR to solve this issue however there may be some cases for other models aswell.

**Environment**:
- Kubernetes version (`kubectl version`): `1.27.3`
- OS (e.g., MacOS 10.13.6): `MacOS 12.6.5`
- Python version (`python --version`): `3.11.0`
- Python client version (`pip list | grep kubernetes`): `23.6.0`

PR #2090, opened 2023-07-26
fix: type `datetime.datetime` handling on `to_dict` of `v1_pod`

younver:
#### What type of PR is this?
/kind bug

This PR fixes a type error during the json serialisation of `datetime.datetime` in `V1Pod.to_dict()` by converting datetime object into string using `isoformat()` function which is compatible with the repos format.

Fixes https://github.com/kubernetes-client/python/issues/2089

```release-note
NONE
```

### DIFF
--- a/kubernetes/client/models/v1_pod.py
+++ b/kubernetes/client/models/v1_pod.py
@@ -12,6 +12,7 @@
 
 import pprint
 import re  # noqa: F401
+from datetime import datetime
 
 import six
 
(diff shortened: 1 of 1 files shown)


yliaog:
The version of the OpenAPI document: release-1.27
    Generated by: https://openapi-generator.tech

======================

The file is generated. please fix it in the generator.

kubernetes/client/models/v1_pod.py
```diff
@@ -12,6 +12,7 @@
 
 import pprint
 import re  # noqa: F401
+from datetime import datetime
```
roycaihw, line 15: This file is generated using openapi-generator (see the top of this file). Could you send the fix to the upstream?

younver: my bad, thank you for clarification. do you mean openapi by upstream or something else?

Ananya2001-an:
closing this PR since fix has to be sent to the upstream repo
/close

PR #2098, opened 2023-08-03
Expose websocket header to support base64.channel.k8s.io or v4.base64.channel.k8s.io channel protocol

DrAuYueng:
#### What type of PR is this?
/kind feature

The channel protocols supported by k8s include:
- channel.k8s.io
- base64.channel.k8s.io
- v4.channel.k8s.io
- v4.base64.channel.k8s.io

Currently, it is not possible to pass the websocket request header sec-websocket-protocol to wsclient because it cannot be changed in the generated code, and only v4.channel.k8s.io can be used by default in [ws_client.py](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/stream/ws_client.py#L450).

The sec-websocket-protocol header needs to be exposed to support base64-encoded messages.

```release-note
NONE
```

```docs

```
Signed-off-by: DrAuYueng [ouyang1204@gmail.com](mailto:ouyang1204@gmail.com)

### DIFF
--- a/kubernetes/base/stream/ws_client.py
+++ b/kubernetes/base/stream/ws_client.py
@@ -25,6 +25,7 @@ import time
 
 import six
 import yaml
+import base64
 
 from six.moves.urllib.parse import urlencode, urlparse, urlunparse
 from six import StringIO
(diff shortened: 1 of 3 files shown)


DrAuYueng:
E2E test failed, let me check.

kubernetes/base/stream/ws_client.py
```diff
@@ -200,7 +206,9 @@ def update(self, timeout=0):
                 if six.PY3:
                     data = data.decode("utf-8", "replace")
                 if len(data) > 1:
-                    channel = ord(data[0])
+                    channel = data[0]
```
DrAuYueng, line 209: See https://github.com/kubernetes/kubernetes/blob/42c05a547468804b2053ecf60a3bd15560362fc2/staging/src/k8s.io/apiserver/pkg/util/wsstream/conn.go#L262C1-L262C1.

PR #2100, opened 2023-08-07
Fix UTF-8 failures in Watch

### LINKED ISSUE #2087: Crash when streaming event with non-UTF-8 data
```
72 [00000]   File "/root/.pyenv/versions/3.10.11/lib/python3.10/site-packages/kubernetes/watch/watch.py", line 165, in stream
72 [00000]     for line in iter_resp_lines(resp):
72 [00000]   File "/root/.pyenv/versions/3.10.11/lib/python3.10/site-packages/kubernetes/watch/watch.py", line 58, in iter_resp_lines
72 [00000]     seg = seg.decode('utf8')
72 [00000] UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 2046-2047: unexpected end of data
```

If there's invalid data in an event, it should at least be _possible_ to discard that data or event and continue. I'm not sure what the default behavior should be; it's possible that something which is currently `str` should be `bytes`, or that we should use the unicode replacement character �, or something else.

I'm afraid I don't currently have a minimal reproducer with real Kubernetes but it should be possible to reproduce with a test by adding a `bytes` object with invalid UTF-8 to https://github.com/kubernetes-client/python/blob/dbe83efe0a631a591a0a5a3fc883ef0e646a293a/kubernetes/watch/watch_test.py#L38-L39

I suspect this is the same issue as https://github.com/kubernetes-client/python-base/issues/245

**Environment**:
- Kubernetes version (`kubectl version`): 1.23
- OS (e.g., MacOS 10.13.6): Linux
- Python version (`python --version`): 3.10.10
- Python client version (`pip list | grep kubernetes`): 26.1.0

davidopic:
#### What type of PR is this?

/kind bug

With the old implementation of `kubernetes.watch.Watch.stream()`, users may experience intermittent crashes when subscribed to resources that include multi-byte UTF-8 characters (for example, `ConfigMap`s with non-ASCII characters in the `.data.*` fields). The bug lies in `kubernetes.watch.iter_resp_lines`, which decodes each `bytes` segment to a UTF-8 string without waiting for the rest of the event to be received. When that `bytes` segment ends with only part of a multi-byte sequence, the `.decode` step will fail. Here's how this symptom presents:

```
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 2046-2047: unexpected end of data
```

The new implementation of `iter_resp_lines` buffers partial events in a `bytearray` rather than a `str`, decoding only after a complete event has been received. Furthermore, the call to `.decode` now includes the argument, `errors="replace"`, which uses the canonical UTF-8 fallback character, �, when encountering invalid UTF-8.

To confirm the fix, I added three new tests, two of which failed using the old implementation.

Fixes #2087

`iter_resp_lines` is only called by `Watch.stream`, and the interface did not change.

The old code implicitly assumed that only `bytes` and `str` were valid segment types, and that assumption has now been made explicit with a `TypeError`.

I only tested with python 3.11. I assume that CI will test the code against each of the officially supported python versions. If that's not the case, my code should be reviewed with an eye for any new features I unknowingly relied upon.

```release-note
Fix UTF-8 failures in Watch
```

```docs
N/A
```

### DIFF
--- a/kubernetes/base/watch/watch.py
+++ b/kubernetes/base/watch/watch.py
@@ -52,20 +52,33 @@ def _find_return_type(func):
 
 
 def iter_resp_lines(resp):
-    prev = ""
-    for seg in resp.stream(amt=None, decode_content=False):
-        if isinstance(seg, bytes):
-            seg = seg.decode('utf8')
-        seg = prev + seg
-        lines = seg.split("\n")
-        if not seg.endswith("\n"):
-            prev = lines[-1]
-            lines = lines[:-1]
+    buffer = bytearray()
+    for segment in resp.stream(amt=None, decode_content=False):
+
+        # Append the segment (chunk) to the buffer
+        #
+        # Performance note: depending on contents of buffer and the type+value of segment,
+        # encoding segment into the buffer could be a wasteful step. The approach used here
+        # simplifies the logic farther down, but in the future it may be reasonable to
+        # sacrifice readability for performance.
+        if isinstance(segment, bytes):
+            buffer.extend(segment)
+        elif isinstance(segment, str):
+            buffer.extend(segment.encode("utf-8"))
         else:
-            prev = ""
-        for line in lines:
+            raise TypeError(
+                f"Received invalid segment type, {type(segment)}, from stream. Accepts only 'str' or 'bytes'.")
+
+        # Split by newline (safe for utf-8 because multi-byte sequences cannot contain the newline byte)
+        next_newline = buffer.find(b'\n')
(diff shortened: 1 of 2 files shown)


superosku:
Hi! Any updates on this?

dolshevsk:
Also interested in this PR to be merged

superosku:
@roycaihw any updates regarding this?

roycaihw:
Sorry for the late response. LGTM! Thanks for the contribution!

/lgtm
/approve

Issue #2129, opened 2023-09-28
Allow both urllib3 1 and urllib3 2 in the kubernetes python package


gibsondan:
**What is the feature and why do you need it**:
https://github.com/kubernetes-client/python/commit/27459d5c123c20b096d838ebb98e735134a2a981 implies that the urllib3 pin was added because google-auth doesn't support urllib3 2. However, as of the latest release earlier this week, google-auth now supports that version: https://github.com/googleapis/google-auth-library-python/pull/1389#issuecomment-1738056081 (in fact, it has a lower bound pin, although they are in the process of changing that). I believe that change can now be reverted.

**Describe the solution you'd like to see**:
The kubernetes package supporting both urllib3 1 and urllib3 2.

PR #2131, opened 2023-10-04
Drop urllib3 upper bound

hauntsaninja:
#### What type of PR is this?

/kind cleanup

This bound was added in https://github.com/kubernetes-client/python/commit/27459d5c123c20b096d838ebb98e735134a2a981.

Note that google-auth now requires urllib3>=2 in https://github.com/googleapis/google-auth-library-python/pull/1389

(I'm also not sure that I follow the logic in https://github.com/kubernetes-client/python/pull/2105, since dependency resolvers would have been able to work out a compatible set of packages based on google-auth's dependencies without propagating the bound into this project)

Fixes https://github.com/kubernetes-client/python/issues/2129

```release-note
NONE
```


ingvaldlorentzen:
Could we get a release with this change? Would solve a lot of dependency resolving issues

PR #2167, opened 2023-12-19
Adds UTC defaults to TimezoneInfo class constructor

kpitzen:

This PR offers reasonable defaults to the `TimezoneInfo` class in `dateutils`. The constructor itself performs input validations on these arguments, but sets the timezone name to `UTC` by default - if we're defaulting internal representations of these arguments, it makes sense to allow for a default constructor to be exposed via `TimezoneInfo()` which uses those same defaults.

Fixes #

None

```release-note
Added UTC default values to TimezoneInfo class constructor arguments
```

```docs

```

### DIFF
--- a/kubernetes/base/config/dateutil.py
+++ b/kubernetes/base/config/dateutil.py
@@ -18,7 +18,7 @@ import re
 
 
 class TimezoneInfo(datetime.tzinfo):
-    def __init__(self, h, m):
+    def __init__(self, h=0, m=0):
         self._name = "UTC"
         if h != 0 and m != 0:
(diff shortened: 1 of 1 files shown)


kubernetes/base/config/dateutil.py
```diff
@@ -18,7 +18,7 @@
 
 
 class TimezoneInfo(datetime.tzinfo):
-    def __init__(self, h, m):
+    def __init__(self, h=0, m=0):
```
roycaihw, line 21: I'm not sure why we need to default the inputs. Is there any use case? Could you elaborate a little?

PR #2194, opened 2024-02-15
Enable binary support for WSClient

### LINKED ISSUE #1471: Allow retrieval of binary data from WSClient output stream

The API client currently does not offer any capabilities for copying files from Pods / containers (similiar to `kubectl cp`).
We have created a workaround by opening a stream to the Pod, creating a tar archive of the files to copy, and outputting the data to stdout. Once retrieved, the data is then extracted from the archive.

However, this does not work for binary files due to how the WSClient handles incoming data.

`kubernetes\stream\ws_client.py:161-178` (kubernetes v17.17.0)

```python
    def update(self, timeout=0):
        """Update channel buffers with at most one complete frame of input."""
        if not self.is_open():
            return
        if not self.sock.connected:
            self._connected = False
            return
        r, _, _ = select.select(
        if r:
            op_code, frame = self.sock.recv_data_frame(True)
            if op_code == ABNF.OPCODE_CLOSE:
                self._connected = False
                return
            elif op_code == ABNF.OPCODE_BINARY or op_code == ABNF.OPCODE_TEXT:
                data = frame.data
                if six.PY3:
                    data = data.decode("utf-8", "replace")
```

The last line always tries to decode to UTF-8 while replacing all characters that cannot be properly decoded.
In the case of binary PCAP files, this results in corrupted data.

I think that changing the signature of `update` to allow the user the following options might resolve the issue:
1. Choose a different error handling than `replace`. According to the [docs](https://docs.python.org/3/library/stdtypes.html#bytes.decode), `strict` and `ignore` are options as well. I tested out `ignore` and it results in the desired output when decoding and then encoding again, which both byte object being the same.
2. Allow the user to skip conversion alltogether. This would mean adding a flag which would cause skipping of 
   ```python
   if six.PY3:
       data = data.decode("utf-8", "replace")
   ```

If needed, I can implement the agreed on solution myself and open a pull request.

meln5674:
#### What type of PR is this?

/kind feature

Currently, under python 3, the WSClient decodes all data via UTF-8. This will break, e.g. capturing the stdout of tar or gzip. This adds a new 'binary' kwarg to the WSClient class and websocket_call function. If this is set to true, then the decoding will not happen, and all channels will be interpreted as binary.
This does raise a slight complication, as the OpenAPI-generated client will convert the output to a string, no matter what, which it ends up doing by (effectively) calling repr(). This requires a bit of magic to recover the orignial bytes, and is inefficient. However, this is only the case when using the default _preload_content=True, when setting this to False and manually calling read_all or read_channel, this issue does not arise.

Fixes #1471

```release-note
Added 'binary' keyword argument to WSClient to use binary channels instead of string (utf8).
```

```docs

```

### DIFF
--- a/kubernetes/base/stream/ws_client.py
+++ b/kubernetes/base/stream/ws_client.py
@@ -26,8 +26,9 @@ import time
 import six
 import yaml
 
+
 from six.moves.urllib.parse import urlencode, urlparse, urlunparse
-from six import StringIO
+from six import StringIO, BytesIO
 
 from websocket import WebSocket, ABNF, enableTrace
 from base64 import urlsafe_b64decode
@@ -48,7 +49,7 @@ class _IgnoredIO:
 
 
 class WSClient:
-    def __init__(self, configuration, url, headers, capture_all):
+    def __init__(self, configuration, url, headers, capture_all, binary=False):
         """A websocket client with support for channels.
 
             Exec command uses different channels for different streams. for
@@ -58,8 +59,10 @@ class WSClient:
         """
         self._connected = False
         self._channels = {}
+        self.binary = binary
+        self.newline = '\n' if not self.binary else b'\n'
         if capture_all:
-            self._all = StringIO()
+            self._all = StringIO() if not self.binary else BytesIO()
         else:
             self._all = _IgnoredIO()
         self.sock = create_websocket(configuration, url, headers)
@@ -92,8 +95,8 @@ class WSClient:
         while self.is_open() and time.time() - start < timeout:
             if channel in self._channels:
                 data = self._channels[channel]
-                if "\n" in data:
-                    index = data.find("\n")
+                if self.newline in data:
+                    index = data.find(self.newline)
                     ret = data[:index]
                     data = data[index+1:]
                     if data:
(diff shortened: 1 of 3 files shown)


kubernetes/base/stream/ws_client.py
```diff
@@ -518,13 +523,17 @@ def websocket_call(configuration, _method, url, **kwargs):
     _request_timeout = kwargs.get("_request_timeout", 60)
     _preload_content = kwargs.get("_preload_content", True)
     capture_all = kwargs.get("capture_all", True)
-
+    binary = binary=kwargs.get('binary', False)
```
yliaog, line 526: typo?

kubernetes/e2e_test/test_client.py
```diff
@@ -154,6 +169,32 @@ def test_pod_apis(self):
         resp.update(timeout=5)
         self.assertFalse(resp.is_open())
 
+        resp = stream(api.connect_post_namespaced_pod_exec, name, 'default',
+                      command='/bin/sh',
+                      stderr=True, stdin=True,
+                      stdout=True, tty=False,
+                      binary=True,
+                      _preload_content=False)
```
yliaog, line 177: add a test case that is the same as this one, but with _preload_content=True ?

kubernetes/base/stream/ws_client.py
meln5674: Indeed, fixed.

kubernetes/e2e_test/test_client.py
meln5674: The test above this one leaves _preload_test as its default value of True. Using update() and write_stdin(), read_stdout(), and read_stderr() are not valid when this is the case.

yliaog: ok

yliaog:
thanks for the PR

/lgtm
/approve

PR #2204, opened 2024-03-01
raise `ConfigException` on invalid named-arguments to `load_incluster_config`

### LINKED ISSUE #2203: `load_config` does not really support "context"
similar to  #2038

passing `context` as an argument to `load_config` -- even if that argument is `None` does not work because `load_incluster_config` does not list it as an argument

A `context` argument should be able to be passed into `load_incluster_config`  and if it's value is non-None, then perhaps a `ConfigException` should be raised

* start a bare ubuntu lxd,
* create a virtual environment
* pip install kubernetes
* pass an `context=None`, since there is no `.kube/config` it will try to load the incluster config and fail
```
(venv) ubuntu@focal:~$ python -c "import kubernetes.config; kubernetes.config.load_config(context=None)"
kube_config_path not provided and default location (~/.kube/config) does not exist. Using inCluster Config. This might not work.
  File "<string>", line 1, in <module>
  File "/home/ubuntu/venv/lib/python3.8/site-packages/kubernetes/config/__init__.py", line 49, in load_config
    load_incluster_config(**kwargs)
TypeError: load_incluster_config() got an unexpected keyword argument 'context'
```

**Environment**:
- Kubernetes version (`kubectl version`): N/A
- OS (e.g., MacOS 10.13.6): N/A
- Python version (`python --version`): py3.8.10 (but doesn't matter)
- Python client version (`pip list | grep kubernetes`):
```
venv) ubuntu@focal:~$ pip list | grep kubernetes
kubernetes         29.0.0  
```
-

addyess:
#### What type of PR is this?
/kind bug
/kind api-change

Gracefully ignore named-arguments to `load_incluster_config` which have a value of `None`

Fixes #2203

I couldn't find any unit tests for calling `load_incluster_config` directly.  Let me know if i should add some.

No, it should be transparent.  The exception changes from a TypeError to ConfigException

Before:
```
...
  File "/home/runner/work/pytest-operator/pytest-operator/.tox/integration-2.9/lib/python3.10/site-packages/kubernetes/config/__init__.py", line 49, in load_config
    load_incluster_config(**kwargs)
TypeError: load_incluster_config() got an unexpected keyword argument 'context'
```
to
```
  File "/home/runner/work/pytest-operator/pytest-operator/.tox/integration-2.9/lib/python3.10/site-packages/kubernetes/config/__init__.py", line 49, in load_config
    load_incluster_config(**kwargs)
ConfigException: Unimplemented named-argument context for incluster config
```

### DIFF
--- a/kubernetes/base/config/incluster_config.py
+++ b/kubernetes/base/config/incluster_config.py
@@ -115,7 +115,7 @@ def load_incluster_config(client_configuration=None, try_refresh_token=True, **k
     cluster. It's intended for clients that expect to be running inside a pod
     running on kubernetes. It will raise an exception if called from a process
     not running in a kubernetes environment."""
-    for val in kds.values():
+    for val in kwds.values():
         if val is not None:
             raise ConfigException(f"Unimplemented named-argument {val} for incluster config.")
     InClusterConfigLoader(
--- a/kubernetes/base/config/incluster_config.py
+++ b/kubernetes/base/config/incluster_config.py
@@ -109,12 +109,15 @@ class InClusterConfigLoader(object):
             ) + self._token_refresh_period
 
 
-def load_incluster_config(client_configuration=None, try_refresh_token=True):
+def load_incluster_config(client_configuration=None, try_refresh_token=True, **kwds):
     """
     Use the service account kubernetes gives to pods to connect to kubernetes
     cluster. It's intended for clients that expect to be running inside a pod
     running on kubernetes. It will raise an exception if called from a process
     not running in a kubernetes environment."""
+    for val in kds.values():
+        if val is not None:
+            raise ConfigException(f"Unimplemented named-argument {val} for incluster config")
     InClusterConfigLoader(
         token_filename=SERVICE_TOKEN_FILENAME,
         cert_filename=SERVICE_CERT_FILENAME,
--- a/kubernetes/base/config/incluster_config.py
+++ b/kubernetes/base/config/incluster_config.py
@@ -117,7 +117,7 @@ def load_incluster_config(client_configuration=None, try_refresh_token=True, **k
     not running in a kubernetes environment."""
     for val in kds.values():
         if val is not None:
-            raise ConfigException(f"Unimplemented named-argument {val} for incluster config")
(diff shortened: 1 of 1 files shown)


addyess:
@fabianvf and @yliaog 
thanks for taking the moment to review this easy fix

kubernetes/base/config/incluster_config.py
```diff
@@ -109,12 +109,15 @@ def _read_token_file(self):
...
 
 
-def load_incluster_config(client_configuration=None, try_refresh_token=True):
+def load_incluster_config(client_configuration=None, try_refresh_token=True, **kwds):
     """
     Use the service account kubernetes gives to pods to connect to kubernetes
     cluster. It's intended for clients that expect to be running inside a pod
     running on kubernetes. It will raise an exception if called from a process
     not running in a kubernetes environment."""
+    for val in kds.values():
```
yliaog, line 118: there is no kds, do you mean kwds?

kubernetes/base/config/incluster_config.py
```diff
@@ -109,12 +109,15 @@ def _read_token_file(self):
...
+        if val is not None:
+            raise ConfigException(f"Unimplemented named-argument {val} for incluster config.")
```
yliaog, line 120: how is ConfigException better than TypeError ?

kubernetes/base/config/incluster_config.py
addyess: yep!  `kwds`

```suggestion
    for val in kwds.values():
```

kubernetes/base/config/incluster_config.py
addyess: its only slightly better because the client code only has to catch one exception --  `ConfigException`

Secondarily, this will allow callers to: [`load_config`](https://github.com/kubernetes-client/python/blob/851dc2a0b10d1bc07826c6ade489d1bd2ef431c3/kubernetes/base/config/__init__.py#L24)

to call with a keyword arg like `context=None` and `load_incluster_config` won't bark with the `TypeError`, but if there is a `context=<val>` that's non-none, we'll get a meaningful exceptions from `load_incluster_config`

PR #2210, opened 2024-03-15
Don't delete all nodes when empty string is passed

### LINKED ISSUE #2209: kubernetes.client.api.core_v1_api.delete_node_with_http_info deletes all nodes when `name` is an empty string.

**The python client library allows deletion of all nodes in a cluster when passing in an empty string. THIS IS BAD!!**

As part of code we use, we call:

```
k8s_client = kubernetes.client.ApiClient(configuration=client_config)
k8s_client.delete_node(node_name, body=kubernetes.client.V1DeleteOptions())
```
When we do this, we discovered an issue where we accidentally passed in a `node_name=""`, which leads to this message:

```
5924       1 httplog.go:132] "HTTP" verb="DELETE" URI="/api/v1/nodes/" latency="182.687862ms" userAgent="OpenAPI-Generator/27.2.0/python" audit-ID="86d16672-050c-4f8d-a720-052e782c049e" srcIP="10.10.1.195:7186" apf_pl="exempt" apf_fs="exempt" apf_execution_time="153.526678ms" resp=200
```

Which turns out to delete all nodes from the API server on the cluster. **OOPSIE**!

Corresponding queries to the API show that our nodes indeed disappear.

The `delete_node`/`delete_node_with_http_info` method should fail if the name is blank `(name="")`.

Open a python shell and import the kubernetes python code. Initialize it and call `delete_node("")` which should raise an ApiValueError

**Environment**:
- Kubernetes version (`kubectl version`): 1.27
- OS (e.g., MacOS 10.13.6): Flatcar Linux, but has been replicated on MacOS and Fedora Linux
- Python version (`python --version`) 3.11.4
- Python client version (`pip list | grep kubernetes`) 27.2.0

herlo:
Fixed a bug that allows an empty string for a node name when calling the `delete_node`/`delete_node_with_http_info` method.

#### What type of PR is this?

/kind bug

https://github.com/kubernetes-client/python/issues/2209 explains the issue well.

Fixes #2209 

```
NONE
```

### DIFF
--- a/scripts/core_v1_api.diff
+++ b/scripts/core_v1_api.diff
@@ -1,0 +1,16 @@
+diff --git a/kubernetes/client/api/core_v1_api.py b/kubernetes/client/api/core_v1_api.py
+index 282d9b88..8de09a11 100644
+--- a/kubernetes/client/api/core_v1_api.py
++++ b/kubernetes/client/api/core_v1_api.py
+@@ -13680,7 +13680,10 @@ class CoreV1Api(object):
+         del local_var_params['kwargs']
+         # verify the required parameter 'name' is set
+         if self.api_client.client_side_validation and ('name' not in local_var_params or  # noqa: E501
+-                                                        local_var_params['name'] is None):  # noqa: E501
++                                                        local_var_params['name'] is None or  # noqa: E501
++                                                        not len(local_var_params['name'].strip())):  # noqa: E501
++            raise ApiValueError("Missing the required parameter `name` when calling `delete_node`")  # noqa: E501
++        if not len(name):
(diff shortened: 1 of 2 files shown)


roycaihw:
This behavior is probably happening for all APIs in this client. 

We are discussing if this is expected behavior or something we want to change. One idea is such an empty string validation can happen at a higher level.

/cc @yliaog @roycaihw @lumingx

roycaihw:
@herlo Could you verify if this behavior applies to other APIs as well? If so, it's probably unrealistic to patch all the delete methods in this client.

PR #2215, opened 2024-03-30
replace utcnow with now, due to utcnow will be deprecated

### LINKED ISSUE #2213: Deprecation warning datetime.datetime.utcnow().
We are using using this library for some e2e testing and receive a few deprecation warnings about this line

https://github.com/kubernetes-client/python/blame/851dc2a0b10d1bc07826c6ade489d1bd2ef431c3/kubernetes/base/config/kube_config.py#L83

The warning looks like this:
```
(...)
/my/paht/kubernetes/config/kube_config.py:83: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

python version: `python3.12`
using venv and installed the library with pip via requirements.txt file.
```
(...)
kubernetes==29
(...)
```

showjason:
datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version

Fixes #2213

```release-note

```

```docs

```

### DIFF
--- a/kubernetes/base/config/kube_config_test.py
+++ b/kubernetes/base/config/kube_config_test.py
@@ -30,7 +30,7 @@ from six import PY3, next
 from kubernetes.client import Configuration
 
 from .config_exception import ConfigException
-from .dateutil import format_rfc3339, parse_rfc3339
+from .dateutil import UTC, format_rfc3339, parse_rfc3339
 from .kube_config import (ENV_KUBECONFIG_PATH_SEPARATOR, CommandTokenSource,
                           ConfigNode, FileOrData, KubeConfigLoader,
                           KubeConfigMerger, _cleanup_temp_files,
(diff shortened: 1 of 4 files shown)


kubernetes/base/config/kube_config_test.py
```diff
@@ -1028,7 +1028,7 @@ def test_load_gcp_token_no_refresh(self):
     def test_load_gcp_token_with_refresh(self):
         def cred(): return None
         cred.token = TEST_ANOTHER_DATA_BASE64
-        cred.expiry = datetime.datetime.utcnow()
+        cred.expiry = datetime.datetime.now(tz=UTC).replace(tzinfo=None)
```
roycaihw, line 1031: Could you elaborate why do we do `.replace(tzinfo=None)`?

roycaihw:
Had one question. LGTM otherwise.
/release-note-none

kubernetes/base/config/kube_config_test.py
roycaihw: Got it. Thanks for the explanation!

PR #2224, opened 2024-05-03
remove some six usage

a-detiste:
/kind cleanup 

six was only needed to _also_ support Python2 with the same codebase.

https://wiki.debian.org/Python3-six-removal

I will continue if this first one succeeds.

```
grep 'six.iteritems(self.openapi_types)' -rl | while read f do
    sed -i 's|six.iteritems(self.openapi_types)|self.openapi_types.items()|' $f
done

pyflakes3 . | grep "'six' imported but unused" | cut -d: -f1 | while read f do
    sed -i '/^import six$/d' $f
done
```


kubernetes/client/models/admissionregistration_v1_service_reference.py
```diff
@@ -13,7 +13,6 @@
 import pprint
 import re  # noqa: F401
 
-import six
```
roycaihw, line 16: The file is generated by openapi-generator. The change needs to be made upstream. Otherwise the change will be override when we re-generate the client

a-detiste: Thanks. I think it's already gone. https://github.com/OpenAPITools/openapi-generator/pull/16675
Wait & see

PR #2225, opened 2024-05-05
Typo in 'host configuration'

behradiy:
/kind documentation

Fixes #
**I noticed that the part 'host configuration' was mistakenly repeated, so I went ahead and removed the duplicate occurrence to avoid any confusion.**

"NONE"  

/release-note-none

docs

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -60,8 +60,6 @@ configuration.api_key['authorization'] = 'YOUR_API_KEY'
 # Defining host is optional and default to http://localhost
 configuration.host = "http://localhost"
 
-# Defining host is optional and default to http://localhost
-configuration.host = "http://localhost"
 # Enter a context with an instance of the API kubernetes.client
 with kubernetes.client.ApiClient(configuration) as api_client:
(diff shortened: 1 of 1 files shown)


kubernetes/README.md
```diff
@@ -60,8 +60,6 @@ configuration.api_key['authorization'] = 'YOUR_API_KEY'
 # Defining host is optional and default to http://localhost
 configuration.host = "http://localhost"
 
-# Defining host is optional and default to http://localhost
-configuration.host = "http://localhost"
```
roycaihw, line 64: This file seems to be generated: https://github.com/kubernetes-client/python/commits/master/kubernetes/README.md. Perhaps this comes from the upstream openapi-generator?

behradiy: According to git blame, this part of README.md ( _lines 60 and 61_) was added by **OpenAPI generator v4.3.0**, but by mistake, _lines 63 and 64_ were added by **Yu Liao** when updating to version **1.26**.
although it was written 4 years ago by OpenAPI generator.

here is the screenshot : 

image

PR #2243, opened 2024-05-30
Updated setup.py

anirudh-hegde:
/kind feature

Enhances the app as I have added Kubernetes packages.

NA

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,9 +60,9 @@ setup(
     tests_require=TESTS_REQUIRES,
     extras_require=EXTRAS,
     packages=['kubernetes', 'kubernetes.client', 'kubernetes.config',
-              'kubernetes.watch', 'kubernetes.client.api',
-              'kubernetes.stream', 'kubernetes.client.models',
-              'kubernetes.utils', 'kubernetes.client.apis',
+              'kubernetes.watch', 'kubernetes.client.api', 'kubernetes.client.api_client'
+              'kubernetes.client.rest', 'kubernetes.stream', 'kubernetes.client.models',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -60,9 +60,9 @@
     tests_require=TESTS_REQUIRES,
     extras_require=EXTRAS,
     packages=['kubernetes', 'kubernetes.client', 'kubernetes.config',
-              'kubernetes.watch', 'kubernetes.client.api',
-              'kubernetes.stream', 'kubernetes.client.models',
-              'kubernetes.utils', 'kubernetes.client.apis',
+              'kubernetes.watch', 'kubernetes.client.api', 'kubernetes.client.api_client'
+              'kubernetes.client.rest', 'kubernetes.stream', 'kubernetes.client.models',
+              'kubernetes_asyncio', 'kubernetes.utils', 'kubernetes.client.apis',
```
roycaihw, line 65: It's unclear to me what issue this PR is addressing. Could you elaborate?

anirudh-hegde: Aims to provide various functionalities necessary for interacting with Kubernetes clusters using the Kubernetes Python client libraries

roycaihw: `kubernetes_asyncio` is not part of this repo. I think you can include it separately in your own environment?

anirudh-hegde: > `kubernetes_asyncio` is not part of this repo. I think you can include it separately in your own environment?

However it's a python package which supports both core Kubernetes API resources and custom resource definitions (CRDs), making it a versatile tool for various Kubernetes-related tasks.

Issue #2030, opened 2023-03-20
Could I use apply in DynamicClient ?


hermitgreen:
**What is the feature and why do you need it**:

Can 'DynamicClient' support 'apply'？

**Describe the solution you'd like to see**:

While I try https://github.com/kubernetes-client/python/blob/master/examples/dynamic-client/service.py example，I replace .create to .apply and except this can create if not exist or patch（like kubectl apply）, but I got

```shell
AttributeError: 'DynamicClient' object has no attribute 'apply
```

how could I use apply?

my code

```python
from kubernetes import config, dynamic
from kubernetes.client import api_client


def main():
    client = dynamic.DynamicClient(
        api_client.ApiClient(configuration=config.load_kube_config())
    )
    api = client.resources.get(api_version="v1", kind="Service")
    name = "frontend-service"
    service_manifest = {
        "apiVersion": "v1",
        "kind": "Service",
        "metadata": {"labels": {"name": name}, "name": name, "resourceversion": "v1"},
        "spec": {
            "ports": [
                {"name": "port", "port": 80, "protocol": "TCP", "targetPort": 80}
            ],
            "selector": {"name": name},
        },
    }
    service = api.apply(body=service_manifest, namespace="default")


if __name__ == "__main__":
    main()
```

yabea:
Can't use apply in DynamicClient: 
https://github.com/kubernetes-client/python/blob/56eea103a761c2310af76c63d4ed9cdd7171aeed/kubernetes/base/dynamic/client.py#L72

You can get() first, if the get succeeds excute patch(), else execute create().

Issue #1737, opened 2022-03-04
Missing an example in how to run a "apply -f"


mazzespazze:
**Link to the issue (please include a link to the specific documentation or example)**:
Most of the resources https://github.com/kubernetes-client, https://github.com/kubernetes-client/python/tree/master/examples are lacking one of the most needed examples: how to perform a `kubectl apply -f file.yaml`

**Description of the issue (please include outputs or screenshots if possible)**:
Currently it seems the API has examples for how to create and delete resources, and most of its examples are around connectivity. An example on the apply -f should be given, especially since these closed tickets should have not been closed at all as no solution was given: 
https://github.com/kubernetes-client/python/issues/1215
https://github.com/kubernetes-client/python/issues/1168

mazinesy:
Adding [server side apply doc](https://kubernetes.io/docs/reference/using-api/server-side-apply/)

RonMallory:
I have submitted a PR adding additional examples for using utils.create_from_directory and utils.create_from_yaml.

I believe this is some of the functionality that you are looking for an examples of.

It will not be the full functionality of kubectl apply -f but it gives you a better example than what was previously up there.

roycaihw:
Thanks @RonMallory 
/close

sm-moore:
I was looking at the examples in the PR Ron mentioned above https://github.com/kubernetes-client/python/pull/1752 and don't see any example of kubectl apply, only kubectl create. It appears to me this library doesn't actually support the apply operation. Is that true?

PR #2252, opened 2024-06-27
Add ability to use server_side_apply to  utils.create_from_yaml

### LINKED ISSUE #2251: Allow utils.create_from_yaml to use server side apply
Currently `utils.create_from_yaml` will only create resources. If the resource already exists then it throws an error.

We could add an `apply` parameter to the functions in the utils class. If we provide a default value of `False` to that parameter, it shouldn't break any existing behavior, but if `apply=True` is passed in then utils could use a dynamic client and use the `server_side_apply` function. This would allow applying yaml manifests directly even if the resources already exist.

I currently have a working model locally and would be willing to contribute  a patch for this if it seems like a good approach to y'all.

A similar approach was discussed in https://github.com/kubernetes-client/python/issues/1168, but it seems it never landed.

dcmcand:
#### What type of PR is this?
/kind feature

Allows `utils.create_from_yaml` functions to optionally use server side apply. This is a feature that has been requested numerous times. It will allow applying a manifest more than once without throwing an error or updating manifests and reapplying (similar to how `kubectl apply` works).

Fixes #2251 
Addresses #2030
Part of #1737

```release-note
Added the ability to use the optional `apply` parameter for functions within the `utils.create_from_yaml` submodule. This allows these functions to optionally use the `DynamicClient.server_side_apply` function to apply yaml manifests.
```

```docs
[Usage]: create_from_yaml(kubernetes_client, manifest, apply=True)
```

### DIFF
--- a/kubernetes/e2e_test/test_utils.py
+++ b/kubernetes/e2e_test/test_utils.py
@@ -71,6 +71,48 @@ class TestUtils(unittest.TestCase):
             except ApiException:
                 continue
 
+    def test_create_apps_deployment_from_yaml_with_apply_is_idempotent(self):
+        """
+        Should be able to create an apps/v1 deployment.
+        """
+        k8s_client = client.api_client.ApiClient(configuration=self.config)
+        try:
+            utils.create_from_yaml(
+                k8s_client, self.path_prefix + "apps-deployment.yaml")
+            app_api = client.AppsV1Api(k8s_client)
+            dep = app_api.read_namespaced_deployment(name="nginx-app",
+                                                    namespace="default")
+            self.assertIsNotNone(dep)
+            self.assertEqual("nginx-app", dep.metadata.name)
+            self.assertEqual("nginx:1.15.4", dep.spec.template.spec.containers[0].image)
+            self.assertEqual(80, dep.spec.template.spec.containers[0].ports[0].container_port)
+            self.assertEqual("nginx", dep.spec.template.spec.containers[0].name)
+            self.assertEqual("nginx", dep.spec.template.metadata.labels["app"])
+            self.assertEqual(3, dep.spec.replicas)
+
+            utils.create_from_yaml(
+                k8s_client, self.path_prefix + "apps-deployment.yaml", apply=True)
+            dep = app_api.read_namespaced_deployment(name="nginx-app",
+                                                    namespace="default")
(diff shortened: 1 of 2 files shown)


dcmcand:
@yliaog, is there anything I can do to help get this looked at?

dcmcand:
@yliaog @roycaihw Is there anything I can do to help get this looked at?

marcelovilla:
Hey @yliaog @roycaihw! We could use this feature in our project. Are you planning to review this anytime soon? Thanks for all the work!

yliaog:
sure, i'll take a look in the next couple of days.

kubernetes/e2e_test/test_utils.py
```diff
@@ -71,6 +71,48 @@ def test_create_apps_deployment_from_yaml(self):
...
+            self.assertIsNotNone(dep)
+            self.assertEqual("nginx-app", dep.metadata.name)
+            self.assertEqual("nginx:1.15.4", dep.spec.template.spec.containers[0].image)
+            self.assertEqual(80, dep.spec.template.spec.containers[0].ports[0].container_port)
+            self.assertEqual("nginx", dep.spec.template.spec.containers[0].name)
+            self.assertEqual("nginx", dep.spec.template.metadata.labels["app"])
+            self.assertEqual(3, dep.spec.replicas)
+
+            utils.create_from_yaml(
+                k8s_client, self.path_prefix + "apps-deployment.yaml", apply=True)
```
yliaog, line 94: the prior deployment is still in the cluster, so this 'apply' does nothing?

dcmcand: correct, we want to make sure that the apply is idempotent and can be applied multiple times without error.

scottrbrtsn:
This is beautiful.  For folks wanting this prior to its release, you can add the lib and point the version to the master branch of this repo. 

 pip: https://stackoverflow.com/questions/20101834/pip-install-from-git-repo-branch
 poetry: https://python-poetry.org/docs/dependency-specification/#git-dependencies

yliaog:
https://github.com/kubernetes/sig-release/tree/master/releases/release-1.32

it will be included in 1.32 release, which is coming soon.

MatthewCane:
Great news! Thank you for the update!

Issue #2264, opened 2024-08-14
Stream ws_client has a bug related to the time library


EyalPazz:
**What happened (please include outputs or screenshots)**:
{OverflowError}OverflowError('timestamp too large to convert to C _PyTime_t')

**What you expected to happen**:
Not Crash

**How to reproduce it (as minimally and precisely as possible)**:
 run WSClient.readline_channel with default values

**Anything else we need to know?**:

the problem is here:
`    
    if timeout is None:
            timeout = float("inf")
`
I fixed it, just want to know if i should open a PR

**Environment**:
- Kubernetes version (`kubectl version`): v1.27.1
- OS (e.g., MacOS 10.13.6): Ubuntu 22.04
- Python version (`python --version`): Python 3.10.12
- Python client version (`pip list | grep kubernetes`): 30.1.0

PR #2265, opened 2024-08-14
fix: time bug in ws_cilent

EyalPazz:
[https://github.com/kubernetes-client/python/issues/2264](#2264
)

### DIFF
--- a/kubernetes/base/stream/ws_client.py
+++ b/kubernetes/base/stream/ws_client.py
@@ -87,7 +87,7 @@ class WSClient:
             del self._channels[channel]
         return ret
 
-    def readline_channel(self, channel, timeout=0):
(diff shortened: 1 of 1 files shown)


kubernetes/base/stream/ws_client.py
```diff
@@ -87,10 +87,8 @@ def read_channel(self, channel, timeout=0):
             del self._channels[channel]
         return ret
 
-    def readline_channel(self, channel, timeout=None):
+    def readline_channel(self, channel, timeout=1):
```
EyalPazz, line 90: timeout can also default to 0 (it won't crash but it won't do anything)

EyalPazz:
@yliaog Can you please have a look?

PR #2269, opened 2024-08-21
More GEP-2257 tweaks

kflynn:
This is best reviewed commit by commit.

First, refactor the formatter for GEP-2257 durations to just do things by hand rather than relying on durationpy's formatter. This is primarily a less fragile approach, but it also allows the three extant GEP-2257 implementations (Rust, Go, and Python) to be more similar.

Second, add an example program. Should've done this earlier.

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/examples/gep2257.py
+++ b/examples/gep2257.py
@@ -1,0 +1,39 @@
+#!/usr/bin/python3
+# -*- coding:utf-8 -*-
+
+"""
+This example uses kubernetes.utils.duration to parse and display
+a GEP-2257 duration string (you can find the full specification at
+https://gateway-api.sigs.k8s.io/geps/gep-2257/).
+
+Good things to try:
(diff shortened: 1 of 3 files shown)


examples/gep2257.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/python3
+# -*- coding:utf-8 -*-
+
+"""
+This example uses kubernetes.utils.duration to parse and display
+a GEP-2257 duration string (you can find the full specification at
+https://gateway-api.sigs.k8s.io/geps/gep-2257/).
+
+Good things to try:
+>>> python examples/gep2257.py 1h
```
yliaog, line 10: nit: could you change the file name to something like duration-gep2257, to make it more clear what it is about ?

kflynn: Done! 🙂

PR #2283, opened 2024-09-16
Change examples/dynamic-client creation of DynamicClient to include configuration object that is not None

### LINKED ISSUE #2279: The dynamic client example, `kubernetes.client.ApiClient(...)` takes in the result of `kubernetes.config.load_kube_config(...)`, which is empty
Example where `client.ApiClient(configuration=config.load_kube_config())` is used:
https://github.com/kubernetes-client/python/blob/master/examples/dynamic-client/configmap.py#L27
Link to where config function `load_kube_config()` does not return anything:
https://github.com/kubernetes-client/python/blob/master/kubernetes/base/config/kube_config.py#L799

I would assume that when configuring an ApiClient, I should send the argument of the configuration object. However, since the `config.load_kube_config()` does not return anything, I feel that this example is wrong.

I would assume I should do the following:
```
from kubernetes import config, dynamic
from kubernetes.client import api_client
configuration = client.Configuration()
config.load_kube_config()
client = api_client.ApiClient(configuration=configuration)
```
Since the `config.load_kube_config()` has this in the description:
"Loads authentication and cluster information from kube-config file and stores them in kubernetes.client.configuration."
So that the `client.Configuration()` returned value is updated after being instantiated and thus can be passed to the `api_client.ApiClient(...)` function.

Can anyone please correct me if I am wrong? I am unsure of what the correct way to get an ApiClient.
Thanks in advance!

tonur:
#### What type of PR is this?
/kind documentation

Changed the examples to showcase DynamicClient instantiation without the parameter since `config.load_kube_config()` does not return anything. See more about this in https://github.com/kubernetes-client/python/issues/2279.

Fixes #2279

```release-note
NONE
```

```docs

```

### DIFF
--- a/examples/dynamic-client/accept_header.py
+++ b/examples/dynamic-client/accept_header.py
@@ -22,12 +22,14 @@ from kubernetes.client import api_client
 
 def main():
     # Creating a dynamic client
-    client = dynamic.DynamicClient(
-        api_client.ApiClient(configuration=config.load_kube_config())
-    )
+    configuration = api_client.Configuration()
+    config.load_kube_config()
+    client = api_client.ApiClient(configuration=configuration)
+    dynamic_client = dynamic.DynamicClient(client)
+
 
     # fetching the node api
-    api = client.resources.get(api_version="v1", kind="Node")
+    api = dynamic_client.resources.get(api_version="v1", kind="Node")
 
     # Creating a custom header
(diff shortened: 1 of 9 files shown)


tonur:
This is currently not working, I am at a loss for words on why. I will try to dive into the `load_kube_config` method to understand how to prepare a configuration object for the `api_client.ApiClient(configuration=configuration)`.

PR #2289, opened 2024-10-05
Added Shell=True for windows flow. Fixes file not found error while r…

### LINKED ISSUE #2276: Kuberenetes Python client doesn't work with AWS EKS (Exec plugin) in Windows
Kubernetes Python client doesn't work with AWS EKS in Windows.
I have AWS EKS clusters created and setup. And my environment to testing is Windows OS. When I run ```config.load_config_file=KUBE_CONFIG_PATH)```, it works well in other OS but in Windows OS, it shows:

```
ERROR:root:[WinError 2] The system cannot find the file specified
```

**I identified issue is actually coming from [here](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/config/exec_provider.py#L71-L78), in Windows OS seems like we need to have ```shell=True``` argument also added when executing command or it will fail**

Expect Kubernetes Python Client should work for all platforms

- Follow https://docs.aws.amazon.com/eks/latest/userguide/create-cluster.html to create AWS EKS cluster
- Follow https://docs.aws.amazon.com/eks/latest/userguide/create-kubeconfig.html to update local ```kubeconfig``` file
- Run Python interactive program: ```python```
- ```>>> from kubernetes import config```
- ``` config.load_kube_config(config_file="{PATH_TO_THE_KUBECONFIG_FILE")```
- Error showing up: ```ERROR:root:[WinError 2] The system cannot find the file specified```

Kindly check the code link, I think that's the root cause. Verified that adding ```shell=True``` in [here](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/config/exec_provider.py#L71-L78) make Kubernetes Python client work well to load kubeconfig
 in Windows. And did we handle similar things for calling ```subprocess.Popen``` in Windows before?

**Environment**:
- Kubernetes version (`kubectl version`): Directly using Kubernetes Python client with Version: 30.1.0
- OS (e.g., MacOS 10.13.6): Microsoft Windows [Version 10.0.20348.2655]
- Python version (`python --version`) Python 3.11.9
- Python client version (`pip list | grep kubernetes`): 30.1.0

PranoyD:
What type of PR is this?
/kind bug

What this PR does / why we need it:
Which issue(s) this PR fixes:
Fixes #2276

Special notes for your reviewer:
NONE
Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

### DIFF
--- a/kubernetes/base/config/exec_provider.py
+++ b/kubernetes/base/config/exec_provider.py
@@ -75,7 +75,8 @@ class ExecProvider(object):
             stdin=sys.stdin if is_interactive else None,
             cwd=self.cwd,
             env=self.env,
-            universal_newlines=True)
+            universal_newlines=True,
+            shell=True)
         (stdout, stderr) = process.communicate()
         exit_code = process.wait()
         if exit_code != 0:


yliaog:
please sign cla

Ark-kun:
I don't think you need `shell=True` on Windows. What you need is to find a proper path for the program that needs to be executed.

PR #2294, opened 2024-10-08
google-auth Made as optional dependency

shivanand007:
#### What type of PR is this?
/kind feature

Makes Google auth as optional dependency, gives users a option to install it separately.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ DEVELOPMENT_STATUS = "3 - Alpha"
 # http://pypi.python.org/pypi/setuptools
 
 EXTRAS = {
-    'adal': ['adal>=1.0.2']
+    'adal': ['adal>=1.0.2'],
     'google-auth': ['google-auth>=1.0.0']
 }
 REQUIRES = []
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ DEVELOPMENT_STATUS = "3 - Alpha"
 
 EXTRAS = {
     'adal': ['adal>=1.0.2']
+    'google-auth': ['google-auth>=1.0.0']
 }
 REQUIRES = []
 with open('requirements.txt') as f:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -28,7 +28,8 @@
 # http://pypi.python.org/pypi/setuptools
 
 EXTRAS = {
-    'adal': ['adal>=1.0.2']
+    'adal': ['adal>=1.0.2'],
+    'google-auth': ['google-auth>=1.0.0']
```
roycaihw, line 32: Shall we specify `google-auth>=1.0.1`, to be consistent with the requirements.txt removal?

roycaihw (commented):
Please add a release note with instructions of how to install google-auth going forward.

We should also make sure the tests pass. I wonder if we need to update the test pipeline to install google-auth as well. 

/retest

setup.py
shivanand007: > Shall we specify `google-auth>=1.0.1`, to be consistent with the requirements.txt removal?

Yeah, That's fine we can make it consistent to version 1.0.1

shivanand007:
> Please add a release note with instructions of how to install google-auth going forward.
> I have added the required release note, kindly have a review for the same.

> We should also make sure the tests pass. I wonder if we need to update the test pipeline to install google-auth as well.
> Yes, As we have Removed google-auth from requirements.txt, This must be a dependency in code which makes test cases to fail, 
    I will check the code and see if we can make those code blocks optional as well, 
    Additionally, we have update the test pipeline to test this use case.
> /retest

setup.py
shivanand007: That change has been done

shivanand007:
@roycaihw 
I have made the necessary changes in code to make google-auth as optional, Requesting to Change Test Pipeline, So That we can pass all the E2E test Cases.

roycaihw:
Sorry for the late reply. 

/ok-to-test

kriszkern:
Hello!
What is the status of this PR?
This change would be very much appreciated to get rid of google-auth as a mandatory dependency, since many use cases of this client does not require it.

kriszkern:
Thanks @yliaog!
Cheers!

PR #2317, opened 2025-01-08
Add support for allowWatchBookmarks to the dynamic client

anvilpete:
#### What type of PR is this?
/kind feature

- There's currently no way to pass `allow_watch_bookmark` to the dynamic client's `watch` method. This PR adds this parameter.

- the `allowWatchBookmarks` query parameter is currently never set. This PR sets it based on the `allow_watch_bookmarks` keyword arg.

Fixes #

```release-note
Added support for allowWatchBookmarks to the dynamic client
```

```docs

```

### DIFF
--- a/examples/watch/watch_recovery.py
+++ b/examples/watch/watch_recovery.py
@@ -1,0 +1,78 @@
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


anvilpete:
There's some evidence in the commit/issue/PR history that people have been managing to enable watch bookmarks. But I'm not sure how this is possible without setting the `allowWatchBookmarks` query parameter.

But in any case, this PR makes it possible.

anvilpete:
@fabianvf any comments on this? It's a pretty trivial addition, and I can't see how bookmarks could work without it.

anvilpete:
@fabianvf, @yliaog is there any reason we can't merge this? The check failures seem unrelated to my changes.

yliaog:
do you mind to add an example on how it is used?

anvilpete:
The `allow_watch_bookmarks` parameter doesn't really affect the client behaviour, apart from allowing the server to send a BOOKMARK event. But I've added an example of using `resource_version` to recover from connection issues.

Is this what you had in mind?

yliaog:
thanks for the PR

/lgtm
/approve

PR #2335, opened 2025-01-26
Allow passing of custom pool manager

### LINKED ISSUE #2329: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed when using KUBECONFIG env in config.load_kube_config(config_file=
I do not have permission to reopen #1767 but the issue persists into 2025 and needs a fix or generic workaround.

Setting the config_file variable in the config.load_kube_config arguments to point to a kube config different than the default ~/.kube/config breaks with
```
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='XXX', port=16443): Max retries exceeded with url: /api/v1/nodes (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)')))
```
(with host variable replaced by XXX) while kubectl can use the file just fine via KUBECONFIG env var.

kubernetes-client/python should be able to use a different config than the default one.

have a KUBECONFIG environment variable set to a kube config file *different* than the default ~/.kube/config file
```
>>> import kubernetes
>>> import os
>>> os.path.exists(os.environ["KUBECONFIG"])
True
>>> kubernetes.config.load_kube_config(os.environ["KUBECONFIG"])
>>> v1 = kubernetes.client.CoreV1Api()
>>> v1.list_node()
```
This throws the SSLCertVerificationError error shown above.

Original ticket is #1767 

**Environment**:
- Kubernetes version (`kubectl version`): Client Version: v1.31.5 Kustomize Version: v5.4.2 Server Version: v1.30.8
- OS (e.g., MacOS 10.13.6): Ubuntu 22.04.5 LTS
- Python version (`python --version`) Python 3.12.3
- Python client version (`pip list | grep kubernetes`) kubernetes         31.0.0

rossigee:
#### What type of PR is this?

/kind feature

(although from some perspectives it could be considered a bug!)

Summary: Then using the `load_kube_config` method, the python library is (apparently?!) unable to connect *securely* to a cluster where the cluster API endpoints are secured with CA certificates that have been signed by an internal/external CA authority (i.e. not self-signed).

Why is this? It seems that only the cluster CA gets stored in the `~/.kube/config` file, not the whole certificate chain. Therefore, the library can only resolve the chain if it is self-contained (self-signed). If it depends on additional certificates it will fail an exception something like this...

```
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='k8s-api.cluster.xyz', port=6443): Max retries exceeded with url: /api/v1/namespaces/vm-nodes/serviceaccounts (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get issuer certificate (_ssl.c:1000)')))
```

Issue: Currently, there does not seem to be any way to control the way the REST client `pool_manager` is created, and it's default configuration makes it impossible (AFAICS?!) to correctly/cleanly specify the CA chain to use.

Fixes #2329

N/A

```release-note
Adds the ability to specify a custom CA chain for cases where the Kubernetes API is using externally-signed certificates.
```

```docs
If you are using `load_kube_config` with a cluster that requires a certificate chain, you can supply a custom PoolManager that like this:

\`\`\`
    kubernetes.config.load_kube_config()
    custom_pool_manager = urllib3.PoolManager(
        cert_reqs="CERT_REQUIRED",
        ca_certs="/etc/ssl/certs/ca-certificates.crt"
    )
    api_client = kubernetes.client.ApiClient(pool_manager=custom_pool_manager)
    v1 = kubernetes.client.CoreV1Api(api_client)
\`\`\`

```

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -66,13 +66,13 @@ class ApiClient(object):
     _pool = None
 
     def __init__(self, configuration=None, header_name=None, header_value=None,
-                 cookie=None, pool_threads=1):
+                 cookie=None, pool_threads=1, pool_manager=None):
         if configuration is None:
             configuration = Configuration.get_default_copy()
         self.configuration = configuration
         self.pool_threads = pool_threads
 
-        self.rest_client = rest.RESTClientObject(configuration)
+        self.rest_client = rest.RESTClientObject(configuration, pool_manager=pool_manager)
         self.default_headers = {}
         if header_name is not None:
             self.default_headers[header_name] = header_value
--- a/kubernetes/client/rest.py
+++ b/kubernetes/client/rest.py
@@ -50,12 +50,17 @@ class RESTResponse(io.IOBase):
 
 class RESTClientObject(object):
 
-    def __init__(self, configuration, pools_size=4, maxsize=None):
+    def __init__(self, configuration, pools_size=4, maxsize=None, pool_manager=None):
         # urllib3.PoolManager will pass all kw parameters to connectionpool
         # https://github.com/shazow/urllib3/blob/f9409436f83aeb79fbaf090181cd81b784f1b8ce/urllib3/poolmanager.py#L75  # noqa: E501
         # https://github.com/shazow/urllib3/blob/f9409436f83aeb79fbaf090181cd81b784f1b8ce/urllib3/connectionpool.py#L680  # noqa: E501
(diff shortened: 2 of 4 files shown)


rossigee:
Pls ignore. It wasn't clear from the docs, but I've (eventually!!) figured out how to do this.

PR #2336, opened 2025-01-27
fix: exec_provider.py

### LINKED ISSUE #2136: `ExecProvider` bug when accessing `stderr` before `None` type check
When authentication method is `exec` and the provider is `aws`, the command can sometimes return non-zero status without `stderr`. Below is an example of the output ot `config.load_kube_config()` in this case:

```
The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile.
ERROR:root:'NoneType' object has no attribute 'strip'
```

in this case the command returns a `255` return code with `stderr: None`. This causes `ConfigException` to not be raised.

Proposing a change to `kubernetes/base/config/exec_provider.py` as such:

```
--- a/kubernetes/base/config/exec_provider.py
+++ b/kubernetes/base/config/exec_provider.py
@@ -80,8 +80,8 @@ class ExecProvider(object):
         exit_code = process.wait()
         if exit_code != 0:
             msg = 'exec: process returned %d' % exit_code
-            stderr = stderr.strip()
             if stderr:
+                stderr = stderr.strip()
                 msg += '. %s' % stderr
             raise ConfigException(msg)
         try:
```

`ConfigException` should be raised if exec fails

Use exec auth configuration with non-valid SSO creds.

**Environment**:
- Kubernetes version (`kubectl version`): v1.28.3
- OS (e.g., MacOS 10.13.6): macOS 13.6
- Python version (`python --version`): 3.11.6
- Python client version (`pip list | grep kubernetes`): 28.1.0

calebAtIspot:
see https://github.com/kubernetes-client/python/issues/2136

#### What type of PR is this?

/kind bug

Fixes #2136

```release-note
Yes, but only in the error message
```

```docs

```

### DIFF
--- a/kubernetes/base/config/exec_provider.py
+++ b/kubernetes/base/config/exec_provider.py
@@ -87,8 +87,8 @@ class ExecProvider(object):
         exit_code = process.wait()
         if exit_code != 0:
             msg = 'exec: process returned %d' % exit_code
-            stderr = stderr.strip()
             if stderr:
+                stderr = stderr.strip()
                 msg += '. %s' % stderr
             raise ConfigException(msg)
         try:


yliaog:
please sign cla

PR #2369, opened 2025-03-07
feat: add switch control when watching events whether deserialization…

Kevinz857:
#### What type of PR is this?
/kind feature
/kind optimization

This PR adds an option to disable deserialization in the Watch.stream() method to improve performance for high-frequency watch operations. When dealing with large-scale Kubernetes clusters or high-frequency resource updates, the default deserialization of every event can cause significant latency overhead and potential processing delays.

By allowing clients to opt-out of automatic deserialization when only basic JSON parsing is needed, we can significantly reduce time cost and improve event processing throughput. This is particularly important in scenarios with high event volumes or resource constraints.

Key changes:
- Added a `deserialize` parameter to `Watch.stream()` method (defaults to True for backward compatibility)
- When `deserialize=False`, events are only JSON parsed without model conversion
- Maintains the original behavior when `deserialize=True`
- Reduces CPU overhead for use cases that don't require full object models

Fixes #

The change is backward compatible and doesn't affect existing code. Users can opt-in to the performance optimization by explicitly setting `deserialize=False` when they don't need the full model objects.

Performance improvement example:


Kevinz857:
@fabianvf @roycaihw pls take a review when u r free , thanks a  lot

kubernetes/base/watch/watch.py
```diff
@@ -172,14 +172,19 @@ def stream(self, func, *args, **kwargs):
...
         while True:
             resp = func(*args, **kwargs)
             try:
                 for line in iter_resp_lines(resp):
                     # unmarshal when we are receiving events from watch,
                     # return raw string when we are streaming log
                     if watch_arg == "watch":
-                        event = self.unmarshal_event(line, return_type)
+                        if deserialize:
+                            event = self.unmarshal_event(line, return_type)
```
roycaihw, line 184: We actually have the [return_type flag](https://github.com/kubernetes-client/python/blob/d80165da21a9fec1d5abc785cb446ab16f6affdb/kubernetes/base/watch/watch.py#L87) which could serve the purpose, but it looks pretty unintuitive-- IIUC `_raw_return_type` can be [persisted](https://github.com/kubernetes-client/python/blob/d80165da21a9fec1d5abc785cb446ab16f6affdb/kubernetes/base/watch/watch.py#L87) by setting `return_type` to `False`, which would result in deserialization [being skipped](https://github.com/kubernetes-client/python/blob/d80165da21a9fec1d5abc785cb446ab16f6affdb/kubernetes/base/watch/watch.py#L110-L116)

That being said, the flag name seems unintuitive and I think `deserialize = False` makes more sense.

cc @yliaog

kubernetes/base/watch/watch.py
```diff
@@ -172,14 +172,19 @@ def stream(self, func, *args, **kwargs):
...
+                        else:
+                            # Only do basic JSON parsing, no deserialize
+                            event = json.loads(line)
```
roycaihw, line 187: I also noticed that in this way, the resource version is not tracked like [here does](https://github.com/kubernetes-client/python/blob/d80165da21a9fec1d5abc785cb446ab16f6affdb/kubernetes/base/watch/watch.py#L119-L126).

I think we can refactor the unmarshal_event method to keep track of the resource version even when we skip deserialization.

roycaihw, line 187: Please update the tests to ensure the [retry](https://github.com/kubernetes-client/python/blob/d80165da21a9fec1d5abc785cb446ab16f6affdb/kubernetes/base/watch/watch_test.py#L439-L464) and [resource version](https://github.com/kubernetes-client/python/blob/d80165da21a9fec1d5abc785cb446ab16f6affdb/kubernetes/base/watch/watch_test.py#L221) behaviors when `deserialize = False`

kubernetes/base/watch/watch.py
yliaog: +1 on deserialize = False

Kevinz857:
Thank's ur comments , Setting the deserialize default value to false is actually more appropriate, adjust this and add test code commit . Pls take a look agin , thanks @roycaihw @yliaog

PR #2374, opened 2025-03-17
corrected bad quotes in CoreV1Api.md

Hellfire01:
In the readme a few quotes where written as \&quot; and interpreted as such rather than the intended &#39;
This causes them to be displayed incorrectly.
Note that the issue is also possibly in other files, I have only checked this one

#### What type of PR is this?
/kind documentation

Simply improves readability of the README document

There are possibly more issues like this one, especially if a tool is used for the generation of the documentation and is not configured correctly

No


yliaog:
the file is automatically generated, please update the upstream, or the generator tool.

Hellfire01:
> the file is automatically generated, please update the upstream, or the generator tool.

Where can I find the original document used for the generator ? That I may see if I can correct / commit there

Issue #2075, opened 2023-06-12
leaderelection do not stop leading properly


ealyn:
**What happened (please include outputs or screenshots)**:
After `onstarted_leading` function throw exception and finish, the process still leading.

**What you expected to happen**:
The process stop leading immediatly.

**How to reproduce it (as minimally and precisely as possible)**:
Just follow the `leaderelection/example.py`.

**Anything else we need to know?**:
Go client use `ctx.Done()` to find out the OnStartedLeading finish, and then release lock:
https://github.com/kubernetes/client-go/blob/1309f64d6648411b4a36a2f7fa84dd8df31884b6/tools/leaderelection/leaderelection.go#L265-L291

**Environment**:
- Kubernetes version (`kubectl version`): v1.20.15
- OS (e.g., MacOS 10.13.6): 10.15.7
- Python version (`python --version`): 3.8.10
- Python client version (`pip list | grep kubernetes`): 21.7.0

PR #2379, opened 2025-03-30
fix: Force Lease Expiration When Leader Exits

RaghavRoy145:
#### What type of PR is this?
/kind bug

Currently, when the leader exits (say, after receiving a `SIGINT`), the workers need to wait for its lease to expire before a leader is re-elected. This patch mimics the behaviour of the Go Client implementation of using `ctx.Done()`: https://github.com/kubernetes/client-go/blob/1309f64d6648411b4a36a2f7fa84dd8df31884b6/tools/leaderelection/leaderelection.go#L265-L291. It captures the `SIGINT` and forces the lease to exit by setting the expiration to a date in the past, and it also sets the `acquire_time` to None to force a leader election.

- #### Issue Reproduction
    As mentioned in the issue: https://github.com/kubernetes-client/python/issues/2075, to reproduce this issue you can follow leaderelection/example.py. Run it on 2-3 nodes (or tmux screens), and once a leader is elected hit Ctrl+C to force the leader to exit. The workers then wait for the leader's lease to expire before a new leader is elected.

- #### Expected behavior
    The leader exiting should trigger a leader election without having the workers wait for the lease to expire.

Fixes https://github.com/kubernetes-client/python/issues/2075

This is still not a complete fix, and it definitely hacky at the moment, and I would love any guidance here. Currently, the patch only handles `SIGINT`, but a leader may exit for various reasons, and there should be a more elegant way of handling this, probably using the thread context, but I was not able to figure that out. Further, the implementation of the `forceLeaseExpiration()` function is not elegant, you shouldn't need to set acquire_time to None and setting expiration to false is also a code smell in my opinion. This patch is a proof of concept because of this.

I also had to change the imports to point to my definitions of `electionconfig.py` and `leaderelectionrecord.py `for this to work, and I'm sure there is a better way of handling this. 

If its more sensible to mark this PR a draft, I'm happy to do so!

```release-note
None
```

```docs
None
```

### DIFF
--- a/kubernetes/base/leaderelection/electionconfig.py
+++ b/kubernetes/base/leaderelection/electionconfig.py
@@ -16,10 +16,10 @@ import sys
 import logging
 logging.basicConfig(level=logging.INFO)
 
-
 class Config:
+
     # Validate config, exit if an error is detected
-    def __init__(self, lock, lease_duration, renew_deadline, retry_period, onstarted_leading, onstopped_leading):
+    def __init__(self, lock, lease_duration, renew_deadline, retry_period, onstarted_leading, onstopped_leading, context):
         self.jitter_factor = 1.2
 
         if lock is None:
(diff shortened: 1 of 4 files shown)


RaghavRoy145:
> /assign @yliaog

Oops, I was supposed to do that after the reviews 🙃

PR #2407, opened 2025-05-31
feat: Make resource lookup case-insensitive to match Kubernetes API behavior

### LINKED ISSUE #2402: [discovery] Resource lookup is case-sensitive while it shouldn't

The following code fails with a resource not found error.
```python
kubernetes.config.load_kube_config()
with kubernetes.client.ApiClient() as client:
  v1 = kubernetes.dynamic.DynamicClient(client)
  rsrc = v1.resources.get(kind='service')
```

That it returns the `Service` resource, just like Kubernetes API and kubectl does.

Run the above snippet

Kevinz857:
# Make Resource Lookup Case-Insensitive to Match Kubernetes API Behavior

## What type of PR is this?
/kind feature
/kind bug

## What this PR does / why we need it
This PR implements case-insensitive resource lookup in the dynamic client to match the actual behavior of the Kubernetes API. Currently, when using the dynamic client to look up resources, the search is case-sensitive, which causes issues when users attempt to look up resources with different casing than what's defined in the API server.

The Kubernetes API itself is case-insensitive when it comes to resource kinds. For example, both `kubectl get pod` and `kubectl get Pod` work correctly. This PR makes the Python client's behavior consistent with the API server's behavior.

## Key changes
- Added a `_find_resource_case_insensitive` method to the `Discoverer` class
- Modified the `search` methods in both `LazyDiscoverer` and `EagerDiscoverer` classes to:
  - First try exact case matching (original behavior)
  - If no match is found and a kind is specified, attempt case-insensitive lookup
  - On failure, refresh the cache and retry both strategies
- Added test cases to verify case-insensitive lookup functionality

## Special notes for your reviewer
This change is backward compatible with existing code as it maintains the same behavior for exact case matches. It only adds the additional capability to find resources when the case doesn't exactly match.

## Does this PR introduce a user-facing change?
```release-note
Make resource lookup case-insensitive in the dynamic client to match Kubernetes API behavior
```

## Which issue(s) this PR fixes:
Fixes #2402

### DIFF
--- a/kubernetes/base/dynamic/discovery.py
+++ b/kubernetes/base/dynamic/discovery.py
@@ -195,6 +195,27 @@ class Discoverer(object):
             resources[resource_list.kind].append(resource_list)
         return resources
 
+    def _find_resource_case_insensitive(self, kind, resources_dict):
+        """
+        Find a resource in the resources_dict where the key matches the specified kind,
+        regardless of case.
+        
+        Args:
+            kind: The kind to search for (case-insensitive)
+            resources_dict: The resources dictionary to search in
+            
+        Returns:
+            The actual key if found, None otherwise
+        """
+        if not kind:
+            return None
+            
+        kind_lower = kind.lower()
+        for key in resources_dict.keys():
+            if key.lower() == kind_lower:
+                return key
+        return None
+    
     def get(self, **kwargs):
         """ Same as search, but will throw an error if there are multiple or no
             results. If there are multiple results and only one is an exact match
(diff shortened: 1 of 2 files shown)


roycaihw:
> The Kubernetes API itself is case-insensitive when it comes to resource kinds. For example, both kubectl get pod and kubectl get Pod work correctly. This PR makes the Python client's behavior consistent with the API server's behavior.

I think the case-insensitive behavior comes from kubectl. Kubernetes API server is case sensitive. One can verify that by running kubectl with --v=9 to check the actual API request. 

IIRC other clients like client-go are also case-sensitive. 

/priority awaiting-more-evidence

Issue #2414, opened 2025-06-12
stream api will lose data when use sslsocket


Novelfor:
**What happened (please include outputs or screenshots)**:
I use stream api to implement "kubectl exec", when i test my code, sometimes it will lose data. After debuging, i think the reason is kubernete client handle non blocking sslsocket error.

According to the Python documentation on SSL Sockets (https://docs.python.org/3/library/ssl.html#ssl-nonblocking), "SSL socket may still have data available for reading without [select()](https://docs.python.org/3/library/select.html#select.select) being aware of it".

Current code: 
https://github.com/kubernetes-client/python/blob/3e6cc5871997cd52fd8f4c8324894508f3a75136/kubernetes/base/stream/ws_client.py#L168

Can FIX it use sock.pending function
``` Python
    def update(self, timeout=0):
        """Update channel buffers with at most one complete frame of input."""
        if not self.is_open():
            return
        if not self.sock.connected:
            self._connected = False
            return

        # The options here are:
        # select.select() - this will work on most OS, however, it has a
        #                   limitation of only able to read fd numbers up to 1024.
        #                   i.e. does not scale well. This was the original
        #                   implementation.
        # select.poll()   - this will work on most unix based OS, but not as
        #                   efficient as epoll. Will work for fd numbers above 1024.
        # select.epoll()  - newest and most efficient way of polling.
        #                   However, only works on linux.
        ssl_pending = 0
        if self.sock.is_ssl():
            ssl_pending = self.sock.sock.pending()

        if hasattr(select, "poll"):
            poll = select.poll()
            poll.register(self.sock.sock, select.POLLIN)
            if timeout is not None:
                timeout *= 1_000  # poll method uses milliseconds as the time unit
            r = poll.poll(timeout)
            poll.unregister(self.sock.sock)
        else:
            r, _, _ = select.select(
                (self.sock.sock, ), (), (), timeout)

        if r or ssl_pending > 0:
```

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
Simple test to exec vim, it will lose cursor easily.. but should write a simple tty, here is my code. vim\htop can easily reproduce the data lose.
```
from kubernetes.stream import ws_client
import termios
import pty
import fcntl
import struct
import json
import signal
import tty
import select
import os
import threading
import yaml
import sys
import time
from datetime import datetime, timezone

def append_file(data):
    with open("test.txt", "a") as f:
[...]

PR #2422, opened 2025-07-22
fix stream bug when use ssl

Novelfor:
#### What type of PR is this?

/kind bug

Stream api will block data when use ssl socket.  

Fixes https://github.com/kubernetes-client/python/issues/2414

According to this document https://docs.python.org/3/library/ssl.html#ssl-nonblocking, select() may lose some data, it only happen when use  ssl socket

```release-note
NONE
```

```docs

```


ivaradi:
I think this could be improved by calling poll/select only when ssl_pending is 0.

kubernetes/base/stream/ws_client.py
```diff
@@ -193,7 +197,7 @@ def update(self, timeout=0):
             r, _, _ = select.select(
                 (self.sock.sock, ), (), (), timeout)
 
-        if r:
+        if r or ssl_pending > 0:
```
roycaihw, line 200: Could you add a test?

Issue #2397, opened 2025-05-26
Kubernetes Python Client Returns "Running" Status for Terminating Pods Instead of "Terminating"


RichwithPoor:
When using list_namespaced_pod() or list_pod_for_all_namespaces() from the Python Kubernetes client, pods that are in the process of being deleted (showing "Terminating" status in kubectl) continue to report their status as "Running" until they completely disappear from the cluster.

What happened:
When a pod is being deleted (kubectl delete pod <name>)
kubectl get pods correctly shows status "Terminating"
However, both list_namespaced_pod() and list_pod_for_all_namespaces() keep reporting the status as "Running"
This persists until the pod is completely removed from the cluster

What you expected to happen:
The Python client should return "Terminating" status when a pod is in the deletion process, matching kubectl behavior

**Environment**:
- Kubernetes version (`kubectl version`):  v1.31.0
- OS (e.g., MacOS 10.13.6): Macbook 14.5 (23F79)
- Python version (`python --version`) Python 3.12.2
- Python client version (`pip list | grep kubernetes`) 32.0.1

roycaihw:
The response should come from the api-server. This client doesn't cache the status.

I wonder if this could be caused by using resource version when listing the pods, and the response in from a cached version in the api-server. Could you share more on how you make the call and how to reproduce the issue?

RichwithPoor:
v1 = client.CoreV1Api()
pods = v1.list_namespaced_pod(namespace, label_selector=label_selector)
        pod_status_list = [{
            "name": pod.metadata.name,
            "ip": pod.status.pod_ip,
            "status": pod.status.phase,
            "container_name": pod.spec.containers[0].name if pod.spec.containers else "None"
        } for pod in pods.items]

This is currently the case. When the pod enters the terminal, the fetch is still Running

PR #2425, opened 2025-07-26
Kubernetes Python Client to return "Terminating" for terminating pods

p172913:
What type of PR is this?
/kind bug

What this PR does / why we need it:
This PR makes Kubernetes Python Client to return "Terminating" for terminating pods

Which issue(s) this PR fixes:
Fixes https://github.com/kubernetes-client/python/issues/2397

Special notes for your reviewer:
None

None

Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
None


kubernetes/client/api_client.py
```diff
@@ -261,7 +261,14 @@ def deserialize(self, response, response_type):
         except ValueError:
             data = response.data
 
-        return self.__deserialize(data, response_type)
+        result = self.__deserialize(data, response_type)
+        if response_type == "V1PodList":
+            for pod in result.items:
+                if pod.metadata and pod.metadata.deletion_timestamp:
+                    pod.status.phase = "Terminating"
```
roycaihw, line 268: This seems to be more of a request to change the upstream API behavior. You may want to check with upstream Kubernetes on the behavior here. 

IIUC you're expecting the pod status to be marked `Terminating` if the deletion timestamp is set. What did you see before this change? 

Also this is generated code, so we should not change it directly.

p172913: @roycaihw I’ve updated the PR so that the generated code is not changed— I moved the logic into a helper file(kubernetes\client\helpers\patch.py) and added a script(scripts\insert_patch_import.sh) to handle the behavior. It’s working as expected now. And made changes to scripts\insert_proxy_config.sh there is error in path. Please review my changes...

PR #2432, opened 2025-08-11
fix: set status default to "True" in V1PodFailurePolicyOnPodConditionsPattern __init__

### LINKED ISSUE #2430: Documentation on V1PodFailurePolicyOnPodConditionsPattern - status defaulting to true is misleading
https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1PodFailurePolicyOnPodConditionsPattern.md

The documentation states that the V1PodFailurePolicyOnPodConditionsPattern `status` param defaults to `True`, but there is validation that causes a V1PodFailurePolicyOnPodConditionsPattern with no defined `status` to fail since it defaults to `None`.

I had this for example in my code:
```python
    ignore_disruption_rule = kubernetes.client.V1PodFailurePolicyRule(
        action="Ignore",
        on_pod_conditions=[
            kubernetes.client.V1PodFailurePolicyOnPodConditionsPattern(
                type="DisruptionTarget",
            )
        ],
    )
```

but I got this error:

```JSON
{"levelname": "ERROR", "message": "Invalid value type: Invalid value for `status`, must not be `None`",
```

I fixed it by adding the status="True" like so:

```python
    ignore_disruption_rule = kubernetes.client.V1PodFailurePolicyRule(
        action="Ignore",
        on_pod_conditions=[
            kubernetes.client.V1PodFailurePolicyOnPodConditionsPattern(
                type="DisruptionTarget",
                status="True",
            )
        ],
    )
```

So I think we need to either make the default true or else remove that from the documentation and say it is a required parameter

mayooot:
#### What type of PR is this?
/kind feature

Fixes #2430

### DIFF
--- a/kubernetes/client/models/v1_pod_failure_policy_on_pod_conditions_pattern.py
+++ b/kubernetes/client/models/v1_pod_failure_policy_on_pod_conditions_pattern.py
@@ -42,7 +42,7 @@ class V1PodFailurePolicyOnPodConditionsPattern(object):
         'type': 'type'
     }
 
-    def __init__(self, status=None, type=None, local_vars_configuration=None):  # noqa: E501
+    def __init__(self, status="True", type=None, local_vars_configuration=None):  # noqa: E501
         """V1PodFailurePolicyOnPodConditionsPattern - a model defined in OpenAPI"""  # noqa: E501
         if local_vars_configuration is None:
             local_vars_configuration = Configuration()


yliaog:
the file is generated automatically, please make changes in the upstream, or the generator.

mayooot:
> the file is generated automatically, please make changes in the upstream, or the generator.

Ok, I've opened a PR [here](https://github.com/kubernetes/kubernetes/pull/133479).

PR #2437, opened 2025-08-14
Fixed watch_test.py indentation error

yliaog:
Fixed watch_test.py indentation error

./kubernetes/base/watch/watch_test.py:580:2: E999 IndentationError: expected an indented block after 'if' statement on line 579 def test_watch_with_deserialize_param(self):
 ^
1     E999 IndentationError: expected an indented block after 'if' statement on line 579

#### What type of PR is this?
/kind bug

Fixes #

```release-note
NONE
```

```docs

```


roycaihw (commented):
I think there is an extra `if __name__ == '__main__':` that we should delete

kubernetes/base/watch/watch_test.py
```diff
@@ -577,43 +579,43 @@ def test_pod_log_empty_lines(self):
             self.api.delete_namespaced_pod.assert_called_once_with(name=pod_name, namespace=self.namespace)
 
 if __name__ == '__main__':
```
roycaihw, line 581: @yliaog I think we should remove this line. The indentation below doesn't have to change. Also note that there is another `if __name__ == '__main__': ` towards the end of this file-- that's the correct `if __name__ == '__main__':` that we should keep. WDYT?

yliaog: Yes, that's right.

PR #2451, opened 2025-09-21
Update kube_config.py

sejalP07:
This issue occurs because parse_rfc3339 in the Kubernetes Python client does not guard against re.search() returning None. When the input string is not a valid RFC3339 timestamp, calling .groups() on None raises an AttributeError, which in kube_config.py is logged only as 'NoneType' object has no attribute 'groups' without showing the problematic string. This makes debugging difficult and allows execution to continue with an invalid expiry value. The fix is to update parse_rfc3339 to raise a clear ValueError when parsing fails and to improve logging in kube_config.py so the invalid string is included in the error message. This will provide clearer errors, prevent silent misconfigurations, and make the client more robust.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/kubernetes/base/config/kube_config.py
+++ b/kubernetes/base/config/kube_config.py
@@ -510,14 +510,19 @@ class KubeConfigLoader(object):
                     base64_file_content=False,
                     temp_file_path=self._temp_file_path).as_file()
             else:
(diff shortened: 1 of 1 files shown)


roycaihw:
Please sign the CLA

Issue #2424, opened 2025-07-24
Vulnerability [CVE-2025-50182, CVE-2025-50181] is getting reported for package urllib3


sureshnikuliya1:
Vulnerability [https://github.com/advisories/GHSA-48p4-8xcf-vxj5, https://github.com/advisories/GHSA-pq67-6m6q-mj2v] is getting reported for the Python package urllib3 while installing the Kubernetes package version 33.1.0. Can you please provide the details on when we are releasing the latest version with this fix in it? Please provide the details of the Kubernetes version in which this has been fixed.

```
kubernetes-33.1.0-py2.py3-none-any.whl (Root Library)
❌ urllib3-2.4.0-py3-none-any.whl (Vulnerable Library)
```

smemsh:
#2458 appears to supersede this

hoerup:
Solved in https://github.com/kubernetes-client/python/pull/2461

sathieu:
https://github.com/kubernetes-client/python/pull/2482 for 34.x branch

bonastreyair:
I wonder when it will be released...

yurnov:
Released fix is 35.0.0 (2026-01-16) and later, including the current 36.0.3 (urllib3!=2.6.0,>=1.24.2 on PyPI). I think that this issue can be closed

PR #2461, opened 2025-10-14
feat: Remove the urllib3 pinning

ZPascal:
#### What type of PR is this?

/kind feature
/kind regression

##### Pull Request Description

The PR removes the version pinning for the urllib3 library.

###### Background  
Starting with `urllib3` version `2.4.0`, stricter certificate validity checks were introduced when running on Python 3.13+ ([urllib3#3571](https://github.com/urllib3/urllib3/pull/3571)).  
This change can trigger errors such as:

```
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Missing Authority Key Identifier
```

These errors occur only in environments where **Kubernetes clusters are configured with invalid or incomplete certificates**, most notably on **AWS EKS clusters created with Kubernetes v1.16 and earlier**.

##### Change  
We are **not pinning** the `urllib3` version. Instead, we keep the dependency flexible (optionally with a temporary upper bound if strictly necessary), ensuring compatibility with future versions of `urllib3`.

##### Rationale / Why not pin the version

- 📦 **Avoid unnecessary technical debt**  
  Pinning a widely used core dependency like `urllib3` can quickly lead to upgrade blockers and complicate maintenance down the line.

- 🔐 **Don’t block important security updates**  
  `urllib3` is a critical dependency for HTTPS communication. Pinning could delay or prevent applying security patches or other improvements.

- 🧹 **The issue only affects outdated infrastructure**  
  The certificate verification error occurs exclusively on clusters with invalid or outdated certificates. These environments should be updated rather than forcing the dependency to remain outdated.

- 🧭 **Fixing the root cause is better than downstream workarounds**  
  It’s more sustainable to address the infrastructure issue (fixing or rotating cluster certificates) than holding back modern library versions.

- ⚡ **Keeps upgrade paths open**  
  By not pinning the version, we can easily adopt future `urllib3` releases without major dependency refactors.

##### Recommendation for affected users  
If this error occurs, the correct fix is to update or properly configure the Kubernetes cluster certificates. A temporary upper bound may be set in local deployments if an immediate infrastructure fix is not feasible.

---

✅ **Summary:**  
Instead of pinning `urllib3` to a specific version, we keep the dependency flexible and address the underlying infrastructure issue.  
This approach is **more secure**, **easier to maintain**, and **aligned with dependency management best practices**.

```release-note
We no longer pin the `urllib3` version.  
`urllib3` 2.4.0 introduced stricter certificate checks on Python 3.13+, which may trigger: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Missing Authority Key Identifier

This only affects clusters (version <=1.16) with outdated or misconfigured certificates (e.g. old AWS EKS versions). Please update the Kubernetes cluster certificates.
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,5 +6,5 @@ pyyaml>=5.4.1  # MIT
 websocket-client>=0.32.0,!=0.40.0,!=0.41.*,!=0.42.* # LGPLv2+
 requests # Apache-2.0
 requests-oauthlib # ISC
-urllib3>=1.24.2,<2.4.0  # MIT
+urllib3>=1.24.2 # MIT
 durationpy>=0.7 # MIT


arielvb:
Hi,

Would it be possible to cut a new release after merging either this pull request or PR #2439? This would help users stay up-to-date with security patches and compatibility improvements in urllib3.

Other packages are starting to require urllib3 >= 2.5 and this is bringing some incompatibilities.

Thanks!

ZPascal:
Hi @roycaihw & @yliaog, can you please check the PR?

NadhiyaPM:
We need the release for this PR to fix this - https://github.com/kubernetes-client/python/issues/2424 .

sathieu:
@roycaihw & @yliaog Could you please make a release including this fix :pray: ?

roycaihw:
https://github.com/kubernetes-client/python/pull/2461#issuecomment-3595237183 It will be included in the upcoming v35 release, pending Kubernetes 1.35 cut.

jonaslb:
It would be appropriate to cut a v 34 bugfix with this as a backport. It was never appropriate to put an upper bound on urllib3 like that

joshzcold:
Just to hopefully help push for a new release. 
2 new CVE's came out since this was merged for the locked version of urlib3

CVE-2025-66471: https://nvd.nist.gov/vuln/detail/CVE-2025-66471
CVE-2025-66418: https://nvd.nist.gov/vuln/detail/CVE-2025-66418

yliaog:
i'm working on 1.35 release, it will be alpha/beta this week, and GA next week.

PR #2466, opened 2025-11-01
Support dict[...] syntax in ApiClient.__deserialize

### LINKED ISSUE #2463: Incorrect type declaration in `openapi_types` for `dict`
I created a Pydantic converter/adapter for the generated client models and used the `openapi_types` dict to infer the fields types (I wanted to include some of the fields in my Pydantic-based configuration). That works pretty well - one can use that type string directly for the Pydantic fields, with one exception - where the type is a literal `dict` of something (and not a forward reference to an existing model), because the dict type is declared as `dict(str, str)` ([see this example](https://github.com/kubernetes-client/python/blob/a486091a91fdd0d0d389bc9538533118ee6fa3c8/kubernetes/client/models/v1_label_selector.py#L36-L37)).

I believe this type declaration should follow the current python typing syntax, which is `dict[str, str]` (with square brackets). Interestingly enough, the `list` in the example already uses square brackets. This alone leaves me with the assumption this inconsistency is a bug.

I've just written a simple wrapper to convert the parentheses to square brackets, but I'm thinking this might as well get fixed downstream. I've also looked through how the code is generated, but I couldn't find it.

n/a

You're great! 🙂 

**Environment**:
- Kubernetes version (`kubectl version`): 1.34
- OS (e.g., MacOS 10.13.6): Fedora Linux 41
- Python version (`python --version`): 3.13

mohsinm-dev:
## Summary
- Adds support for modern `dict[str, str]` syntax in `ApiClient.__deserialize()` 
- Maintains backward compatibility with existing `dict(str, str)` syntax
- Enables Pydantic adapter compatibility for `openapi_types`

## Changes
- Added post-generation patch `scripts/api_client_dict_syntax.diff`
- Updated `scripts/update-client.sh` to apply patch during client generation
- Added comprehensive unit tests covering both syntaxes and nested dicts

## Test plan
- [x] Unit test `test_deserialize_dict_syntax_compatibility` passes
- [x] Legacy `dict(str, str)` syntax continues to work
- [x] New `dict[str, str]` syntax works correctly  
- [x] Nested `dict[str, dict[str, str]]` syntax works
- [x] No breaking changes to existing functionality

Fixes #2463

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -285,6 +285,25 @@ class ApiClient(object):
                 return {k: self.__deserialize(v, sub_kls)
                         for k, v in six.iteritems(data)}
 
+            if klass.startswith('dict[') and klass.endswith(']'):
+                # Parse dict[key_type, value_type] respecting nested brackets
+                inner = klass[len('dict['):-1]
+                bracket_depth = 0
+                comma_pos = -1
+                for i, char in enumerate(inner):
+                    if char in '([{':
+                        bracket_depth += 1
+                    elif char in ')]}':
+                        bracket_depth -= 1
+                    elif char == ',' and bracket_depth == 0:
+                        comma_pos = i
+                        break
+                
+                if comma_pos != -1:
+                    value_type = inner[comma_pos + 1:].strip()
+                    return {k: self.__deserialize(v, value_type)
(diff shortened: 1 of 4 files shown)


mohsinm-dev:
Closing to recreate from proper feature branch

PR #2467, opened 2025-11-01
Support dict[...] syntax in ApiClient.__deserialize

### LINKED ISSUE #2463: Incorrect type declaration in `openapi_types` for `dict`

mohsinm-dev:
## Summary
- Adds support for modern `dict[str, str]` syntax in `ApiClient.__deserialize()` 
- Maintains backward compatibility with existing `dict(str, str)` syntax
- Enables Pydantic adapter compatibility for `openapi_types`

## Changes
- Added post-generation patch `scripts/api_client_dict_syntax.diff`
- Updated `scripts/update-client.sh` to apply patch during client generation
- Added comprehensive unit tests covering both syntaxes and nested dicts

## Test plan
- [x] Unit test `test_deserialize_dict_syntax_compatibility` passes
- [x] Legacy `dict(str, str)` syntax continues to work
- [x] New `dict[str, str]` syntax works correctly  
- [x] Nested `dict[str, dict[str, str]]` syntax works
- [x] No breaking changes to existing functionality

Fixes #2463

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -285,6 +285,25 @@ class ApiClient(object):
                 return {k: self.__deserialize(v, sub_kls)
                         for k, v in six.iteritems(data)}
 
+            if klass.startswith('dict[') and klass.endswith(']'):
+                # Parse dict[key_type, value_type] respecting nested brackets
+                inner = klass[len('dict['):-1]
+                bracket_depth = 0
+                comma_pos = -1
+                for i, char in enumerate(inner):
+                    if char in '([{':
+                        bracket_depth += 1
+                    elif char in ')]}':
+                        bracket_depth -= 1
+                    elif char == ',' and bracket_depth == 0:
+                        comma_pos = i
+                        break
+                
+                if comma_pos != -1:
+                    value_type = inner[comma_pos + 1:].strip()
+                    return {k: self.__deserialize(v, value_type)
+                            for k, v in six.iteritems(data)}
+
             # convert str to class
             if klass in self.NATIVE_TYPES_MAPPING:
                 klass = self.NATIVE_TYPES_MAPPING[klass]
--- a/kubernetes/test/test_api_client.py
+++ b/kubernetes/test/test_api_client.py
@@ -25,6 +25,32 @@ class TestApiClient(unittest.TestCase):
         atexit._run_exitfuncs()
         self.assertIsNone(client._pool)
 
+    def test_deserialize_dict_syntax_compatibility(self):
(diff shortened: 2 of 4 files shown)


kubernetes/client/api_client.py
```diff
@@ -285,6 +285,25 @@ def __deserialize(self, data, klass):
...
+                    if char in '([{':
+                        bracket_depth += 1
+                    elif char in ')]}':
+                        bracket_depth -= 1
+                    elif char == ',' and bracket_depth == 0:
+                        comma_pos = i
+                        break
+                
+                if comma_pos != -1:
+                    value_type = inner[comma_pos + 1:].strip()
```
vit-zikmund: Why wasn't reusing the usual regex approach (used for `dict(`) satisfactory?

vit-zikmund (commented):
Hello Mohsin, Thanks a lot for picking up my issue!
I'd like to ask two questions. One is in the review comments and the other is: What's the change needed to be done on the openapi generator side? Do you know?

Thanks a lot!

PS: There are some failed automatic checks which add the `do-not-merge` labels. I'm not sure how the review process works in this repo, but it surely is understaffed (as usual) and a human reviewer won't IMHO pay attention to this PR until it's labeled `lgtm` (= all the automatic PR checks are resolved).

yliaog:
openapi generator has been upgraded to the laste v7.25.0+snapshot, i think the modern dict[...] syntax is supported now.

closing this PR

/close

Issue #2419, opened 2025-07-09
Support for Python 3.12 and 3.13


etiennnr:
**What is the feature and why do you need it**:
I see in the main README that the last supported version is 3.11 (![pypi supported versions](https://img.shields.io/pypi/pyversions/kubernetes.svg)).

Would it be possible / complicated to also support python 3.12 and 3.13? Or is it just that the pypi metadata has simply never been updated ?

**Describe the solution you'd like to see**:

AshwinNS:
Hi, Is this package compatible with python 3.12 ?

roycaihw:
We can add the support in the next release (1.34). cc @yliaog 
Ref required changes: https://github.com/kubernetes-client/python/pull/2029, https://github.com/kubernetes-client/python/pull/2003

etiennnr:
@roycaihw I see that release [1.34.1](https://github.com/kubernetes-client/python/releases/tag/v34.1.0) is already out, however, still no official support for python 3.12 or 3.13 or 3.14 (out since last week now)... 

Is this still in scope ?

durera:
Python 3.12 has been in the test config since the start of the year: https://github.com/kubernetes-client/python/commit/f7701ac020645c847f537553e28721c234789090

Other than updating https://github.com/kubernetes-client/python/blob/master/setup.py#L80-L86 (which list 3.6 to 3.11 as the supported versions, [perhaps time to remove 3.6 to 3.9 as they are all now out of support](https://devguide.python.org/versions/)) is there anything else that needs doing (e.g. anything that someone such as myself could volunteer to help with) to help get the client officially supporting 3.12?

etiennnr:
Hello @durera,

Thanks for the reply! Yeah, we started using it since I saw that there were already unit testing in place for 3.12 (in GHA). It's working pretty well, we didn't have any issues. Hence, I think we "just" need to add it to the officially supported list. I also agree to stop supporting 3.6-3.9, I guess maintainers can agree or not in the PR's review.

I think efforts should instead be put in supporting 3.13 and 3.14 IMHO.

PR #2475, opened 2025-12-01
Changing which python version is "officially" supported

etiennnr:
#### What type of PR is this?

/kind documentation
/kind deprecation

Removing Python 3.6, 3.7, 3.8, 3.9 and added 3.12 from `setup.py`, which will affect the "officially" supported version in the README.

Fixes https://github.com/kubernetes-client/python/issues/2419

```release-note
Removing Python 3.6, 3.7, 3.8, 3.9 and added 3.12 from the supported versions list
```

```docs

```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,6 +78,8 @@ setup(
         "Operating System :: OS Independent",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
         "Programming Language :: Python :: 3.12",


setup.py
```diff
@@ -78,11 +78,8 @@
         "Operating System :: OS Independent",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.6",
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
```
yliaog: we are running tests for 3.8, 3.9, 3.10, 3.11 and 3.12, i think you can remove 3.6 and 3.7 from the list

PR #2483, opened 2025-12-12
Fix: handle invalid RFC3339 expirationTimestamp in exec plugin output

11-Sarthak:
#### What type of PR is this?
/kind bug

This PR fixes a crash in `_load_from_exec_plugin()` where
`parse_rfc3339()` fails if the exec plugin returns an invalid
`expirationTimestamp`.

Changes:
- Wrap `parse_rfc3339` in a try-except block
- Log a clear error message if parsing fails
- Set `self.expiry = None` instead of crashing

Prevents Airflow KubernetesPodOperator from crashing with 
"'NoneType' object has no attribute 'groups'" error.

Fixes #

No user-facing changes. Only improves error handling.

No

```release-note
NONE
```

### DIFF
--- a/kubernetes/base/config/kube_config.py
+++ b/kubernetes/base/config/kube_config.py
@@ -530,11 +530,16 @@ class KubeConfigLoader(object):
                 logging.error('exec: missing token or clientCertificateData '
                               'field in plugin output')
                 return None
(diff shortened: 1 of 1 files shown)


yliaog:
it's probably safer, and better to have well defined expiry than None.

PR #2486, opened 2025-12-14
implement pod exec websockets v5

aojea:
/kind feature
/kind api-change
/kind deprecation

```release-note
support new exec v5 websocket subprotocol
```

```docs
- [KEP]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/4006-transition-spdy-to-websockets#proposal-new-remotecommand-sub-protocol-version---v5channelk8sio
```

cc: @yliaog @siyuanfoundation

### DIFF
--- a/kubernetes/base/stream/ws_client.py
+++ b/kubernetes/base/stream/ws_client.py
@@ -39,6 +39,10 @@ STDOUT_CHANNEL = 1
 STDERR_CHANNEL = 2
 ERROR_CHANNEL = 3
 RESIZE_CHANNEL = 4
+CLOSE_CHANNEL = 255
+
+V4_CHANNEL_PROTOCOL = "v4.channel.k8s.io"
+V5_CHANNEL_PROTOCOL = "v5.channel.k8s.io"
 
 class _IgnoredIO:
     def write(self, _x):
@@ -59,6 +63,8 @@ class WSClient:
         """
         self._connected = False
         self._channels = {}
+        self._closed_channels = set()
+        self.subprotocol = None
         self.binary = binary
         self.newline = '\n' if not self.binary else b'\n'
         if capture_all:
@@ -66,19 +72,24 @@ class WSClient:
         else:
             self._all = _IgnoredIO()
         self.sock = create_websocket(configuration, url, headers)
+        self.subprotocol = getattr(self.sock, 'subprotocol', None)
         self._connected = True
         self._returncode = None
 
     def peek_channel(self, channel, timeout=0):
         """Peek a channel and return part of the input,
         empty string otherwise."""
+        if channel in self._closed_channels and channel not in self._channels:
+            return b"" if self.binary else ""
         self.update(timeout=timeout)
         if channel in self._channels:
             return self._channels[channel]
-        return ""
+        return b"" if self.binary else ""
 
     def read_channel(self, channel, timeout=0):
         """Read data from a channel."""
+        if channel in self._closed_channels and channel not in self._channels:
+            return b"" if self.binary else ""
         if channel not in self._channels:
             ret = self.peek_channel(channel, timeout)
         else:
@@ -93,6 +104,7 @@ class WSClient:
             timeout = float("inf")
         start = time.time()
         while self.is_open() and time.time() - start < timeout:
+            # Always try to drain the channel first
             if channel in self._channels:
                 data = self._channels[channel]
                 if self.newline in data:
@@ -104,6 +116,14 @@ class WSClient:
                     else:
                         del self._channels[channel]
                     return ret
+
+            if channel in self._closed_channels:
+                if channel in self._channels:
+                    ret = self._channels[channel]
+                    del self._channels[channel]
+                    return ret
+                return b"" if self.binary else ""
+
             self.update(timeout=(timeout - time.time() + start))
 
     def write_channel(self, channel, data):
@@ -118,6 +138,14 @@ class WSClient:
 
         payload = channel_prefix + data
         self.sock.send(payload, opcode=opcode)
+
+    def close_channel(self, channel):
+        """Close a channel (v5 protocol only)."""
+        if self.subprotocol != V5_CHANNEL_PROTOCOL:
+            return
+        data = bytes([CLOSE_CHANNEL, channel])
+        self.sock.send(data, opcode=ABNF.OPCODE_BINARY)
+        self._closed_channels.add(channel)
 
     def peek_stdout(self, timeout=0):
         """Same as peek_channel with channel=1."""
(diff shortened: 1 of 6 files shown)


kubernetes/base/stream/ws_client.py
```diff
@@ -104,6 +111,14 @@ def readline_channel(self, channel, timeout=None):
                     else:
                         del self._channels[channel]
                     return ret
+
+            if channel in self._closed_channels:
```
seans3, line 127: Do we need this same short-circuiting code in `peek_channel` and `read_channel` ?

kubernetes/base/stream/ws_client.py
```diff
@@ -119,6 +134,14 @@ def write_channel(self, channel, data):
         payload = channel_prefix + data
         self.sock.send(payload, opcode=opcode)
 
+    def close_channel(self, channel):
+        """Close a channel (v5 protocol only)."""
+        if self.subprotocol != V5_CHANNEL_PROTOCOL:
+            return
+        data = bytes([255, channel])
```
seans3: Should we define a constant (e.g. CLOSE_CHANNEL = 255 or V5_HALF_CLOSE = 255) for this magic number, and reference the constant?

kubernetes/base/stream/ws_client_test.py
```diff
@@ -123,6 +126,117 @@ def test_websocket_proxycare(self):
...
+            mock_ws.subprotocol = V4_CHANNEL_PROTOCOL
+            mock_ws.connected = True
+            mock_create.return_value = mock_ws
+
+            client = WSClient(self.config_mock, "ws://test", headers=None, capture_all=True)
+            client.close_channel(0)
+
+            mock_ws.send.assert_not_called()
+
+    def test_update_receives_close_v5(self):
```
seans3, line 187: Should we add an additional unit test to verify how `readline_channel` handles a closed channel with leftover data?

While the current unit tests cover the parsing of the close signal itself, the new logic you added inside `readline_channel`—which flushes the remaining buffer even if it lacks a newline—is currently untested. Having a test that asserts `readline_channel` successfully flushes leftover data (e.g. "hello") and then returns an empty string on the subsequent call would ensure this specific edge-case logic is protected from future regressions.

kubernetes/e2e_test/test_client.py
```diff
@@ -201,6 +201,63 @@ def test_pod_apis(self):
...
+            self.assertEqual("test-close", line)
+
+            # Close stdin (channel 0)
+            # This should send EOF to cat, causing it to exit.
+            resp.close_channel(0)
+
+            # Wait for process to exit
+            resp.update(timeout=5)
+            start = time.time()
+            while resp.is_open() and time.time() - start < 10:
```
seans3: Should we simplify this polling loop by using resp.run_forever(timeout=15) instead of manually checking resp.is_open()? This would align it closely with how the other exec tests in this file wait for streams to close.

kubernetes/base/stream/ws_client_test.py
```diff
@@ -123,6 +126,117 @@ def test_websocket_proxycare(self):
...
+            patch('select.select') as mock_select:
+
+            mock_ws = MagicMock()
+            mock_ws.subprotocol = V5_CHANNEL_PROTOCOL
+            mock_ws.connected = True
+            mock_ws.sock.fileno.return_value = 10
+
+            # Setup frame with close signal for channel 0
+            frame = MagicMock()
+            frame.data = b'\xff\x00'
```
seans3: Would this be better as constants, such as `frame.data = bytes([CLOSE_CHANNEL, STDIN_CHANNEL])` or whatever the 255 constant is defined as.  I believe other places that we see `b'\xff\x00'` would be more readable with the constants.

seans3 (changes requested):
This looks good and substantially complete.

One high-level observation: I believe falling back to `v4.channel.k8s.io` is the right approach for backwards compatibility, but we should probably note that for clusters < v1.30, the client will silently revert to the broken behavior and hang on EOF. 

I've left a few minor inline comments mostly about tests and constants.

kubernetes/base/stream/ws_client.py
yliaog: +1 to have a constant

kubernetes/base/stream/ws_client.py
```diff
@@ -200,13 +223,24 @@ def update(self, timeout=0):
...
-                if len(data) > 1:
+                if len(data) > 0:
+                    # Parse channel from raw bytes to support v5 CLOSE signal AND avoid charset issues
                     channel = data[0]
-                    if six.PY3 and not self.binary:
-                        channel = ord(channel)
+                    # In Py3, iterating bytes gives int, but indexing bytes gives int.
+                    # websocket-client frame.data might be bytes.
+
+                    if channel == 255 and self.subprotocol == V5_CHANNEL_PROTOCOL: # v5 CLOSE
```
yliaog: better to replace 255 with a constant

kubernetes/base/stream/ws_client.py
aojea: ack

kubernetes/base/stream/ws_client.py
aojea: ack

kubernetes/base/stream/ws_client.py
aojea: done

kubernetes/base/stream/ws_client_test.py
aojea: added tests also for `read_channel` and `peek_channel` to validate channels are drained and follow the expected semantics

kubernetes/e2e_test/test_client.py
aojea: done

kubernetes/base/stream/ws_client_test.py
aojea: moved magic numbers to constants

aojea:
@seans3 @yliaog PTAL

kubernetes/base/stream/ws_client_test.py
aojea: added

seans3:
Looks great--thanks.

/lgtm

kubernetes/e2e_test/test_client.py
```diff
@@ -201,6 +201,60 @@ def test_pod_apis(self):
...
+        resp = stream(api.connect_post_namespaced_pod_exec, name, 'default',
+                      command=['/bin/sh', '-c', 'cat'],
+                      stderr=True, stdin=True,
+                      stdout=True, tty=False,
+                      _preload_content=False)
+
+        if resp.subprotocol != "v5.channel.k8s.io":
+            resp.close()
+            api.delete_namespaced_pod(name=name, body={}, namespace='default')
+            self.skipTest("Skipping test: v5.channel.k8s.io subprotocol not negotiated")
```
yliaog, line 237: the test is skipped in the e2e test flow: https://github.com/kubernetes-client/python/actions/runs/25740181177/job/75636044165

how to make it run?

yliaog: kubernetes/e2e_test/test_client.py::TestClient::test_pod_exec_close_channel SKIPPED

aojea: we need to get a more recent version of kind, let me update that

aojea: it was not the version, the e2e tests were not initializing the subprotocol so it never matched ... now the client if is not preconfigured, uses the subprotocol obtained during the negotiation

aojea:
@yliaog is it normal for the e2e tests to take so long? they take more than 3 hours, that sounds like a lot

aojea:
> 6a83c7e

it runs in less than 2 minutes in my vm https://github.com/kubernetes-client/python/pull/2486/commits/6a83c7e951ecb396e2afeef8d2aedcddf86214c1 , it seems there can be a deadlock on stop

yliaog:
the e2e takes a long time in the informer tests, i have not had a chance to look into why it takes so long.

aojea:
> the e2e takes a long time in the informer tests, i have not had a chance to look into why it takes so long.

Fixed im the latest commit, there was deadlockimg when.closing, added a comment

yliaog:
thanks for the fix @aojea 

thanks for the review @seans3 

/lgtm
/approve

Issue #2477, opened 2025-12-05
Kubernetes Clients breaks with urllib3 2.6.0


potiuk:
**What happened (please include outputs or screenshots)**:

When urllib3 2.6.0 is used with kubernetes client, it fails because internally it uses "getheaders()" which have been [deprecated in 2.0.0](https://github.com/urllib3/urllib3/blob/main/CHANGES.rst#deprecated) and [removed in 2.6.0](https://github.com/urllib3/urllib3/blob/main/CHANGES.rst#removals) .

The offending line is:

https://github.com/kubernetes-client/python/blob/a49d85dbc593e083974099e766809c7b13b243de/kubernetes/client/exceptions.py#L91

Possibly there are other pleaces.

When there is an API exception returned and urllib 2.6.0 you get an exception:

```python
../.venv/lib/python3.10/site-packages/kubernetes/client/api/core_v1_api.py:23999: in read_namespaced_pod
    return self.read_namespaced_pod_with_http_info(name, namespace, **kwargs)  # noqa: E501
../.venv/lib/python3.10/site-packages/kubernetes/client/api/core_v1_api.py:24086: in read_namespaced_pod_with_http_info
    return self.api_client.call_api(
../.venv/lib/python3.10/site-packages/kubernetes/client/api_client.py:348: in call_api
    return self.__call_api(resource_path, method,
../.venv/lib/python3.10/site-packages/kubernetes/client/api_client.py:180: in __call_api
    response_data = self.request(
../.venv/lib/python3.10/site-packages/kubernetes/client/api_client.py:373: in request
    return self.rest_client.GET(url,
../.venv/lib/python3.10/site-packages/kubernetes/client/rest.py:244: in GET
    return self.request("GET", url,
../.venv/lib/python3.10/site-packages/kubernetes/client/rest.py:238: in request
    raise ApiException(http_resp=r)
../.venv/lib/python3.10/site-packages/kubernetes/client/exceptions.py:91: in __init__
    self.headers = http_resp.getheaders()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <kubernetes.client.rest.RESTResponse object at 0x7f0b51d92c50>

    def getheaders(self):
        """Returns a dictionary of the response headers."""
>       return self.urllib3_response.getheaders()
[...]

PR #2491, opened 2025-12-30
feat: Avoid urllib3 2.6.0

sathieu:
#### What type of PR is this?

/kind bug

See https://github.com/kubernetes-client/python/issues/2477

(not marking as fixing, because this will come again in a future version)

Fixes #

```release-note
Avoid urllib3 2.6.0 (breaks because of missing getheaders method)
```

```docs

```


yliaog:
/lgtm
/approve

thanks for the PR

PR #2479, opened 2025-12-08
Fix: correect header retrieval  for urllib3 compatibility

franck-ada:
What type of PR is this?
/kind bug
/kind deprecation

What this PR does / why we need it:
Correct issue link to upgrade of urlib3

Which issue(s) this PR fixes:
Fixes #2280 / #2477

Special notes for your reviewer:
Does this PR introduce a user-facing change?
NONE

PR #2497, opened 2026-01-08
Remove use of deprecated urllib3 functionality

stephenfin:
#### What type of PR is this?

/kind feature

Remove use of deprecated urllib3 functionality. As noted in the commit message, this functionality was briefly removed in 2.6.0 but later re-added in 2.6.1. Ultimately, we should fix this by bumping the codegenerator version used here but that's a much larger change that should be tackled separately.

/assign @yliaog

Fixes #

This is an alternative to #2479.

```release-note
NONE
```

(Technically we've deprecated a method but we won't remove this).

```docs

```

### DIFF
--- a/kubernetes/client/api_client.py
+++ b/kubernetes/client/api_client.py
@@ -196,8 +196,7 @@ class ApiClient(object):
         if _return_http_data_only:
             return (return_data)
         else:
-            return (return_data, response_data.status,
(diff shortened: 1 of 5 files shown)


stephenfin:
/close

I think we'd be better off working bumping the `openapi-generator` version than layering on even more patches to this. #2498 is a move in this direction at least.

PR #2513, opened 2026-02-13
Apply proxy config after hotfixes

### LINKED ISSUE #2460: NO_PROXY/no_proxy env vars not working as expected, as of 34.1.0
Upgrading libraries in one of my images. Last image was built 4 days ago, works fine. We were running kubernetes 33.1.0.
Today's image is pulling 34.1.0. KO

Picking apart each package version that changed as of my previous image, I ended up suspecting kubernetes client itself.
https://github.com/kubernetes-client/python/compare/v33.1.0...v34.1.0

Obviously, this:
https://github.com/kubernetes-client/python/commit/e3b373fc1f31821ab4cd9c288c0263c3bb1b1354#diff-977bedbaf9339b25749bc96a24266c1fa3e99ab521a835562fe84b2ab87b7001

And then again, the rest client, which is the one throwing in my case, looks OK?
https://github.com/kubernetes-client/python/blob/master/kubernetes/client/rest.py

That "should_bypass_proxy" ... looks correct ...
https://requests.readthedocs.io/en/latest/_modules/requests/utils/

We are using proxies. Ansible would connect public services (eg: aws). Our proxies would only allow for specific names out.
Ansible would connect to our clusters directly/no proxy.

Our environment would include:

```
  HTTP_PROXY: http://my-http-proxy:8080
  HTTPS_PROXY: http://my-http-proxy:8080
  NO_PROXY: github.corp.com,.github.corp.com,localhost,.local,.svc,10.0.0.0/8,127.0.0.0/8,127.0.0.1,::1,.corp.com
  http_proxy: http://my-http-proxy:8080
  https_proxy: http://my-http-proxy:8080
  no_proxy: github.corp.com,.github.corp.com,localhost,.local,.svc,10.0.0.0/8,127.0.0.0/8,127.0.0.1,::1,.corp.com
```

My cluster FQDNs would follow format `<cluster-name>.<where-it-runs>.corp.com`
So far, and testing again with my previous images running 33.1.0, connections to AWS exits through proxies, connections to clusters exits directly.
Testing 34.1.0, connection to clusters are mistakenly sent to my proxy, which denies the request.
Forcing HTTPS_PROXY/https_proxy to empty string, when calling my terraform provider: access to cluster OK, but then I broke access to AWS.

I lost that link, but I found somewhere suggesting that some client may handle subdomains differently, so I tried to add `corp.com` alongside my previous `.corp.com`, in no_proxy env var. Still no luck.
Then I tried adding the exact cluster FQDN to no_proxy env var: I can still see ansible connecting through my proxies.

domains/subdomains listed in my NO_PROXY/no_proxy should be connected to directly.

Set proxy variables as shown above. Include Kubernetes cluster domain in no_proxy.
Switch from 33.1.0 to 34.1.0. check access logs in proxy.

**Environment**:
- Kubernetes version (`kubectl version`): 1.29.7
- OS (e.g., MacOS 10.13.6): RHEL9
- Python version (`python --version`): 3.9

jcox10:
#### What type of PR is this?

/kind bug

Currently the `no_proxy` configuration is reset after it is configured due to a duplicate `self.no_proxy = None` line being added in the autogenerated file [configuration.py](https://github.com/kubernetes-client/python/blob/master/kubernetes/client/configuration.py#L173)

Initially the thought was the script [insert_proxy_config.sh](https://github.com/kubernetes-client/python/blob/master/scripts/insert_proxy_config.sh) was applying this duplicate. However, when this script is run on the openapi generated code, `configuration.py` looks correct.

Looking at the full [release.sh](https://github.com/kubernetes-client/python/blob/master/scripts/release.sh#L211) script, there is another script applying proxy settings by cherry-picking commit https://github.com/kubernetes-client/python/commit/95a893cd1c34de11a4e3893dd1dfde4a0ca30bdc in [apply-hotfixes.sh](https://github.com/kubernetes-client/python/blob/master/scripts/apply-hotfixes.sh#L70).  This commit adds `self.no_proxy = None`.  When applied after `insert_proxy_config.sh` this results in the duplicate lines and the no_proxy becoming None.

Changing the order of the applied scripts in `release.sh` so that `insert_proxy_config.sh` is run **after** `apply-hotfixes.sh` resolves the problem because `insert_proxy_config.sh` is able to detect the existing `self.no_proxy = None` and insert it's code correctly without resetting the no_proxy.

Fixes #2460 

```release-note
NONE
```

```docs

```

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -207,14 +207,14 @@ git diff-index --quiet --cached HEAD || git commit -am "update changelog"
 
 # Re-generate the client
 scripts/update-client.sh
-#edit comfiguration.py files
-scripts/insert_proxy_config.sh
 # Apply hotfixes
 rm -r kubernetes/test/
 git add .
 git commit -m "temporary generated commit"
 scripts/apply-hotfixes.sh
 git reset HEAD~2
+# Apply proxy config after hotfixes
+scripts/insert_proxy_config.sh
 
 # Custom object API is hosted in gen repo. Commit custom object API change
 # separately for easier review


jcox10:
@roycaihw @yliaog Can you review this PR?

yliaog:
/lgtm
/approve

thanks for the pr

PR #2523, opened 2026-03-13
Fix: properly load and parse NO_PROXY environment variable

Ndugu2:
#### What type of PR is this?

/kind bug

This PR fixes the handling of the `NO_PROXY` environment variable in the `Configuration` class. Previously:

1. **No parsing**: The value was stored as a raw string instead of being parsed into a list of domains
2. **Precedence issues**: When both `NO_PROXY` and `no_proxy` were set, the lowercase version could take precedence incorrectly
3. **Empty string handling**: An empty `NO_PROXY=""` was treated the same as if the variable wasn't set
4. **Default value**: `self.no_proxy` defaulted to `None` instead of an empty list

These issues caused proxy bypass settings to be ignored, breaking functionality for users relying on `no_proxy` configurations.

Fixes #2520

The fix implements proper precedence checking using `in os.environ` rather than `os.getenv()` to correctly detect when a variable exists even if it's empty. The value is now parsed into a list of domains with whitespace stripped.

Test cases added cover:
- Uppercase `NO_PROXY`
- Lowercase `no_proxy`
- Both uppercase and lowercase (precedence test)
- Empty string values
- No environment variable set
- Spaces in comma-separated values

```release-note
Fixed: The NO_PROXY environment variable is now properly parsed into a list of domains. Uppercase NO_PROXY correctly takes precedence over lowercase no_proxy. Empty NO_PROXY="" now results in an empty list rather than falling back to the lowercase variant.

```

### DIFF
--- a/kubernetes/client/configuration.py
+++ b/kubernetes/client/configuration.py
@@ -161,15 +161,28 @@ class Configuration(object):
         self.proxy = None
         """Proxy URL
         """
-        self.no_proxy = None
-# Load proxy from environment variables (if set)
+
+        self.no_proxy = []
+        # Load proxy from environment variables (if set)
         if os.getenv("HTTPS_PROXY"): self.proxy = os.getenv("HTTPS_PROXY")
         if os.getenv("https_proxy"): self.proxy = os.getenv("https_proxy")
         if os.getenv("HTTP_PROXY"): self.proxy = os.getenv("HTTP_PROXY")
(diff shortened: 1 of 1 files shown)


yliaog:
please take a look at the following fix PR:
https://github.com/kubernetes-client/python/pull/2459
https://github.com/kubernetes-client/python/pull/2513

PR #2527, opened 2026-03-20
Add health check to `watch.stream` for silent connection drops

### LINKED ISSUE #2462: Fail watch gracefully on control plane upgrade
When I run a watch on k8s jobs (probably applies to other objects) and k8s control plane gets upgraded in the meantime, the watcher just silently stops producing events. No exception is being raised.
Exception being raised or events continuing to arrive
1. start watcher: 

```python
from kubernetes import client, config, watch

config.load_incluster_config()
batch_v1_api = client.BatchV1Api()
watcher = watch.Watch()

for event in watcher.stream(
        func=batch_v1_api.list_namespaced_job,
        namespace="mynamespace",
        timeout_seconds=0,
    ):
        print(event)
```
2. upgrade k8s control plane
3. create some new jobs
4. new events are not coming in. no exceptions are being raised

Reproduced on GKE stable channel with the last two control plane upgrades
**Environment**:
- Kubernetes version (`kubectl version`): 1.33.5-gke.1080000
- OS (e.g., MacOS 10.13.6): linux
- Python version (`python --version`): 3.12.12
- Python client version (`pip list | grep kubernetes`): 33.1.0

Urvashi0109:
#### What type of PR is this?

/kind bug
/kind feature

When running a watch on Kubernetes objects (e.g., Jobs, Pods, Namespaces) and the Kubernetes control plane gets upgraded, the watch connection is silently dropped. The watcher hangs indefinitely - No exception is raised and no new events are received. This is because the TCP connection enters a half-open state where the client believes the connection is still alive, but the server side has been torn down during the upgrade.

This PR adds a `_health_check_interval` parameter to `watch.stream()` that detects silent connection drops and automatically reconnects:

- When `_health_check_interval` is set to a value > 0, a socket-level read timeout (`_request_timeout`) is configured on the HTTP connection
- If no data arrives within the specified interval, `urllib3` raises a `ReadTimeoutError`
- The watch catches this exception and automatically reconnects using the last known `resource_version`, ensuring no events are missed
- The feature is disabled by default (`_health_check_interval=0`), preserving full backward compatibility
- When disabled, `ReadTimeoutError` propagates to the caller as before

Fixes #2462

- This PR takes approach: leveraging urllib3's existing read timeout mechanism (`_request_timeout`) to break out of the blocking read, then catching the resulting `ReadTimeoutError`/`ProtocolError` exceptions
- The _ prefix on `_health_check_interval` follows the existing convention in this codebase (e.g., _preload_content, _request_timeout) for parameters that are consumed by the client library rather than passed to the API server
- 5 new unit tests added, all 24 tests (19 existing + 5 new) pass with zero regressions

```release-note
Added `_health_check_interval` parameter to `watch.stream()` to detect and recover from silent connection drops during Kubernetes control plane upgrades. When set to a value > 0 (seconds), the watch will automatically reconnect if no data is received within the specified interval. Disabled by default for backward compatibility.

```

```docs
NONE
```

### DIFF
--- a/kubernetes/base/watch/watch.py
+++ b/kubernetes/base/watch/watch.py
@@ -264,7 +264,7 @@ class Watch(object):
                 if should_retry:
                     # Add a small sleep to avoid a tight reconnect loop
                     # in case the endpoint is hard-down or errors immediately.
-                    time.sleep(min(1.0,health_check_interval))
+                    time.sleep(min(1.0, health_check_interval))
                 else:
                     raise
             finally:
--- a/kubernetes/base/watch/watch.py
+++ b/kubernetes/base/watch/watch.py
@@ -15,6 +15,9 @@
 import json
 import pydoc
 import sys
+import time
+
+from urllib3.exceptions import ProtocolError, ReadTimeoutError
 
 from kubernetes import client
 
@@ -154,6 +157,14 @@ class Watch(object):
         :param func: The API function pointer. Any parameter to the function
                      can be passed after this parameter.
 
+        :param int _health_check_interval: Optional. Number of seconds to wait
+            for data before assuming the connection has been silently dropped
+            (e.g., during a control plane upgrade). When set to a value > 0,
+            the watch will automatically detect silent connection drops and
+            reconnect using the last known resource_version. Default is 0
+            (disabled). This is useful for long-running watches that need to
+            survive cluster upgrades.
+
         :return: Event object with these keys:
                    'type': The type of event such as "ADDED", "DELETED", etc.
                    'raw_object': a dict representing the watched object.
@@ -172,6 +183,11 @@ class Watch(object):
                 ...
                 if should_stop:
                     watch.stop()
+
+        Example with health check (survives control plane upgrades):
+            for e in watch.stream(v1.list_namespace, _health_check_interval=30):
+                # If no data received for 30s, watch reconnects automatically
+                print(e['type'], e['object'].metadata.name)
         """
 
         self._stop = False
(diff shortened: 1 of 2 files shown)


roycaihw:
I thought we already have multiple timeouts (supporting both client side and server side). Could you check if the existing timeouts address this problem? Or is it (HTTP connection timeout) supported in client-go only?

roycaihw:
Here is an example of using `timeout_seconds` with watch: https://github.com/kubernetes-client/python/blob/bdfbaea9094fff7b5223d1c1b114f07f6d9f17ca/examples/watch/pod_namespace_watch.py#L18

cc @yliaog

Urvashi0109:
Thanks for feedback @roycaihw , I will take a look 😊

Urvashi0109:
@roycaihw Great question! I've investigated the existing timeout mechanisms, and they don't fully address the silent connection drop problem during control plane upgrades. Here's why:

- `timeout_seconds` (Server-side timeout): Setting `timeout_seconds` disables automatic retries ([line 203](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/watch/watch.py#L203)), so when a connection drops, the watch stops completely. During control plane upgrades, we need the watch to *recover* automatically, not just stop.
-  `_request_timeout` (Client-side socket timeout): While `_request_timeout` can detect stalled connections via `ReadTimeoutError`, without this PR the exception propagates and the watch stops. Users would need to manually catch the exception, track `resource_version`, and implement reconnection logic.

With this PR: `_health_check_interval` combines `_request_timeout` detection (([line 218](https://github.com/kubernetes-client/python/pull/2527/files#diff-5e3b3c3d3f3b3c3d3f3b3c3d3f3b3c3d)): sets read timeout) with automatic reconnection logic ([lines 256-269](https://github.com/kubernetes-client/python/pull/2527/files#diff-5e3b3c3d3f3b3c3d3f3b3c3d3f3b3c3d))


this PR makes the Python client behave like client-go does during control plane upgrades - automatic recovery.

Urvashi0109:
Hi @roycaihw , just following up to see if there are any updates here. Please let me know if you have any questions or need any additional information from my side 😊

PR #2532, opened 2026-03-27
upgraded openapi generator to v6.6.0

yliaog:
#### What type of PR is this?
/kind cleanup

upgraded openapi generator to v6.6.0, cleaned up unneeded patches

Fixes #

```release-note
Legacy dict(str, str) syntax is no longer supported in in ApiClient deserializer. Only modern dict[str, str] syntax is supported.
```

```docs

```


kubernetes/client/api/admissionregistration_v1_api.py
```diff
@@ -42,25 +42,34 @@ def create_mutating_webhook_configuration(self, body, **kwargs):  # noqa: E501
...
+        :type field_manager: str
+        :param field_validation: fieldValidation instructs the server on how to handle objects in the request (POST/PUT/PATCH) containing unknown or duplicate fields. Valid values are: - Ignore: This will ignore any unknown fields that are silently dropped from the object, and will ignore all but the last duplicate field that the decoder encounters. This is the default behavior prior to v1.23. - Warn: This will send a warning via the standard warning response header for each unknown field that is dropped from the object, and for each duplicate field that is encountered. The request will still succeed if there are no other errors, and will only persist the last of any duplicate fields. This is the default in v1.23+ - Strict: This will fail the request with a BadRequest error if any unknown fields would be dropped from the object, or if any duplicate fields are present. The error returned from the server will contain all unknown and duplicate fields encountered.
```
roycaihw, line 57: field_manager and field_validation are new k8s APIs in 1.36?

kubernetes/client/apis/__init__.py
```diff
@@ -1,13 +0,0 @@
-from __future__ import absolute_import
```
roycaihw, line 1: Note to myself: we need to keep this file

kubernetes/client/models/admissionregistration_v1_service_reference.py
```diff
@@ -49,7 +52,7 @@ class AdmissionregistrationV1ServiceReference(object):
     def __init__(self, name=None, namespace=None, path=None, port=None, local_vars_configuration=None):  # noqa: E501
         """AdmissionregistrationV1ServiceReference - a model defined in OpenAPI"""  # noqa: E501
         if local_vars_configuration is None:
-            local_vars_configuration = Configuration()
+            local_vars_configuration = Configuration.get_default_copy()
```
roycaihw, line 55: I recall we have something related to default config loading in the past. Does this impact the behavior?

kubernetes/client/api_client.py
```diff
@@ -120,9 +120,10 @@ def set_default_header(self, header_name, header_value):
     def __call_api(
             self, resource_path, method, path_params=None,
             query_params=None, header_params=None, body=None, post_params=None,
-            files=None, response_type=None, auth_settings=None,
```
roycaihw, line 123: Is this a breaking change, or is it okay because it's internal?

kubernetes/client/api_client.py
```diff
@@ -499,45 +521,68 @@ def select_header_accept(self, accepts):
         else:
             return ', '.join(accepts)
 
-    def select_header_content_type(self, content_types):
+    def select_header_content_type(self, content_types, method=None, body=None):
```
roycaihw, line 524: Note to myself: this is another place that we need to pay attention

kubernetes/client/api_client.py
```diff
@@ -499,45 +521,68 @@ def select_header_accept(self, accepts):
...
         else:
             return content_types[0]
 
-    def update_params_for_auth(self, headers, querys, auth_settings):
+    def update_params_for_auth(self, headers, queries, auth_settings,
+                               request_auth=None):
         """Updates header and query params based on authentication setting.
 
         :param headers: Header parameters dict to be updated.
-        :param querys: Query parameters tuple list to be updated.
```
roycaihw, line 522: Similar here. Is this a breaking change?

kubernetes/client/configuration.py
```diff
@@ -161,17 +206,6 @@ def __init__(self, host="http://localhost",
         self.proxy = None
         """Proxy URL
         """
-        self.no_proxy = None
```
roycaihw, line 164: Are we continue to carry the no proxy and http proxy patches?

kubernetes/client/configuration.py
```diff
@@ -345,12 +394,14 @@ def auth_settings(self):
         :return: The Auth Settings information dict.
         """
         auth = {}
-        if 'authorization' in self.api_key:
```
roycaihw, line 348: Was this a bug?

kubernetes/client/rest.py
```diff
@@ -160,12 +154,7 @@ def request(self, method, url, query_params=None, headers=None,
             if method in ['POST', 'PUT', 'PATCH', 'OPTIONS', 'DELETE']:
                 if query_params:
                     url += '?' + urlencode(query_params)
-                if (re.search('json', headers['Content-Type'], re.IGNORECASE) or
```
roycaihw, line 163: This is another patch that we need to continue carry.

roycaihw:
Let's also add back the `dict[]` test (in `kubernetes/test/test_api_client.py`) as discussed.

kubernetes/client/apis/__init__.py
roycaihw: Confirmed this is kept.

kubernetes/client/models/admissionregistration_v1_service_reference.py
roycaihw: Let's announce this. The upstream change is: https://github.com/OpenAPITools/openapi-generator/pull/8500

kubernetes/client/api_client.py
roycaihw: This is okay because it's internal (`__call_api`)

kubernetes/client/api_client.py
roycaihw: Let's announce this.

kubernetes/client/configuration.py
roycaihw: Confirmed that the patch is kept.

kubernetes/client/configuration.py
roycaihw: Let's announce this.

kubernetes/client/rest.py
roycaihw: Confirmed that this is kept.

kubernetes/client/api/admissionregistration_v1_api.py
roycaihw: They are k8s API, so it's fine: https://github.com/kubernetes/kubernetes/pull/105916

kubernetes/client/api_client.py
roycaihw: Confirmed the patch is kept.

kubernetes/client/rest.py
```diff
@@ -226,15 +221,22 @@ def request(self, method, url, query_params=None, headers=None,
         if _preload_content:
             r = RESTResponse(r)
 
-            # In the python 3, the response.data is bytes.
-            # we need to decode it to string.
-            if six.PY3:
-                r.data = r.data.decode('utf8')
```
roycaihw, line 232: @yliaog Do you know why this was dropped? Could it be related to the `str response type cannot be decoded ` issue you saw?

During your test, could it be that 
1. The api_client.py decoded str once 
2. The rest.py decoded str for a second time, hitting the "cannot be decoded" error.

If that's the case, maybe the correct solution is:
1. Remove the decode in rest.py (same as what you have in this PR)
2. Let api_client.py to handle decode for all types.

kubernetes/base/stream/ws_client.py
```diff
@@ -254,8 +254,16 @@ def close(self, **kwargs):
             self.sock.close(**kwargs)
 
 
-WSResponse = collections.namedtuple('WSResponse', ['data'])
+#WSResponse = collections.namedtuple('WSResponse', ['data', 'status', 'getheader'])
```
roycaihw: Remove this line?

kubernetes/client/rest.py
yliaog: previously, as you can see in the deleted code here, response.data is bytes, bytes can be decoded to string

after the upgrade to v6.6.0, response.data is str, str cannot be decoded, as discussed here, https://share.google/aimode/t9bV2jOOKREzrlbRm

i think it is correct to drop the code here, as str cannot be decoded. and for the same reason, api_client.py should not try to decode str, in addition to file, and bytes there.

kubernetes/client/api_client.py
yliaog: that is correct.

kubernetes/base/stream/ws_client.py
yliaog: done

kubernetes/client/models/admissionregistration_v1_service_reference.py
yliaog: added to Breaking change in CHANGELOG

kubernetes/client/api_client.py
yliaog: added to Breaking change in CHANGELOG

kubernetes/client/configuration.py
yliaog: added to Breaking change in CHANGELOG

roycaihw (approved):
/lgtm
/approve

PR #2534, opened 2026-04-21
Extend Resource.watch() to accept _request_timeout

### LINKED ISSUE #2533: Support _request_timeout in dynamic client watch()
**Problem**

When using the dynamic client’s watch() method:
```
api = self._client.resources.get(
    api_version=self._api_version,
    kind=self._kind
)

for event in api.watch(
    timeout=server_timeout_seconds,
)
```

there is currently no way to configure a client-side request timeout (`_request_timeout`).

**Current Workaround**

To enforce a client-side timeout, we have to bypass `Resource.watch()` and directly use `Watch().stream()`:

```
for raw_event in Watch().stream(
    api.get,
    serialize=False,
    deserialize=False,
    timeout_seconds=self._server_timeout_seconds,
    _request_timeout=self._client_timeout_seconds,
)
```

**Issue**

The dynamic client’s watch() method internally relies on `Watch().stream()` but does not expose `_request_timeout` as a parameter: https://github.com/kubernetes-client/python/blob/master/kubernetes/base/dynamic/client.py#L203

This creates:
Inconsistent API ergonomics
Unnecessary duplication of logic for users needing client-side timeouts

**Proposed Solution**

Extend Resource.watch() to accept an optional `_request_timeout` parameter and pass it through to the underlying Watch().stream() call.

Example:

```
api.watch(
    timeout=server_timeout_seconds,
    _request_timeout=(connect_timeout, read_timeout),
)
```

Polimixanos:
#### What type of PR is this?
/kind feature
/kind api-change

This PR extends the dynamic client's `Resource.watch()` method to accept an optional `_request_timeout` parameter. 

Previously, there was no way to configure a client-side request timeout (e.g., for connection or read timeouts) when using the dynamic client's high-level `watch()` API. Users were forced to bypass `Resource.watch()` and directly use `Watch().stream()` to achieve this. By exposing this parameter in `DynamicClient.watch()`, we provide consistent API ergonomics and eliminate the need for users to duplicate internal logic.

The `_request_timeout` parameter is now passed through to the underlying `Watch().stream()` call and handled by the base `ApiClient`.

Fixes https://github.com/kubernetes-client/python/issues/2533

The change is localized to `kubernetes/base/dynamic/client.py`. Since `Resource`, `Subresource`, and `ResourceList` all proxy their method calls to the `DynamicClient`, this update automatically enables the parameter for all dynamic resource types.

```release-note
Added `_request_timeout` parameter to the dynamic client's `watch()` method to allow configuration of client-side request timeouts.


Example usage:
api.watch(
    timeout=server_timeout_seconds,
    _request_timeout=(connect_timeout, read_timeout),
)

### DIFF
--- a/kubernetes/base/dynamic/client.py
+++ b/kubernetes/base/dynamic/client.py
@@ -163,7 +163,7 @@ class DynamicClient(object):
 
         return self.request('patch', path, body=body, force_conflicts=force_conflicts, **kwargs)
 
-    def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None, watcher=None, allow_watch_bookmarks=None):
+    def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None, watcher=None, allow_watch_bookmarks=None, _request_timeout=None):
         """
         Stream events for a resource from the Kubernetes API
 
@@ -177,6 +177,7 @@ class DynamicClient(object):
         :param timeout: The amount of time in seconds to wait before terminating the stream
         :param watcher: The Watcher object that will be used to stream the resource
         :param allow_watch_bookmarks: Ask the API server to send BOOKMARK events
+        :param _request_timeout: The amount of time in seconds to wait for a request to complete
 
         :return: Event object with these keys:
                    'type': The type of event such as "ADDED", "DELETED", etc.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Polimixanos** [comment]
Just signed CLA

File: kubernetes/base/dynamic/client.py [hunk 1]
```diff
@@ -163,7 +163,7 @@ def server_side_apply(self, resource, body=None, name=None, namespace=None, forc
 
         return self.request('patch', path, body=body, force_conflicts=force_conflicts, **kwargs)
 
-    def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None, watcher=None, allow_watch_bookmarks=None):
+    def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None, watcher=None, allow_watch_bookmarks=None, _request_timeout=None):
```
yliaog, line 166: it's better to use request_timeout instead of _request_timeout

yliaog:
the async client also needs the change?

https://github.com/kubernetes-client/python/blob/master/kubernetes/aio/dynamic/client.py

PR #2547, opened 2026-04-27
added basic asyncio support

yliaog:
#### What type of PR is this?
/kind feature

Fixes #

```release-note
added basic asyncio support, still missing dynamic client, watch, stream etc
```

```docs

```


aojea:
👏

scripts/asyncio/rest_client_patch_read_bufsize.diff
```diff
@@ -0,0 +1,20 @@
...
+             connector=connector,
+-            trust_env=True
++            trust_env=True,
++            # Watch events containing large resource objects can exceed
++            # aiohttp's default read buffer size.
++            #
++            # There is no hard-limit defined by k8s, but the etcd default
++            # maximum request size is 1.5MiB.
++            # https://github.com/kubernetes/kubernetes/issues/19781
++            read_bufsize=2**21
```
roycaihw, line 17: (Non blocking for this PR) Some of the patches like this seem to also apply to the syncio client.

kubernetes_asyncio/config/exec_provider.py
```diff
@@ -0,0 +1,101 @@
...
+                name = item['name']
+                value = item['value']
+                additional_vars[name] = value
+            self.env.update(additional_vars)
+
+    async def run(self, previous_response=None):
+        # Validate the run can be executed on Windows
+        if type(asyncio.get_event_loop()).__name__ == '_WindowsSelectorEventLoop':
+            raise ConfigException(
+                'exec: _WindowsSelectorEventLoop does NOT support subprocesses, see README.md'
```
roycaihw, line 57: We also need to port some of the root README.md to this repo

roycaihw (approved):
One minor comment regarding README. LGTM otherwise.

roycaihw:
Test results look good. 

/lgtm
/approve

scripts/asyncio/rest_client_patch_read_bufsize.diff
yliaog: i'm not sure this applies to synchronous client, the reason to have this patch is due to the use of aiohttp as the comment mentioned.

+            # Watch events containing large resource objects can exceed
+            # aiohttp's default read buffer size

tomplus:
I'm happy to see you are working on it.

I'm ready to help but to be honest I'm worried about upgrading the generator.  I'm not sure if you are aware of this issue: https://github.com/tomplus/kubernetes_asyncio/issues/293 So I decided to stick to the old version of generator in my async lib. I applied patches if needed, I developed typing as a set of `*.pyi` files generated from spec etc.

In addition, I'd suggest copying my entire lib into the kubernetes-client/python-asyncio. Could you consider this? I know, maintaining two libraries may be difficult, but I believe discrepancies between sync and async are inevitable and it wouldn't be any easier maintained in a single repo.

Thanks!

yliaog:
@tomplus it's great to have your help

I was not aware of the issue https://github.com/tomplus/kubernetes_asyncio/issues/293. I did a couple of quick tests using this repo(the openapi generator was upgraded to v6.6.0).

~/prj/client-python$ cat kubernetes_asyncio/.openapi-generator/COMMIT 
Requested Commit/Tag : v6.6.0
Actual Commit        : 7f8b853f502d9039c9a0aac2614ce92871e895ed

~/prj/client-python$ time PYTHONPATH=. python -m kubernetes_asyncio.client.api_client
real    0m0.548s
user    0m0.440s
sys     0m0.108s

the result is 0.5s, not the 12s.

~/prj/client-python$ time PYTHONPATH=. python examples_asyncio/list_pods.py
... ... ... 
... ... ... 

real    0m1.883s
user    0m1.444s
sys     0m0.205s

~/prj/client-python$ time kubectl get po -A

real    0m1.149s
user    0m0.927s
sys     0m0.128s

the result is 1.8s, vs 1.1 using kubectl

I think these results are acceptable, lazy imports may help futher improve it. what do you think? did i miss anything?

yliaog:
> In addition, I'd suggest copying my entire lib into the kubernetes-client/python-asyncio. Could you consider this? I know, maintaining two libraries may be difficult, but I believe discrepancies between sync and async are inevitable and it wouldn't be any easier maintained in a single repo.
> 

the goal is to keep the synchronous and asyncio clients feature parity. Some features and bug fixes are available in synchronous client, but not in asyncio client, and vice versa. The major code lies in kubernetes/ and kubernetes_asyncio directories, i think keeping them in one repo is easier to ask contributors to fix bugs in both clients, and add features to both at the same time.

another goal is to share the code between the two clients as much as possible, ideally, the only differences between the two are the inevitable ones, like await, async. I think one repo also makes it easier to share. For exmaple, the config/in_cluster_config is almost identical. For now, I kept them separate, it may be refactored in the future to share one file.

$ diff kubernetes/config/incluster_config.py kubernetes_asyncio/config/incluster_config.py 
18c18

from kubernetes_asyncio.client import Configuration

tomplus:
> I think these results are acceptable, lazy imports may help futher improve it. what do you think? did i miss anything?

Wow! I have to test it again. Could you provide your Python version and other deps (especially pydantic)?

> The major code lies in kubernetes/ and kubernetes_asyncio directories

Sure, we can try that, fingers crossed :crossed_fingers:

yliaog:
~/prj/client-python$ python --version
Python 3.11.2

the repo head (https://github.com/kubernetes-client/python) is currently using openapi-generator v6.6.0, that is not using pydantic

openapi-generator v7.0.0 is the first version using pydantic 1
➜  resources git:(v7.0.0) grep -r pydantic python*
python/requirements.mustache:pydantic >= 1.10.5, < 2

openapi-generator v7.1.0 is the first version using pydantic 2, and the previous one was copied into python-pydantic-v1.
➜  resources git:(v7.1.0) grep -r pydantic python*
python/requirements.mustache:pydantic >= 2
python-pydantic-v1/requirements.mustache:pydantic >= 1.10.5, < 2

tomplus:
Aha, that explains it. I tested on v7.x and it wasn't good.

PR #2585, opened 2026-05-21
config: write api_key['BearerToken'] so v36+ SDK auth works

### LINKED ISSUE #2582: `Configuration.auth_settings()` returns empty dict in v36.0.0 - all BearerToken-authed requests go anonymous
### What happened

Upgrading from `kubernetes==35.0.0` to `kubernetes==36.0.0` causes every API request that relies on a Bearer token (the in-cluster ServiceAccount auth path, or any caller that sets `Configuration.api_key['authorization']` directly) to be sent without an `Authorization` header. The server treats the request as `system:anonymous` and rejects it (401 on AKS, 403 on minikube).

### Minimal reproducer

Direct check of the method the generated API operations call to attach auth headers before sending HTTP requests:

```python
from kubernetes import client

cfg = client.Configuration()
cfg.api_key['authorization'] = 'Bearer fake-token'

api = client.ApiClient(cfg)
headers = {}
api.update_params_for_auth(headers, [], ['BearerToken'])
print('headers:', headers)
```

### Expected output (matches v35.0.0)

```
headers: {'authorization': 'Bearer fake-token'}
```

### Actual output on v36.0.0

```
headers: {}
```

The Bearer-token-aware code path is silently a no-op. The HTTP request goes out with no `Authorization` header and the server treats it as `system:anonymous`.

### Supporting evidence

`Configuration.auth_settings()` - which `update_params_for_auth` reads to find scheme definitions - is also empty on v36:

```python
>>> cfg.auth_settings()                # v35: {'BearerToken': {...}}
{}                                     # v36
```

So the regression is at the `Configuration` level: the `BearerToken` scheme entry that should be present in `auth_settings()` (and is in v35) is missing in v36, which cascades into `update_params_for_auth` having nothing to apply.

### Impact

Any pod doing in-cluster API calls via this library is broken on v36.0.0 - including (but not limited to) callers of `CoordinationV1Api`, `CoreV1Api`, etc. We caught this when our Lease-coordinated workload pods started failing every `read_namespaced_lease` call with 401 immediately after a rebuild upgraded the lib from v35 to v36.

### Environment

- `kubernetes==36.0.0` (uploaded to PyPI 2026-05-20T20:44Z)
- Python 3.12 / 3.13 (reproduced on both)
- Linux container (also reproduced on macOS in a clean venv)

### Regression window

The bug is present in the entire 36-series:

| Version | PyPI upload | `update_params_for_auth` result |
|---|---|---|
| v35.0.0 | 2026-01-16 | populates `{'authorization': 'Bearer ...'}` |
| v36.0.0a3 | 2026-05-08 | empty `{}` (regression introduced here) |
| v36.0.0b1 | 2026-05-13 | empty `{}` |
| v36.0.0 | 2026-05-20 | empty `{}` |

So the regression was present from the first 36-series prerelease and shipped to stable without a fix. The diff between `v35.0.0..v36.0.0a3` should contain the offending change.

### Workaround

Pin `kubernetes>=29,<36` until a v36 patch ships.

### Hypothesis (haven't dug in further)

This looks like the OpenAPI/swagger-driven code generation for `Configuration` regressed - the dict literal that defines the `BearerToken` entry under `auth_settings()` likely got dropped or scoped out. Worth diffing the generated `Configuration` class between v35.0.0 and v36.0.0.

jmacek:
## What this PR does

Updates the in-cluster and kubeconfig loaders in `kubernetes/base/config/`, plus `kubernetes_asyncio/config/incluster_config.py`, to write the bearer token under `api_key['BearerToken']` (the v36+ lookup key) instead of `api_key['authorization']` (the v35 lookup key). Updates the corresponding test assertions to read the new key. Adds a regression test that exercises the end-to-end flow (Configuration → ApiClient → outgoing Authorization header) which the existing tests did not cover.

## Why

v36 rewrote `Configuration.auth_settings()` to look up the bearer-token credential under `api_key['BearerToken']`, matching the OpenAPI security scheme name. This rename was declared a breaking change in the project CHANGELOG as part of the openapi-generator v6.6.0 upgrade. The hand-written loaders in `kubernetes/base/config/` and `kubernetes_asyncio/config/incluster_config.py` were not updated to follow the rename - they still write `api_key['authorization']`. (Asyncio's `kube_config.py` already writes `'BearerToken'`, so it isn't affected.)

Net effect on v36: every call to `load_incluster_config()` (and `load_kube_config()` with a static token, including the async equivalents) produces a `Configuration` whose `auth_settings()` yields no bearer credential, so outgoing API requests are sent without an `Authorization` header and the apiserver treats them as `system:anonymous`. The failure mode is silent - no warning, no exception, just 401s from every API call.

See #2582 for the user-side report and a minimal repro.

## Test plan

- Added `test_load_incluster_sets_request_authorization_header` to `kubernetes/base/config/incluster_config_test.py`. It drives a real `ApiClient.update_params_for_auth()` against a freshly-loaded `Configuration` and asserts the resulting headers contain an `Authorization` entry - the end-to-end invariant that v36 broke.
- Verified the new regression test fails on `kubernetes==36.0.0` against the unfixed loader (`AssertionError: 'authorization' not found in {}`) and passes with the fix.
- Verified all 103 sync tests in `kubernetes/base/config/` and 72 async tests in `kubernetes_asyncio/config/` pass on v36 with the change applied.

Fixes #2582

```release-note
Fix `load_incluster_config()` and `load_kube_config()` (sync and async, with a static token) so requests carry an `Authorization` header on `kubernetes-client/python` v36+. Without this fix, in-cluster pods upgrading to v36 silently send unauthenticated requests and the apiserver rejects them as `system:anonymous`.
```

### DIFF
--- a/kubernetes/base/config/incluster_config.py
+++ b/kubernetes/base/config/incluster_config.py
@@ -88,7 +88,6 @@ class InClusterConfigLoader:
         client_configuration.host = self.host
         client_configuration.ssl_ca_cert = self.ssl_ca_cert
         if self.token is not None:
-            client_configuration.api_key['authorization'] = self.token
             client_configuration.api_key['BearerToken'] = self.token
         if not self._try_refresh_token:
             return
--- a/kubernetes/base/config/incluster_config.py
+++ b/kubernetes/base/config/incluster_config.py
@@ -89,6 +89,7 @@ class InClusterConfigLoader:
         client_configuration.ssl_ca_cert = self.ssl_ca_cert
         if self.token is not None:
             client_configuration.api_key['authorization'] = self.token
+            client_configuration.api_key['BearerToken'] = self.token
         if not self._try_refresh_token:
             return
 
--- a/kubernetes/base/config/incluster_config_test.py
+++ b/kubernetes/base/config/incluster_config_test.py
@@ -108,6 +108,23 @@ class InClusterConfigTest(unittest.TestCase):
         self.assertEqual('bearer ' + _TEST_NEW_TOKEN, loader.token)
         self.assertGreater(loader.token_expires_at, old_token_expires_at)
 
+    def test_load_incluster_sets_request_authorization_header(self):
+        from kubernetes.client import ApiClient
+        cert_filename = self._create_file_with_temp_content(_TEST_CERT)
+        loader = self.get_test_loader(cert_filename=cert_filename)
+        config = Configuration()
+        loader.load_and_set(config)
+
+        api_client = ApiClient(config)
+        headers = {}
+        api_client.update_params_for_auth(headers, [], ['BearerToken'])
+
+        self.assertIn('authorization', headers)
+        self.assertTrue(
+            headers['authorization'].lower().startswith('bearer '),
+            "Expected a Bearer authorization header, got: %r"
+            % headers['authorization'])
+
     def _should_fail_load(self, config_loader, reason):
         try:
             config_loader.load_and_set()
--- a/kubernetes/base/config/incluster_config_test.py
+++ b/kubernetes/base/config/incluster_config_test.py
@@ -91,7 +91,7 @@ class InClusterConfigTest(unittest.TestCase):
         loader.load_and_set(config)
 
         self.assertEqual('bearer ' + _TEST_TOKEN,
-                         config.get_api_key_with_prefix('authorization'))
+                         config.get_api_key_with_prefix('BearerToken'))
         self.assertEqual('bearer ' + _TEST_TOKEN, loader.token)
         self.assertIsNotNone(loader.token_expires_at)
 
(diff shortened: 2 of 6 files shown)


kubernetes/base/config/incluster_config.py
```diff
@@ -89,6 +89,7 @@ def _set_config(self, client_configuration):
         client_configuration.ssl_ca_cert = self.ssl_ca_cert
         if self.token is not None:
             client_configuration.api_key['authorization'] = self.token
+            client_configuration.api_key['BearerToken'] = self.token
```
faust64, line 91: Does it make sense to have 2 configurations with that same token?
Either one or the other would be fine. It has to be consistent everywhere.

Do we know why that BearerToken was introduced? What was the intention?

roycaihw: I think this is related https://github.com/kubernetes-client/python/blob/master/CHANGELOG.md#breaking-change-from-upgrading-openapi-generator-to-v660. It was introduced in upstream client generator. Same question here. I think upstream did a replacement, rather than keeping both. cc @yliaog

jmacek: Thanks both - the original intent of writing both keys was to hedge backward compatibility for any third-party code introspecting `api_key['authorization']` directly. But since the rename was already documented as a breaking change in the v6.6.0 generator upgrade, aligning the loaders with that decision is the right call. Switching to a straight replacement.

yliaog:
thanks for the fix

/lgtm
/approve

aojea:
I thought that this was covered by current e2e tests, can we add one test case?

yliaog:
Yes, the current e2e tests are all using config outside cluster, it's better to add one using in cluster config.

PR #2586, opened 2026-05-21
Nest async client under kubernetes.aio to fix kubernetes-asyncio collision

### LINKED ISSUE #2583: `kubernetes`@36.0.0 overwrites existing `kubernetes-asyncio` package

Installing `kubernetes`@36.0.0 overwrites the existing installation of [`kubernetes-asyncio`](https://pypi.org/project/kubernetes-asyncio/)

With the inclusion of an async version of the client in 36.0.0, I expected the client to be made available as a submodule, e.g. at `kubernetes.kubernetes_asyncio`. Instead, when installed after `kubernetes-asyncio` it overwrites some files from `kubernetes-asyncio`, causing various errors (the same can happen in reverse as well, where installing `kubernetes` and then `kubernetes-asyncio` can cause errors).

Install `kubernetes-asyncio` and then `kubernetes` with pip or another python package manager

I'm not sure if there's planned coordination between the asyncio client included in this repo and the one at [https://github.com/tomplus/kubernetes_asyncio](tomplus/kubernetes_asyncio), but it seems very odd to me that this one would overwrite files in the other's installation. If `kubernetes-asyncio` was not already an existing package on pypi this would be more understandable, but currently it's not mentioned on the changelog that this installs to `kubernetes_asyncio` (it is on the readme as of #2548, but not in the changelog for v36.0.0).

**Environment**:
- Kubernetes version (`kubectl version`)
  - `Client Version: v1.36.0`
- OS (e.g., MacOS 10.13.6):
  - macOS 15.7.5 (24G624) and debian 13 slim (via `docker.io/library/python:3.12-slim`)
  - 3.12, 3.13, 3.14
  - ```
    kubernetes               36.0.0
    kubernetes-asyncio       35.0.1
    ```

karimzakzouk:
#### What type of PR is this?

/kind bug

Right now, `setup-release.py` installs the new async client as a top-level package called `kubernetes_asyncio`. The problem is that this directly collides with the existing `kubernetes-asyncio` package on PyPI. When people try to install both, they overwrite each other's files and break.

This PR fixes that by nesting the async client under `kubernetes.aio` instead (using setuptools `package_dir`). This matches what other big libraries do (like `redis.asyncio` or `grpc.aio`) and cleanly avoids the naming collision. 

The main changes:
- Updated `setup-release.py` to map the package to `kubernetes.aio`
- Switched the absolute imports in `kubernetes_asyncio/` to relative imports so the code works whether it's installed via the release bundle or as a standalone package for dev.
- Added a simple `try/except` in `kubernetes/__init__.py` to optionally expose the `aio` module.
- Updated the examples to use the new `kubernetes.aio` import path.

Fixes #2583

Most of the files changed here are just bulk mechanical fixes (switching `from kubernetes_asyncio.client...` to relative imports) across the code. 

The actual logic changes are pretty small and isolated to `setup-release.py`, the two `__init__.py` files, and `api_client.py`.

*Note: We'll probably want to update the OpenAPI generator templates in a follow-up PR to spit out relative imports by default for future generation runs.*

```release-note
The bundled async client is now available at `kubernetes.aio` instead of `kubernetes_asyncio` to prevent overwriting the external `kubernetes-asyncio` package on PyPI. If you were using the new async client from v36.0.0, please update your imports from `from kubernetes_asyncio import client` to `from kubernetes.aio import client`.
```

```docs
NONE
```


yliaog:
@karimzakzouk thanks for the PR, many files in this repo are generated, it is better to fix it in the script generator.

I created a PR https://github.com/kubernetes-client/python/pull/2590 to switch from kubernetes_asyncio to kuberentes_aio to avoid the conflict. Please take a look.

karimzakzouk:
@yliaog That makes perfect sense, fixing it directly in the generator scripts is the right move. Thanks for the comment

karimzakzouk:
@yliaog Following up on your comment. Yes, I agree we should fix it in the generator scripts, so I updated this PR to do exactly that.

I also added a script to `scripts/update-client-asyncio.sh`. It automatically changes all the absolute imports to relative imports right after the generation finishes.
I left a comment on #2590 to explain how this solves the namespace issue as `kubernetes.aio`, while keeping the repository directory clean and flat for us. 

All tests pass locally. Please take a look.

yliaog:
@karimzakzouk I think it is better for the synchronous and asyncio clients to use similar import path, say, one uses kubernetes.client, the other uses kubernetes.aio.client. we can discuss switching to relative imports with the community, for both sync and asyncio clients. For now, I think it is fine to nest the whole async client under kubernetes/aio/ directory.

PR #2590, opened 2026-05-22
switched kubernetes_asyncio to kubernetes_aio

### LINKED ISSUE #2583: `kubernetes`@36.0.0 overwrites existing `kubernetes-asyncio` package

yliaog:
#### What type of PR is this?
/kind bug

switched kubernetes_asyncio to kubernetes_aio

Fixes https://github.com/kubernetes-client/python/issues/2583

```release-note
NONE
```

```docs

```


yliaog:
@karimzakzouk could you take a look at this PR?

mistydemeo:
Instead of installing a second top-level package, maybe it would make more sense to nest this under `kubernetes`? eg `kubernetes.aio`

roycaihw (approved):
/lgtm
/approve

roycaihw:
/hold

per https://github.com/kubernetes-client/python/pull/2590#issuecomment-4520529708 cc @yliaog

yliaog:
> Instead of installing a second top-level package, maybe it would make more sense to nest this under `kubernetes`? eg `kubernetes.aio`

kubernetes/ for synchronous client
  client/
  config/
  e2e_test/
  ...
kubernetes_aio/ for asynchronous client
  client/
  config/
  e2e_test/
  ...

the two have the same directory structure. It may be better to make them both top-level, rather than one nested inside another, which will look like this:
kubernetes/
  client/
  config/
  e2e_test/
  ...
  aio/
  aio/client/
  aio/config/
  aio/e2e_test/
    ...

Let me add to the CHANGELOG that  aio package is experimental, we may change it in future releasses if the community prefers the other approach.

yliaog:
@roycaihw @mistydemeo

please see the CHANGELOG update:

https://github.com/kubernetes-client/python/pull/2590/changes/2d804045d59fa395db60f77eb3cdcdadb40648da

mistydemeo:
> the two have the same directory structure. It may be better to make them both top-level, rather than one nested inside another, which will look like this:

I understand, but given that there's already been one issue from writing to a package owned by another package, it really feels like the safest solution here is to nest it under the existing package.

yliaog:
> > the two have the same directory structure. It may be better to make them both top-level, rather than one nested inside another, which will look like this:
> 
> I understand, but given that there's already been one issue from writing to a package owned by another package, it really feels like the safest solution here is to nest it under the existing package.

That makes sense. Please take a look at PR: https://github.com/kubernetes-client/python/pull/2593

karimzakzouk:
@yliaog Thanks for the ping

I actually agree with both of you here. @yliaog is totally right that physically putting the `aio` folder inside `kubernetes/` in the repo would make the generator output super messy. 

But I also agree with @mistydemeo that pushing a new top level `kubernetes_aio` package to PyPI is risky and might just cause the exact same namespace collisions we're trying to fix right now.

I pushed a working concept over in #2586. 

Basically, it keeps the `kubernetes_asyncio` directory totally flat in the repo (so the generator stays clean), but  installs under the `kubernetes.aio` namespace for end users. 

I hooked a script into `update-client-asyncio.sh` that automatically swaps the absolute imports to relative imports right after generation. Then `setup-release.py` just uses `package_dir={'kubernetes.aio': 'kubernetes_asyncio'}` during the actual install.

It keeps the repo structure exactly how it is now, but solves the PyPI namespace issue. All tests are passing locally with this setup.Please have a look #2586.

roycaihw (approved):
/lgtm
/approve

roycaihw:
/close 
in favor of #2593

PR #2604, opened 2026-05-29
Restore bearer-token fallback in Configuration.auth_settings()

### LINKED ISSUE #2595: v36.0.0 regression: exec WebSocket returns 401 Unauthorized on AKS (works in v35.0.0)
### What happened?

After upgrading the `kubernetes` Python client from `35.0.0` to `36.0.0`, pod exec via `kubernetes.stream.stream()` consistently fails against an Azure AKS cluster with `401 Unauthorized` during the WebSocket handshake. Regular REST API calls (`v1.list_namespaced_pod`, `v1.read_namespaced_pod`, etc.) continue to work with the **same kubeconfig and same identity**. Downgrading the library to `35.0.0` immediately restores exec functionality with no other changes.

### What did you expect to happen?

`stream(v1.connect_get_namespaced_pod_exec, ...)` should succeed in v36.0.0 the same way it does in v35.0.0, against the same AKS cluster, with the same kubeconfig.

### How can we reproduce it (as minimally and precisely as possible)?

1. AKS cluster (Private Link), `kubectl exec` works fine from the same host.
2. kubeconfig uses an `exec:` auth plugin (`kubelogin` with Azure AD).
3. Run the following Python script with `kubernetes==36.0.0`:

```python
from kubernetes.stream import stream

cfg = client.Configuration()
cfg.verify_ssl = False
config.load_kube_config(config_file="/path/to/aks/kubeconfig", client_configuration=cfg)
cfg.proxy = None
v1 = client.CoreV1Api(api_client=client.ApiClient(cfg))

# REST: works in both v35 and v36
print([p.metadata.name for p in v1.list_namespaced_pod("vault").items])

# Exec WebSocket: works in v35, fails in v36 with 401
resp = stream(
    v1.connect_get_namespaced_pod_exec,
    "hashicorp-vault-0", "vault",
    command=["/bin/sh", "-c", "vault status"],
    stderr=True, stdin=False, stdout=True, tty=False,
)
print(resp)
```

### Anything else we need to know?

The traceback from v36.0.0:

```
File ".../kubernetes/stream/ws_client.py", line 552, in create_websocket
    websocket.connect(url, **connect_opt)
File ".../websocket/_handshake.py", line 158, in _get_resp_headers
    raise WebSocketBadStatusException(
websocket._exceptions.WebSocketBadStatusException: Handshake status 401 Unauthorized -+-+-
{'audit-id': '...', 'cache-control': 'no-cache, private', 'content-type': 'application/json',
 'date': '...', 'content-length': '129', 'connection': 'close'}
-+-+- b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure",
        "message":"Unauthorized","reason":"Unauthorized","code":401}\n'
```

The 401 response carries an `audit-id`, confirming the request reaches the AKS API server and is rejected at the auth layer (not a network/proxy issue).

#### Root cause (suspected)

PR #2486 (`implement pod exec websockets v5`) changed the advertised WebSocket subprotocols from a single value to a comma-separated list:

**`kubernetes/base/stream/ws_client.py` — v35.0.0:**

```python
header.append("sec-websocket-protocol: v4.channel.k8s.io")
```

**`kubernetes/base/stream/ws_client.py` — v36.0.0:**

```python
header.append("sec-websocket-protocol: %s,%s" % (V5_CHANNEL_PROTOCOL, V4_CHANNEL_PROTOCOL))
# → "sec-websocket-protocol: v5.channel.k8s.io,v4.channel.k8s.io"
```

The `Authorization` header handling is identical between the two versions, and the `Authorization` header **is** sent. However, when the AKS API server receives multiple subprotocols in `Sec-WebSocket-Protocol`, it appears to take a different auth code path (presumably the browser-style `base64url.bearer.authorization.k8s.io.<token>` lookup inside the subprotocol header) and fails to fall back to honoring the standard `Authorization` header, returning 401 instead of negotiating down to `v4.channel.k8s.io`.

This makes the regression appear specifically against cluster/control-plane versions that don't natively support the v5 subprotocol.

### Workaround

Pinning `kubernetes==35.0.0` resolves the issue.

### Kubernetes client version

```console
$ pip show kubernetes | grep Version
Version: 36.0.0
```

### Server version (AKS)

`*.privatelink.<region>.azmk8s.io` (Azure AKS, control plane managed by Azure; happy to share exact Kubernetes minor if needed)

### OS

Red Hat Enterprise Linux 9.x (containerized Python 3.12 environment)

### Python version

```console
Python 3.12.11
```

### Python client version

```
kubernetes==36.0.0
websocket-client==1.9.0
```

### Installation method

```
pip3 install kubernetes
```

### Related

- PR #2486 — implement pod exec websockets v5
- KEP-4006 — Transition SPDY to WebSockets

GK-07:

Restores backward compatibility for the bearer-token lookup in
`Configuration.auth_settings()`.

Before v36.0.0, the generated client stored the bearer token under
`api_key['authorization']`. v36.0.0 changed the lookup to
`api_key['BearerToken']` without a fallback, so any caller that still
writes the token under `'authorization'` ends up sending requests
without an `Authorization` header and gets `401 Unauthorized` from the
API server.

The user-visible symptom is "everything that used to work with my
bearer-token kubeconfig now returns 401 against the same cluster after
upgrading to v36.0.0", on both REST and WebSocket exec paths.

This PR keeps the new `'BearerToken'` key as the primary lookup but
adds `'authorization'` as a fallback alias — using the
`alias=` parameter already supported by `get_api_key_with_prefix`. No
mutation of `self.api_key`, no new branches in the hot path.

This matches @yliaog's suggestion in #2595 to carry a patch on
`Configuration.auth_settings()` for a transition period while existing
user code migrates from `'authorization'` to `'BearerToken'`.

Fixes #2595

- Applied symmetrically to the sync (`kubernetes/client/configuration.py`)
  and async (`kubernetes/aio/client/configuration.py`) variants.
- New regression tests in `kubernetes/test/test_api_client.py` cover all
  four cases: `BearerToken` alone, `authorization` alone, both set
  (precedence), and neither set.
- The `kubernetes/client/` and `kubernetes/test/` trees are generated;
  once this is merged, the patch SHA can be added to
  `scripts/apply-hotfixes.sh` so it survives the next regeneration.

```release-note
Restored backward compatibility for `Configuration.auth_settings()`:
the legacy `api_key['authorization']` lookup is honored as a fallback
when `api_key['BearerToken']` is not set, fixing 401 Unauthorized
regressions seen after upgrading to v36.0.0 (#2595).
```

### DIFF
--- a/kubernetes/aio/client/configuration.py
+++ b/kubernetes/aio/client/configuration.py
@@ -415,13 +415,19 @@ conf = client.Configuration(
         :return: The Auth Settings information dict.
         """
         auth = {}
-        if 'BearerToken' in self.api_key:
+        # Backward compatibility: prior to v36.0.0 the bearer token was
+        # stored under the 'authorization' api_key. Continue to honor it
+        # so user code that sets ``config.api_key['authorization']``
+        # directly keeps working with the new 'BearerToken' key the
+        # generated client now looks for.
+        # See: https://github.com/kubernetes-client/python/issues/2595
+        if 'BearerToken' in self.api_key or 'authorization' in self.api_key:
             auth['BearerToken'] = {
                 'type': 'api_key',
                 'in': 'header',
                 'key': 'authorization',
                 'value': await self.get_api_key_with_prefix(
-                    'BearerToken',
+                    'BearerToken', alias='authorization',
                 ),
             }
         return auth
--- a/kubernetes/client/configuration.py
+++ b/kubernetes/client/configuration.py
@@ -406,13 +406,19 @@ conf = client.Configuration(
         :return: The Auth Settings information dict.
         """
         auth = {}
-        if 'BearerToken' in self.api_key:
+        # Backward compatibility: prior to v36.0.0 the bearer token was
+        # stored under the 'authorization' api_key. Continue to honor it
+        # so user code that sets ``config.api_key['authorization']``
+        # directly keeps working with the new 'BearerToken' key the
+        # generated client now looks for.
+        # See: https://github.com/kubernetes-client/python/issues/2595
+        if 'BearerToken' in self.api_key or 'authorization' in self.api_key:
             auth['BearerToken'] = {
                 'type': 'api_key',
                 'in': 'header',
                 'key': 'authorization',
                 'value': self.get_api_key_with_prefix(
-                    'BearerToken',
+                    'BearerToken', alias='authorization',
                 ),
             }
         return auth
--- a/kubernetes/test/test_api_client.py
+++ b/kubernetes/test/test_api_client.py
@@ -78,3 +78,45 @@ class TestApiClient(unittest.TestCase):
             # test
             client = kubernetes.client.ApiClient(configuration=config)
             self.assertEqual( expected_pool, type(client.rest_client.pool_manager) )
+
+
+class TestConfigurationAuthSettings(unittest.TestCase):
+    """Regression tests for Configuration.auth_settings() bearer-token lookup.
+
+    Prior to v36.0.0 the generated client stored the bearer token under
+    ``api_key['authorization']`` (e.g. set by ``load_kube_config`` or by
+    user code directly). v36.0.0 switched the lookup to
+    ``api_key['BearerToken']`` without a fallback, which silently dropped
+    the Authorization header from every outgoing request and caused 401
+    Unauthorized against any cluster relying on bearer tokens.
+    See: https://github.com/kubernetes-client/python/issues/2595
+    """
+
+    def _bearer_value(self, config):
+        settings = config.auth_settings()
+        self.assertIn('BearerToken', settings)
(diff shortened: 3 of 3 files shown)


yliaog:
thanks for the PR

/lgtm
/approve

yliaog:
> [Keywords](https://help.github.com/articles/closing-issues-using-keywords) which can automatically close issues and hashtag(#) mentions are not allowed.

could you please fix the commit message?

GK-07:
Thanks @yliaog! Fixed the commit message, replaced `Fixes #2595` with a plain URL reference (no closing keyword / no hashtag) per the bot's check. Tree is byte-identical, only the message changed. Could you re-apply `/lgtm` when you have a moment?

yliaog (approved):

PR #2612, opened 2026-06-07
Fix pod log stream follow detection

### LINKED ISSUE #2610: Cannot use Watch().stream with read_namespaced_pod_log

`Watch.stream()` raises `ApiTypeError` when used with `CoreV1Api.read_namespaced_pod_log`:

```
kubernetes.client.exceptions.ApiTypeError: Got an unexpected keyword argument 'watch' to method read_namespaced_pod_log
```

Till a recent update this has been working. I feel the reason is that the watch module defines `PYDOC_FOLLOW_PARAM = ":param bool follow:"` but that format is no longer used as it is now split into 2 lines

```:param follow: Follow the log stream of the pod. Defaults to false.
:type follow: bool
```

```bash
kind create cluster
pip install kubernetes>=36.0.0
python reproducer.py
kind delete cluster
```

`reproducer.py`:

```python
"""Reproducer: Watch.stream() fails with read_namespaced_pod_log on kubernetes >= 36.0.0"""

import time

from kubernetes import client, config, watch

NAMESPACE = "default"
POD_NAME = "watch-bug-reproducer"

config.load_kube_config()
v1 = client.CoreV1Api()

pod = client.V1Pod(
    metadata=client.V1ObjectMeta(name=POD_NAME),
    spec=client.V1PodSpec(
        containers=[
            client.V1Container(
                name="busybox",
                image="busybox",
                command=["sh", "-c", "while true; do echo hello from pod; sleep 1; done"],
            )
        ]
    ),
)

print(f"Creating pod {POD_NAME}...")
v1.create_namespaced_pod(namespace=NAMESPACE, body=pod)

print("Waiting for pod to be running...")
while True:
    pod_status = v1.read_namespaced_pod_status(name=POD_NAME, namespace=NAMESPACE)
    if pod_status.status.phase == "Running":
        break
    time.sleep(1)

print("Pod is running. Attempting Watch.stream() with read_namespaced_pod_log...")
w = watch.Watch()
for line in w.stream(v1.read_namespaced_pod_log, name=POD_NAME, namespace=NAMESPACE):
    print(f"Got log line: {line}")
    break
```

I hope the fix is as simple as updating `PYDOC_FOLLOW_PARAM` but I am not sure if there is anything else to be considered.

**Environment**:
- OS Fedora 43:
- Python version 3.14
- Python client version `master` and 36.0.2

goutamadwant:
**What type of PR is this?**
/kind bug
/kind regression

**What this PR does / why we need it:**
Watch.stream decides whether to pass watch= or follow= by looking at the wrapped API method docstring.

The generated pod log method now documents follow as split :param follow: and :type follow: bool lines instead of the older :param bool follow: form. Because the watch code only recognized the older form, it passed watch=True to read_namespaced_pod_log, and the generated client rejected that kwarg.

This keeps the existing old-format check and also recognizes the generated split follow docstring, so pod log streaming uses follow=True while normal watch APIs continue to use watch=True.

**Which issue(s) this PR fixes:**
Fixes #2610

**Special notes for your reviewer:**
Added a regression test for the generated pod log docstring format. This is scoped to the Watch.stream argument selection path and does not change generated client files.

Checked locally with:

```
python3 -m unittest kubernetes.watch.watch_test.WatchTests.test_watch_for_follow kubernetes.watch.watch_test.WatchTests.test_watch_for_follow_with_split_param_type_docstring
python3 -m unittest kubernetes.base.watch.watch_test
python3 -m pytest -q kubernetes/base/watch/watch_test.py
python3 -m unittest discover
git diff --check
```
**Does this PR introduce a user-facing change?**
Yes - Watch.stream(v1.read_namespaced_pod_log, ...) now uses follow=True for pod log streaming with the current generated docstring format, instead of failing with an unexpected watch argument.

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:**
NONE

### DIFF
--- a/kubernetes/base/watch/watch.py
+++ b/kubernetes/base/watch/watch.py
@@ -19,7 +19,7 @@ import pydoc
 from kubernetes import client
 
 PYDOC_RETURN_LABEL = ":rtype:"
-PYDOC_FOLLOW_PARAM = ":param bool follow:"
+PYDOC_FOLLOW_PARAMS = (":param bool follow:", ":param follow:")
 
 # Removing this suffix from return type name should give us event's object
 # type. e.g., if list_namespaces() returns "NamespaceList" type,
@@ -111,7 +111,8 @@ class Watch:
         return return_type
 
     def get_watch_argument_name(self, func):
-        if PYDOC_FOLLOW_PARAM in pydoc.getdoc(func):
+        doc = pydoc.getdoc(func)
+        if any(param in doc for param in PYDOC_FOLLOW_PARAMS):
             return 'follow'
         else:
             return 'watch'
--- a/kubernetes/base/watch/watch_test.py
+++ b/kubernetes/base/watch/watch_test.py
@@ -19,6 +19,7 @@ from unittest.mock import Mock, call
 
 from kubernetes import client, config
 from kubernetes.client import ApiException
+from kubernetes.client.exceptions import ApiTypeError
 
 from .watch import Watch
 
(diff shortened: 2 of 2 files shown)


tamird:
I think https://github.com/kubernetes-client/python/pull/2616 fixes the same issue without increasing complexity

yliaog:
https://github.com/kubernetes-client/python/pull/2616 merged, I think we can close this PR

/close

PR #2616, opened 2026-06-20
Fix Watch follow parameter detection

### LINKED ISSUE #2615: Watch misclassifies generated pod log methods
/kind bug

Kubernetes 36.0.2 generates the `read_namespaced_pod_log` parameter
documentation as:

```
:param follow: Follow the log stream of the pod.
:type follow: bool
```

`Watch.get_watch_argument_name` still looks for the older
`:param bool follow:` form. It therefore classifies this as a resource watch
and injects `watch=True` instead of `follow=True`. The generated client then
raises `ApiTypeError` because `watch` is not accepted by the pod log method.

The existing `test_watch_for_follow` fixture also uses the obsolete docstring
shape, so it does not catch the regression.

tamird:
#### What type of PR is this?

/kind bug

OpenAPI-generated methods now document parameter names and types on separate
lines. `Watch` still looked for the previous combined `:param bool follow:`
form, so it sent `watch=True` to pod log methods. Those methods reject the
unexpected keyword instead of streaming logs.

Match the generated `:param follow:` form and update the existing test fixture
to use the current generated docstring shape.

Fixes #2615

None.

```release-note
Fix Watch.stream selecting watch instead of follow when streaming pod logs.
```

```docs

```

### DIFF
--- a/kubernetes/base/watch/watch.py
+++ b/kubernetes/base/watch/watch.py
@@ -19,7 +19,7 @@ import pydoc
 from kubernetes import client
 
 PYDOC_RETURN_LABEL = ":rtype:"
-PYDOC_FOLLOW_PARAM = ":param bool follow:"
+PYDOC_FOLLOW_PARAM = ":param follow:"
 
 # Removing this suffix from return type name should give us event's object
 # type. e.g., if list_namespaces() returns "NamespaceList" type,
--- a/kubernetes/base/watch/watch_test.py
+++ b/kubernetes/base/watch/watch_test.py
@@ -204,7 +204,8 @@ class WatchTests(unittest.TestCase):
 
         fake_api = Mock()
         fake_api.read_namespaced_pod_log = Mock(return_value=fake_resp)
-        fake_api.read_namespaced_pod_log.__doc__ = ':param bool follow:\n:rtype: str'
+        fake_api.read_namespaced_pod_log.__doc__ = (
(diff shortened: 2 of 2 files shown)


yliaog:
@goutamadwant could you review this PR?

goutamadwant (approved):
Thanks, this looks good to me.

This covers the same regression I tried to address in #2612, but does it with the smaller change by updating the existing fixture to the current generated docstring shape. I agree this is the cleaner approach.

One small non-blocking note: this also addresses #2610, which was the original report linked from #2612. It may be worth linking that issue too, or marking it as a duplicate of #2615.

opokornyy:
Hey @tamird, are you planning to move this PR forward, or should we implement a workaround instead?

tamird:
> Hey @tamird, are you planning to move this PR forward, or should we implement a workaround instead?

@opokornyy is there action required from me? Happy to make changes; were any requested?

opokornyy:
> > Hey @tamird, are you planning to move this PR forward, or should we implement a workaround instead?
> 
> @opokornyy is there action required from me? Happy to make changes; were any requested?

No changes were requested on my end. @roycaihw, would you mind taking a look when you have a chance? Thanks!

yliaog:
thanks for the PR

opokornyy:
Hey @yliaog, is there an ETA for when this will be released?

yliaog:
@opokornyy I can cut another release v36.0.3 this week.

PR #2622, opened 2026-07-03
Fix readline_channel crash and None return on default timeout

### LINKED ISSUE #2621: readline_channel crashes with OverflowError when using the default timeout

`readline_channel` (and `readline_stdout`/`readline_stderr`) default to `timeout=None`, which is converted to `float("inf")`. With no buffered line it calls `update()`, which multiplies the timeout by 1000 and passes it to `select.poll().poll()`. poll expects integer milliseconds and raises on infinity:

    OverflowError: timestamp out of range for platform time_t

When a finite timeout expires, the method also falls through and returns None instead of the empty string/bytes the other read methods return.

The default (no timeout) blocks until a line is available without crashing, and an expired timeout returns '' (or b'' in binary mode).

    import select
    select.poll().poll(float("inf"))   # OverflowError

**Environment**:
- Kubernetes version (`kubectl version`): n/a (client-side)
- OS (e.g., MacOS 10.13.6): any platform where select.poll exists (Linux, macOS)
- Python version (`python --version`): 3.12
- Python client version (`pip list | grep kubernetes`): master

archy-rock3t-cloud:
#### What type of PR is this?
/kind bug

With the default timeout, readline_channel converted None to float("inf") and crashed in update() when it reached select.poll().poll(). An infinite timeout now maps to a blocking poll, and an expired finite timeout returns '' or b'' like the other read methods instead of None.

Fixes #2621

```release-note
Fix readline_channel, readline_stdout and readline_stderr crashing with OverflowError when using the default timeout, and returning None when a timeout expires.
```

### DIFF
--- a/kubernetes/base/stream/ws_client.py
+++ b/kubernetes/base/stream/ws_client.py
@@ -128,6 +128,7 @@ class WSClient:
                 return b"" if self.binary else ""
 
             self.update(timeout=(timeout - time.time() + start))
+        return b"" if self.binary else ""
 
     def write_channel(self, channel, data):
         """Write data to a channel."""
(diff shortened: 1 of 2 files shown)


yliaog:
thanks for the PR

/lgtm
/approve

PR #2631, opened 2026-07-14
Regenerate the synchronous client with the modern Python generator

tamird:
#### What type of PR is this?

/kind feature

The synchronous client is still generated with OpenAPI Generator 6.6's
`python-legacy` backend. OpenAPI Generator removed that backend in v7,
so the current generation path cannot move to a supported generator or
inherit the modern Python generator's native type information.

This PR moves the synchronous client to the modern `python` backend:

- regenerate the client and Sphinx API index with Pydantic models and
  modern generated type information;
- declare the generated runtime requirements and package `py.typed`;
- adapt dynamic, stream, watch, OIDC, YAML, and example helpers to the
  modern generated API;
- preserve Kubernetes IntOrString, CRD JSON-schema, and Namespace-delete
  representations under Pydantic validation;
- preserve dict, JSON Patch, and generated-model PATCH bodies; and
- remove obsolete post-generation hotfixes while retaining the small
  body-aware PATCH media-type correction.

The generated client has 768 models, 65 API modules, and 1,872
API-call validators. Loading all of them during import makes
short-lived callers substantially slower. Generated packages now use
the generator's built-in lazy imports, and validators are built on
first use. Measured cold import falls from 1.19s to 0.16s while public
exports, constructor signatures, first-call validation, and importlib
package metadata are preserved.

No single tracked issue. The regeneration fixes existing urllib
compatibility, typing, and bytes-handling gaps in the synchronous
client.

The paired generator migration is
https://github.com/kubernetes-client/gen/pull/305. It selects the
released OpenAPI Generator v7.24.0, which contains the required legacy
compatibility and API-key-prefix fixes. The generated provenance records
the release tag and resolved commit 203d4867.

This changes synchronous generation only. The async client remains on
its existing generator path and should be regenerated separately; until
then, `six` remains an async requirement.

The changelog documents the synchronous runtime dependency changes,
stricter model and API-call validation, removed low-level transport
interfaces, decoded exception bodies, and Namespace-delete return type.
Existing endpoint names, model names, and wire aliases are preserved.

Legacy `async_req=True` still returns an `ApplyResult` consumed with
`.get()`, and `_preload_content=False` returns the raw response.
Ordinary return annotations intentionally describe the default call:
Watch reflects them to select event models. The generator rationale is
https://github.com/OpenAPITools/openapi-generator/pull/24300.

The initial release render was produced by the normal
`scripts/update-client.sh` Docker path in
https://github.com/tamird/kubernetes-python/actions/runs/29785873376.
Until the paired generator PR merges, reproduce it with that generator
branch checked out locally:

```
GEN_ROOT=/path/to/kubernetes-client-gen scripts/update-client.sh
```

The previous head passed the fork's Python 3.10-3.14 lint and test
matrix, but the upstream E2E run exposed valid IntOrString values and
Namespace-delete responses rejected by the generated validators, along
with invalid scalar exec arguments and lowercase `resourceversion`
fixtures. The replacement run exposed one remaining discovery 415: the
modern selector sent literal `*/*` where the legacy client defaulted to
JSON. This update corrects those cases and adds focused regressions.

The current upstream Python 3.10-3.14 validation and Python 3.10-3.12
E2E matrices are waiting for maintainer approval:
https://github.com/kubernetes-client/python/actions/runs/29936539412
https://github.com/kubernetes-client/python/actions/runs/29936538730
The matching exact-head fork Python 3.10-3.14 validation matrix passes:
https://github.com/tamird/kubernetes-python/actions/runs/29936531993.

```release-note
The synchronous Python client is regenerated with the modern OpenAPI
Generator Python backend. See CHANGELOG for backward-incompatible
changes.
```

#### Additional documentation:

```docs

```


penguinolog:
> Which issue(s) this PR fixes:

technically ton of (urllib compatibility, typing, several broken places with `bytes` processing)

requirements.txt
```diff
@@ -1,10 +1,12 @@
 certifi>=14.05.14 # MPL
 six>=1.9.0  # MIT
```
penguinolog: six is needed only for async after merge. Regenerate too as next PR?

tamird: Yes. This draft intentionally migrates only synchronous generation; async should be regenerated separately. `setup.py` reads `requirements.txt`, while the release package adds `requirements-asyncio.txt`, so I removed `six` from the synchronous requirements and left its async declaration in place.

tamird:
Right; this is not tied to one tracked issue. I updated that section to name the urllib compatibility, typing, and bytes-handling gaps instead of saying `None`.

kubernetes/.openapi-generator/VERSION
```diff
@@ -1 +1 @@
-6.6.0
\ No newline at end of file
+7.24.0-SNAPSHOT
```
penguinolog: it's not released, IMHO released version should be used instead

tamird: Yes, that's why this is a draft. It requires some changed that are not yet released.

tamird: Agreed. The three prerequisite generator fixes have merged, but they are not released yet. I updated the draft description to make that boundary explicit: this stays inspection-only until client-gen can pin the first released version containing them and regenerate canonically.

tamird: Done. OpenAPI Generator v7.24.0 is released and contains the required compatibility fixes. I regenerated this branch from the release via kubernetes-client/gen#305; the checked-in VERSION is now 7.24.0 and the resolved generator commit is 203d4867. The canonical Docker render is https://github.com/tamird/kubernetes-python/actions/runs/29785873376.

yliaog:
@tamird thanks for the PR, do you know whether the import latency issue reported in https://github.com/tomplus/kubernetes_asyncio/issues/293 is resolved with the new openapi generator?

yliaog:
could you also please take a look at the failure tests? https://github.com/kubernetes-client/python/actions/runs/29850863235/job/88758260386?pr=2631

tamird:
@yliaog Both are addressed in 5353a3d5, with the paired generator update in https://github.com/kubernetes-client/gen/pull/305.

The E2E failures were real. The modern client was validating Kubernetes `IntOrString` fields as dictionaries and Namespace DELETE responses as `Status`, so valid rolling-update percentages, service/probe ports, and terminating Namespaces failed. I also corrected the scalar exec arguments and invalid lowercase `resourceversion` fixtures. The remaining conflicts and `UnboundLocalError` failures were cascades. Focused regressions now cover those cases and pass locally.

Import latency is improved, though not eliminated. On the same Python 3.12 setup, the published legacy client imports in about 0.2s and the v7.24 render took about 3s. Deferring the 1,872 Pydantic call validators (supported since Pydantic 2.11, which this client already requires) reduces the render to about 1.1s while retaining first-call argument validation.

The replacement upstream E2E run is waiting for approval:
https://github.com/kubernetes-client/python/actions/runs/29874378845

Could you approve that run? The exact-head fork validation matrix is also running:
https://github.com/tamird/kubernetes-python/actions/runs/29874376044

tamird:
@yliaog The replacement E2E run reduced the failures to one discovery-time 415. The modern selector sent a literal `Content-Type: */*` for DynamicClient discovery GETs; the legacy selector translated that to `application/json`, which those routes require. This is fixed in 11812558, with a request-level regression. The apply PATCH itself remains `application/apply-patch+yaml` with the correct body and field manager.

Could you approve the new E2E run?
https://github.com/kubernetes-client/python/actions/runs/29878983770

yliaog:
are their breaking / backward incompatible changes we should document in CHANGELOG?

tamird:
@yliaog Yes. I added a dedicated breaking-change section, following the
earlier v6.6 precedent:
https://github.com/kubernetes-client/python/blob/581eeaccdb0ebab83bcaffbb8e3f44cfd8037a3f/CHANGELOG.md#L1-L33

It covers the synchronous dependency floors, Pydantic construction and
call validation, removed configuration and low-level transport APIs,
decoded exception bodies, and the Namespace-delete return type.

The compatibility audit also found an avoidable regression: existing
examples pass generated models directly as PATCH bodies, but the modern
signature rejected them before sending a request. The paired generator
update now accepts dict, JSON Patch, and generated-model bodies; a
request-level regression verifies the strategic-merge media type and
wire aliases. The focused generator and client suites pass locally.

Could you approve the replacement upstream runs?
https://github.com/kubernetes-client/python/actions/runs/29883910007
https://github.com/kubernetes-client/python/actions/runs/29883910002

The paired generator run also needs approval:
https://github.com/kubernetes-client/gen/actions/runs/29883743388

yliaog:
@tamird it seems that the import latency is still very big, 19s vs 0.8s, could you please take a look?

$ git checkout pr-2631
$ time PYTHONPATH=. python -c "import kubernetes; print(kubernetes.__file__)"
real    0m19.764s
user    0m18.475s
sys     0m1.302s

$ git checkout master
$ time PYTHONPATH=. python -c "import kubernetes; print(kubernetes.__file__)"
real    0m0.816s
user    0m0.559s
sys     0m0.280s

scripts/apply-hotfixes.sh
```diff
@@ -1,103 +0,0 @@
...
-if [ $? -eq 0 ]
-then
-    echo Successfully patched changes for bearer-token fallback in Configuration.auth_settings
-else
-    echo Failed to patch changes for bearer-token fallback in Configuration.auth_settings
-    git restore --staged .
-    exit 1
-fi;
-
-git commit -m "Apply hotfixes"
```
yliaog, line 103: why is this file removed? i think the hotfixes are still needed after upgrading openapi generator to v7.24.0

tamird:
@yliaog Fixed in bfe130db, with the paired generator update in https://github.com/kubernetes-client/gen/pull/305.

The generated client, API, and model packages now use OpenAPI Generator's built-in lazy imports instead of eagerly loading 768 models and 65 API modules. Pydantic call validators still build on first use, and package metadata remains available to `importlib`.

On the same Python 3.12 setup, cold `import kubernetes` now takes **0.16–0.19s**, down from **1.19s** on the previous PR head. Public imports, constructor signatures, recursive models, and first-call validation are preserved. The focused suites pass under both Pydantic 2.11 and 2.13, and the exact-head Python 3.10–3.14 fork validation matrix passes:
https://github.com/tamird/kubernetes-python/actions/runs/29936531993

The new upstream validation and E2E runs, and the paired generator run, need maintainer approval:
https://github.com/kubernetes-client/python/actions/runs/29936539412
https://github.com/kubernetes-client/python/actions/runs/29936538730
https://github.com/kubernetes-client/gen/actions/runs/29936513495

— tamirdex

scripts/release.sh
```diff
@@ -211,15 +211,6 @@ scripts/update-client.sh
...
 scripts/update-client-asyncio.sh
 
-# Apply hotfixes
-rm -r kubernetes/test/
-git add .
-git commit -m "temporary generated commit"
-scripts/apply-hotfixes.sh
-git reset HEAD~2
-# Apply proxy config after hotfixes
-scripts/insert_proxy_config.sh
```
yliaog, line 221: is this fix no longer needed?

tamird: Correct: this removed block was the release-only workaround for applying the same v6.6-era cherry-picks. It staged a temporary generated commit, ran `apply-hotfixes.sh`, rewound two commits, and then injected proxy settings into generated code.

Both synchronous and asyncio regeneration still run immediately above this block. Custom-object PATCH behavior, bearer-token fallback, proxy/`no_proxy` environment handling, the legacy `apis` alias, and client cleanup are already present in the generated or retained code, so neither the temporary cherry-picks nor proxy insertion is needed. The exact-head Python 3.10–3.14 validation and Python 3.10–3.12 E2E matrices both pass.

— tamirdex

scripts/apply-hotfixes.sh
tamird: The fixes are still present; the removed script is only the old mechanism for cherry-picking v6.6-specific commits after generation.

- Custom-object PATCH defaults to merge patch, with explicit JSON Patch still supported and covered by request-level tests.
- `no_proxy`, proxy environment variables, and proxy bypass are handled by the v7.24 generated `Configuration`/REST client; the existing proxy cases and a real request-level bypass test pass.
- Bearer-token fallback remains in both synchronous and asyncio `Configuration.auth_settings()`; the existing regression tests pass.
- The deprecated `kubernetes.client.apis` alias, API-client context-manager/thread-pool cleanup, and informer/metrics tests are all retained.

The old script would reapply commits against the previous generator layout, while these behaviors are now either generated directly or already checked in. The API-client, informer, and metrics suites pass 55 tests, and the exact-head upstream validation and E2E runs are green.

— tamirdex

scripts/apply-hotfixes.sh
```diff
@@ -1,103 +0,0 @@
...
-else
-    echo Failed to patch changes for enabling from kubernetes import apis
-    git restore --staged .
-    exit 1
-fi;
-
-# Patching commits for Client Context Manager
-# UPDATE: OpenAPI generator v4.3.0 has the context manager as a functionality. Cherry-picking just the tests for completeness.
-# Ref: https://github.com/kubernetes-client/python/pull/1073
-git cherry-pick -X no-renames -n 13dffb897617f87aaaee247095107d7011e002d5
```
yliaog, line 57: the two tests added in this cherry-pick are not perserved. 

test_context_manager_closes_threadpool
test_atexit_closes_threadpool

are they still useful after the upgrade?

scripts/apply-hotfixes.sh
```diff
@@ -1,103 +0,0 @@
...
-then
-    echo Successfully patched changes for no_proxy support
-else
-    echo Failed to patch changes for no_proxy support
-    git restore --staged .
-    exit 1
-fi;
-
-# Patching commit for test_api_client.py, test_informer and test_metrics
-git cherry-pick -n b5f4b4cc504e7f10956b52de6aa676e22c8952f0 d8c380e8bdef99e14c7a0e5a8ee216af370a10d8
```
yliaog, line 82: test_deserialize_dict_syntax_compatibility added in d8c380e8bdef99e14c7a0e5a8ee216af370a10d8 is not preserved.

scripts/apply-hotfixes.sh
```diff
@@ -1,103 +0,0 @@
...
-then
-    echo Successfully patched changes for test_api_client.py, test_informer and test_metrics
-else
-    echo Failed to patch changes for test_api_client.py, test_informer and test_metrics
-    git restore --staged .
-    exit 1
-fi;
-
-# Patching commit for bearer-token fallback in Configuration.auth_settings
-git cherry-pick -n 5621a4c2cf5cc278e8f45ce759ce44c057f1dbe2
```
yliaog, line 93: this patches the BearerToken alias for both sync and async client, i think the openapi generator v7.24.0 already adds the alias for sync client, does it add it also for async client? if it does, then we need to follow up with upgrading the async client to v7.24.0, otherwise, the async client generated by the v6.6.0 for the next release will not have this patch.

scripts/release.sh
yliaog: ok, the following code is generated by openapi generator v7.24.0. we can resolve this.

        # urllib3 does not read proxy environment variables itself:
        # https://github.com/urllib3/urllib3/issues/1785
        if proxy is None or no_proxy is None:
            proxies = getproxies()
            if proxy is None:
                scheme = urlparse(self.host).scheme
                proxy = proxies.get(scheme) or proxies.get("all")
            if no_proxy is None:
                no_proxy = proxies.get("no")

scripts/apply-hotfixes.sh
tamird: Yes. Restored both lifecycle regressions in 7e155ac3. The context-manager test verifies the pool is released; the atexit test invokes only this client’s registered callback instead of running unrelated process-wide exit handlers. Both pass against the v7.24 client.

— tamirdex

scripts/apply-hotfixes.sh
tamird: Restored in 7e155ac3. The regression now verifies both flat and nested Dict[...] deserialization against the v7.24 client without resurrecting the dict(str, str) syntax that v6.6 already stopped supporting.

— tamirdex

scripts/apply-hotfixes.sh
tamird: You’re right: the checked-in async client retained the fallback, but its next v6.6 regeneration would have lost it when the hotfix script disappeared. Fixed in 7e155ac3 by adding the authorization alias to the existing asyncio configuration patch, so canonical regeneration preserves bearer-token fallback and prefix handling without upgrading the async client. Async regressions also verify that BearerToken retains precedence.

— tamirdex

yliaog:
Now the import latency is on par

$ git checkout master
$ time python -c "import kubernetes.client.api_client; print(kubernetes.client.api_client.__file__)"
real    0m0.615s
user    0m0.444s
sys     0m0.195s

$ git checkout pr-2631 
$ time python -c "import kubernetes.client.api_client; print(kubernetes.client.api_client.__file__)"

real    0m0.599s
user    0m0.490s
sys     0m0.128s

yliaog:
@tamird thanks for the PR, could you help with upgrading the async client too?

/lgtm
/approve

tamird:
@yliaog The merged change upgrades the synchronous client; the asyncio client still uses the separate v6.6.0 python-asyncio generator. 7e155ac3 preserves async bearer-token fallback and prefix handling through future regeneration, so the synchronous upgrade does not break asyncio. Upgrading asyncio itself to v7.24.0 is a separate generator and compatibility migration, not part of this merged PR. The paired synchronous generator update is still in https://github.com/kubernetes-client/gen/pull/305.

— tamirdex

tamird:
@yliaog yes I will look into upgrading the async generator!

tamird:
Following up on your request to upgrade the async client: the shared generator migration is https://github.com/kubernetes-client/gen/pull/305, and the regenerated async client and breaking-change documentation are https://github.com/kubernetes-client/python/pull/2647.

— tamirdex

yliaog:
thanks, i'll take a look.

tomplus:
Hi, I'm trying to understand this step 

> Fixed in https://github.com/kubernetes-client/python/commit/bfe130db716c58b14bfc7122e7fa80e4d049a14c, with the paired generator update in https://github.com/kubernetes-client/gen/pull/305.

(https://github.com/kubernetes-client/python/pull/2631#issuecomment-5048568396)

As it was committed via a force-push, I can't see the diff. What change caused such a big speed improvement?

tamird:
> Hi, I'm trying to understand this step
> 
> > Fixed in [bfe130d](https://github.com/kubernetes-client/python/commit/bfe130db716c58b14bfc7122e7fa80e4d049a14c), with the paired generator update in [kubernetes-client/gen#305](https://github.com/kubernetes-client/gen/pull/305).
> 
> ([#2631 (comment)](https://github.com/kubernetes-client/python/pull/2631#issuecomment-5048568396))
> 
> As it was committed via a force-push, I can't see the diff. What change caused such a big speed improvement?

https://github.com/kubernetes-client/gen/pull/305/changes#diff-c793fca58c37cbba87ce8aa71cd873051b5e7d692c2f3a154b385cae8a2322d2R125-R144

tomplus:
Thanks for the clarification, `lazy-imports` does the job.

For the record, here is the ongoing discussion https://github.com/OpenAPITools/openapi-generator/issues/21831

yliaog:
thanks @tomplus and @tamird, this is valid concern. I think we have a few options:

1/ stick with the v6.6 OpenAPI generator
2/ Keep the v7.24 OpenAPI generator, without using lazy-imports
3/ Keep the v7.24 OpenAPI generator, and laze-imports, make it configurable, optional

What do you think?

tamird:
@yliaog @tomplus I think there is a fourth option. OpenAPI Generator v7.24 does not require lazy-imports; the client explicitly enables the generator’s lazyImports option. Going back to v6.6 or disabling lazy loading would forfeit the measured startup improvement, while making the dependency optional would leave generated code importing it whenever laziness is enabled.

I’m preparing and validating a generator-source change that implements lazy model and API exports with Python’s standard-library module __getattr__ and importlib, as described in https://peps.python.org/pep-0562/. That would keep v7.24 and lazy loading, remove lazy-imports and its generated dependency metadata, and eliminate the sys.modules/module-metadata workaround while addressing the supply-chain concern in https://github.com/OpenAPITools/openapi-generator/issues/21831. Once validated, I’ll update the merged synchronous client and the synchronous and asyncio generator changes from the same upstream source.

— tamirdex

yliaog:
totally, the 4th option as you suggested is preferred, that addresses the supply-chain concern, and the model import latency at the same time. thanks for looking into that.

tamird:
@yliaog @tomplus The dependency-free fourth option is now available at https://github.com/OpenAPITools/openapi-generator/pull/24402. It keeps the v7.24 generator and deferred model/API imports while replacing lazy-imports and the sys.modules workaround with Python’s standard library. Once the generator fix is accepted, I’ll regenerate the merged synchronous client and the synchronous and asyncio generator changes from the same source.

— tamirdex

tamird:
@yliaog @tomplus The standard-library lazy-import option agreed in this thread has now landed upstream in https://github.com/OpenAPITools/openapi-generator/pull/24402.

The synchronous generator is https://github.com/kubernetes-client/gen/pull/305 and its regenerated client is https://github.com/kubernetes-client/python/pull/2652. The independent asyncio generator is https://github.com/kubernetes-client/gen/pull/306 and its regenerated client is https://github.com/kubernetes-client/python/pull/2647. The asyncio change uses its own generator configuration; the external python-asyncio.sh and its v6 XML remain unchanged.

All four changes pin the same merged upstream source and truthfully record 7.25.0-SNAPSHOT. GitHub is waiting for maintainer approval of the exact fork runs listed at https://github.com/kubernetes-client/gen/pull/305#issuecomment-5093621171. Could you approve the runs and review the synchronous generator first?

[tamirdex]

kubernetes/base/stream/stream.py
```diff
@@ -13,37 +13,61 @@
 # limitations under the License.
 
 import functools
+from types import SimpleNamespace
 
 from . import ws_client
 
 
 def _websocket_request(websocket_request, force_kwargs, api_method, *args, **kwargs):
-    """Override the ApiClient.request method with an alternative websocket based
```
Ark-kun, line 21: The information about "Override" was valuable. The code here introduces satiny-defying crashes when used in async or multi-threaded scenarios. So, having a comment is sueful.

yliaog: @Ark-kun could you open a PR to add a comment to clarify?

tamird: Perhaps better would be to fix this mechanism, which looks inherently unsound.

Issue #2427, opened 2025-07-28
Windows contributors face multiple issues: symlink folders not portable, permission errors, and missing imports in tests


PraveenMudalgeri:
**Summary**

Trying to contribute as a Windows developer to this repo turned up several significant issues:

1. Several folders (e.g., `kubernetes/config`, `kubernetes/watch`) are symlinks to `base/` subdirectories in the repo. On Windows _and_ when using ZIP downloads from GitHub, these appear as plain text files and break both `pip install -e .` and running tests, since setuptools expects actual directories.

2. At least one test file (`watch_test.py` under `kubernetes/base/watch` and duplicated under `kubernetes/watch`) is missing `import json`. This causes test failures with `NameError: name 'json' is not defined`.

3. Several tests (for example, `test_oidc_with_refresh` inside `kube_config_test.py`) fail on Windows with `PermissionError` when trying to create/write temp files.

**Steps to Reproduce**

- Clone or download the repo on Windows.
- Try `pip install -e .` or run `python -m pytest`.
- Observe errors related to config/watch packages not being directories, import errors in test files, and permission errors for temp files.

**Proposed Fixes**

- Replace symlinks (`kubernetes/config`, `kubernetes/watch`, etc.) with real directories or use another structure to support Windows and ZIP-based contributors.
- Ensure all test modules have necessary imports (e.g., `import json`).
- Refactor or skip the failing temp file tests on Windows, or improve documentation about required permissions.

**Environment**

- OS: Windows 10/11, Python 3.13
- (Include `pip list` or `pip freeze` if desired)

**Thank you! I’m happy to give more details or to help test solutions.**

roycaihw:
The symlink comes from places like https://github.com/kubernetes-client/python/blob/master/kubernetes/config. I'm not familiar with Windows development. Need some Windows expert to input here.

roycaihw:
> Replace symlinks (kubernetes/config, kubernetes/watch, etc.) with real directories or use another structure to support Windows and ZIP-based contributors.

The way we are doing it at the moment is legacy, and I'm concerned that there can be users depending on the current structure. I'd like to see if we can find a less impactful fix for Windows specifically.

roycaihw:
> Ensure all test modules have necessary imports (e.g., import json).

SGTM. Would you like to send a fix?

> Refactor or skip the failing temp file tests on Windows, or improve documentation about required permissions.

It would also be great if we can find the correct alternative for Windows on temp file testing.

[...]

PR #2644, opened 2026-07-21
Fix OIDC token refresh failing on Windows when idp-certificate-authority-data is set

ealeonraz:
**What this does**

`KubeConfigLoader._refresh_oidc` wrote the IdP CA certificate by creating a `tempfile.NamedTemporaryFile(delete=True)` and then re-opening it by name. On Windows that second open raises `PermissionError` (the file is held exclusively while open), so OIDC token refresh fails at runtime for any kubeconfig using `idp-certificate-authority-data`. This is the temp-file portion of #2427.

The fix reuses the module's existing `_create_temp_file_with_content` helper (mkstemp-based, closed before reuse, cleaned up via the existing atexit handler) — the same mechanism this file already uses for other inline certificate data, so behavior is now consistent across platforms and cleanup no longer depends on GC timing.

**Testing**

`pytest kubernetes/base/config/kube_config_test.py` on Windows 11: 75 passed — including `test_oidc_with_refresh`, which fails with `PermissionError` before this change. No test modifications were needed; the existing tests cover the path once the production code is fixed.

Ref #2427 — planned as the first of a short series of small Windows fixes discussed there.

```release-note
Fixed OIDC token refresh failing with PermissionError on Windows when the kubeconfig uses idp-certificate-authority-data.
```

### DIFF
--- a/kubernetes/base/config/kube_config.py
+++ b/kubernetes/base/config/kube_config.py
@@ -383,16 +383,14 @@ class KubeConfigLoader:
         config = Configuration()
 
         if 'idp-certificate-authority-data' in provider['config']:
-            ca_cert = tempfile.NamedTemporaryFile(delete=True)
-
             cert = base64.b64decode(
                 provider['config']['idp-certificate-authority-data']
             ).decode('utf-8')
 
-            with open(ca_cert.name, 'w') as fh:
-                fh.write(cert)
-
-            config.ssl_ca_cert = ca_cert.name
+            # Use the shared temp-file helper instead of reopening a
+            # NamedTemporaryFile by name, which fails with PermissionError
+            # on Windows while the original handle is still open.
+            config.ssl_ca_cert = _create_temp_file_with_content(cert)
 
         elif 'idp-certificate-authority' in provider['config']:
             config.ssl_ca_cert = provider['config']['idp-certificate-authority']


yliaog:
does asyncio client have the same problem?

https://github.com/kubernetes-client/python/blob/master/kubernetes/aio/config/kube_config.py

ealeonraz:
Good catch, yes. The async client has the same underlying problem, though the code shape differs slightly.

`kubernetes/aio/config/kube_config.py` does not re-open the file for writing, so it avoids the exact traceback in this PR. But it writes through the open handle and then passes `certfile.name` to `OpenIDRequestor` while the `NamedTemporaryFile` is still open. On Windows a second open of that path fails, so any consumer reading the CA cert by path hits the same error.

Verified on Windows 11 / Python 3.14:

```
reopen-for-write while open   -> PermissionError   (sync client, this PR)
read-by-name while open       -> PermissionError   (async client pattern)
```

The fix is symmetric: `kubernetes/aio/config/kube_config.py` already defines `_create_temp_file_with_content` (line 87), the same mkstemp-based helper used here, so the async path can use it the same way.

One note on scope: I confirmed the temp-file semantics directly, but I have not exercised the async OIDC refresh end to end, since that needs a live OIDC provider. The failure is reachable whenever the kubeconfig supplies `idp-certificate-authority-data`.

Happy to fix the async client in this PR, or as a separate follow-up if you would rather keep this one narrow. Which do you prefer?

yliaog:
thanks, please open another PR for the async case.

/lgtm
/approve

PR #2646, opened 2026-07-22
Fix pending SSL data handling in stream client

### LINKED ISSUE #2414: stream api will lose data when use sslsocket

mihirduvedi:
#### What type of PR is this?

/kind bug

`WSClient.update()` can wait on `poll()` or `select()` even when an SSL socket already has decrypted data buffered internally. This can delay stream output indefinitely when no additional network event arrives.

This change checks `SSLSocket.pending()` before polling and immediately processes a WebSocket frame when buffered SSL data is available. It also adds a deterministic regression test using `timeout=None` that verifies neither polling path is entered.

Fixes #2414

This addresses the unit-test request from #2422 and the review feedback to avoid calling `poll()` or `select()` when SSL data is already pending.

Tests:

- `python -m pytest -q kubernetes/base/stream/ws_client_test.py` — 17 passed
- `python -m pytest -q kubernetes/base --ignore=kubernetes/base/dynamic/test_client.py --ignore=kubernetes/base/dynamic/test_discovery.py` — 143 passed

The two excluded dynamic-client modules require a live Kubernetes cluster.

```release-note
Fixed stream operations over SSL to process buffered WebSocket data without waiting for another socket event.
```

```docs
NONE
```

### DIFF
--- a/kubernetes/base/stream/ws_client.py
+++ b/kubernetes/base/stream/ws_client.py
@@ -214,7 +214,9 @@ class WSClient:
         #                   efficient as epoll. Will work for fd numbers above 1024.
         # select.epoll()  - newest and most efficient way of polling.
         #                   However, only works on linux.
-        if hasattr(select, "poll"):
+        if self.sock.is_ssl() and self.sock.sock.pending() > 0:
+            r = [self.sock.sock]
+        elif hasattr(select, "poll"):
             poll = select.poll()
             poll.register(self.sock.sock, select.POLLIN)
             if timeout is not None and timeout != float("inf"):
--- a/kubernetes/base/stream/ws_client_test.py
+++ b/kubernetes/base/stream/ws_client_test.py
@@ -18,6 +18,7 @@ from unittest.mock import MagicMock, patch
 from . import ws_client as ws_client_module
 from .ws_client import get_websocket_url, WSClient, V5_CHANNEL_PROTOCOL, V4_CHANNEL_PROTOCOL, CLOSE_CHANNEL, STDIN_CHANNEL
 from .ws_client import websocket_proxycare
+from .ws_client import STDOUT_CHANNEL
 from kubernetes.client.configuration import Configuration
 import os
 import socket
@@ -192,6 +193,7 @@ class WSClientProtocolTest(unittest.TestCase):
             mock_ws = MagicMock()
             mock_ws.subprotocol = V5_CHANNEL_PROTOCOL
             mock_ws.connected = True
+            mock_ws.is_ssl.return_value = False
             mock_ws.sock.fileno.return_value = 10
 
             # Setup frame with close signal for channel 0
@@ -216,6 +218,7 @@ class WSClientProtocolTest(unittest.TestCase):
             mock_ws = MagicMock()
             mock_ws.subprotocol = V4_CHANNEL_PROTOCOL
             mock_ws.connected = True
+            mock_ws.is_ssl.return_value = False
             mock_ws.sock.fileno.return_value = 10
 
             # Setup frame that looks like close signal but should be treated as data
@@ -315,6 +318,7 @@ class WSClientProtocolTest(unittest.TestCase):
             mock_ws = MagicMock()
             mock_ws.subprotocol = V5_CHANNEL_PROTOCOL
             mock_ws.connected = True
+            mock_ws.is_ssl.return_value = False
             mock_ws.sock.fileno.return_value = 10
             mock_create.return_value = mock_ws
 
@@ -350,6 +354,7 @@ class WSClientProtocolTest(unittest.TestCase):
             mock_ws = MagicMock()
             mock_ws.subprotocol = V5_CHANNEL_PROTOCOL
             mock_ws.connected = True
+            mock_ws.is_ssl.return_value = False
             mock_ws.sock.fileno.return_value = 10
             mock_create.return_value = mock_ws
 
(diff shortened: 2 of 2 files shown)


roycaihw:
/assign

@mihirduvedi Can you also check if the asyncio client needs the same fix? https://github.com/kubernetes-client/python/tree/master/kubernetes/aio

mihirduvedi:
Thanks for the review. I checked the asyncio client carefully. The reported issue is fixed in the synchronous client, and the asyncio client does not require the same change.

kubernetes/aio/stream/ws_client.py delegates WebSocket reads to aiohttp through async for msg in ws rather than polling the raw SSL socket. aiohttp drains decrypted data into its WebSocket parser and queues each complete frame.

I also tested the equivalent edge case over WSS by sending two channel frames in a single TLS transport write through the Kubernetes WsApiClient; both frames were received correctly.

Validation:

- Async stream tests: 5 passed
- Combined sync and async stream tests: 22 passed

Therefore, no asyncio source change is necessary.

mihirduvedi:
@roycaihw The failed GitHub Actions jobs stopped during setup because GitHub could not download the required actions (`Service Unavailable`); no tests were executed in those jobs. Could you rerun the failed jobs?

yliaog:
thanks for the PR
/lgtm
/approve

PR #2647, opened 2026-07-23
Regenerate the asyncio client with dependency-free lazy imports

tamird:
Regenerate `kubernetes.aio` with the OpenAPI Generator v7 `aiohttp` backend and dependency-free Python templates from https://github.com/OpenAPITools/openapi-generator/pull/24402. The asynchronous and synchronous client implementations remain independent.

Apply Kubernetes authentication, TLS, watch, WebSocket, and PATCH adaptations in the generator-owned postprocessing from https://github.com/kubernetes-client/gen/pull/306, and remove twelve obsolete downstream patches for the v6 client.

The generated metadata records `7.25.0-SNAPSHOT`; `v7.24.0` is the latest published release: https://github.com/OpenAPITools/openapi-generator/releases/tag/v7.24.0. The remaining prerequisite is merging the generator change so this client can be reproduced from upstream.

/kind feature

```release-note
Regenerate the asynchronous Kubernetes Python client with dependency-free lazy imports.
```

PR #2656, opened 2026-07-28
Restore typed synchronous Kubernetes deletion responses

tamird:
Restore the synchronous client’s `V1Status` deletion responses by reverting 1c5984c4d. https://github.com/kubernetes-client/gen/pull/306 restricts broader deletion schemas to the asyncio generator, making the synchronous output in https://github.com/kubernetes-client/python/pull/2647 inconsistent with its source.

Revert the synchronous generated APIs, documentation, checksum, and obsolete regression together. The asynchronous client and its deletion behavior are unchanged.

/kind bug

```release-note
Restored typed synchronous Kubernetes resource deletion responses.
```


yliaog:
thanks for the fix
/lgtm
/approve

Issue #2608, opened 2026-06-02
Improve Type Hints and Return Type Annotations in the Python Kubernetes Client


EmadDeve20:
Hello everyone,

I use this library extensively in my daily work, and one issue I frequently encounter is the lack of clear type hints and return type annotations in some parts of the client.

In many cases, it is difficult to determine the exact return type of methods or object attributes without manually inspecting the source code or debugging at runtime.

For example, while working with container statuses, I wrote code like this:

```python
pod = k8s_client.read_namespaced_pod(
    name=pod_name,
    namespace=namespace,
)

statuses: list[client.V1ContainerStatus] = (
    pod.status.container_statuses or []
)

target = next(
    (s for s in statuses if s.name == settings.TARGET_CONTAINER),
    None,
)

if not target:
    logger.warning("Container not found")
    return False

state: client.V1ContainerState = target.state

if state.running:
    logger.info("Container is running")
    return True

elif state.waiting:
    waiting: client.V1ContainerStateWaiting = state.waiting
    reason = waiting.reason

    if reason in ["OOMKilled", "Error"]:
        logger.error(f"Container waiting: {reason}")
        return False

elif state.terminated:
    terminated: client.V1ContainerStateTerminated = state.terminated
    reason = terminated.reason

    logger.error(f"Container terminated: {reason}")
    return False
```

One example of the confusion is this line:

```python
statuses = pod.status.container_statuses or []
```

Without proper typing support, it is not immediately obvious that `container_statuses` is expected to be `list[V1ContainerStatus] | None`.

It would be very helpful if more methods and model attributes included accurate Python type hints and return type annotations.

For example:

```python
def read_namespaced_pod(
    self,
    name,
    namespace,
    **kwargs
) -> V1Pod:
    ...
```

Better typing support would improve:

* IDE autocomplete
* Static analysis
* Developer experience
* Readability
* Easier debugging and maintenance

Especially for users working with tools like:

* mypy
* pyright
* pylance
* PyCharm

If the maintainers agree that this improvement would be valuable, I would be happy to work on it and contribute a fix.

Thank you for your work on this project.

penguinolog:
IMHO 99.9% of this issue will be automatically fixed with OpenAPI generator update, but it will be breaking change.

EmadDeve20:
Thanks for the comment! @penguinolog 

I completely agree that upgrading the OpenAPI Generator is probably the most effective way to fix the majority of type hint / return annotation issues automatically.

[...]

PR #2657, opened 2026-07-29
Mark the Kubernetes package as typed

tamird:
Python type checkers recognize a typed distribution only when `py.typed` is installed in its top-level package. The generated clients place markers under `kubernetes.client` and `kubernetes.aio.client`, so Pylance cannot discover the existing precise nested model annotations from the published `kubernetes` package.

Include `kubernetes/py.typed` in both the wheel and source distribution while retaining the generated client marker. Cover the synchronous and asynchronous `V1Pod.status`, `container_statuses`, and container-state type chains.

Addresses https://github.com/kubernetes-client/python/issues/2608#issuecomment-5122500530.

/kind bug

```release-note
Expose nested Kubernetes model type information to Python type checkers and IDEs.
```


setup.py
```diff
@@ -67,7 +67,10 @@
               'kubernetes.leaderelection.resourcelock',
               'kubernetes.informer',
               ],
-    package_data={'kubernetes.client': ['py.typed']},
+    package_data={
+        'kubernetes': ['py.typed'],
+        'kubernetes.client': ['py.typed'],
+    },
```
yliaog, line 73: setup-release.py also need this ?

tamird: @yliaog Yes—good catch. `setup-release.py` builds the combined synchronous and asyncio distribution, so updating only `setup.py` would leave published releases untyped.

Fixed in ca7fa344f. I built the release wheel and source distribution and confirmed that both include `kubernetes/py.typed` as well as the existing synchronous and asyncio client markers. The nested-type regressions pass for both clients.

[tamirdex]

gdlx:
Hi !
Will this fix be released soon ?
It's merged for one month but no version has been released since.
Using v36 with pyright is a mess but I confirm using master fixes almost all issues.
Thanks !

yliaog:
Yes, it will be released soon, after kubernetes 1.37 release.

gdlx:
Thanks @yliaog !
So it won't be released as a v36 patch ?
(I always keep the python lib in sync with my clusters and they won't be upgraded to 1.37 anytime soon...)

yliaog:
the typed hints are added from upgrading openapi generator to latest 7.25.0, which introduces breaking changes, that cannot be put into v36 patch release.

PR #2658, opened 2026-07-30
feat: add proxy_ssl_context to Configuration for a separate proxy TLS context

### LINKED ISSUE #2387: Proxy SSL Context

Ability to provide proxy SSL context when creating Configurtion object.

This would be useful in order to configure the TLS connection made to a proxy server, separate from the TLS connection made to the Kubernetes cluster.

This behavior, provided by urllib3, is described here: https://urllib3.readthedocs.io/en/stable/advanced-usage.html#http-and-https-proxies

Please see: _HTTPS Proxy + HTTPS Destination_

Ability to provide proxy SSL context when creating Configurtion object.

https://github.com/kubernetes-client/python/blob/master/kubernetes/client/rest.py#L99

Thanks in advance!

emmanuel-adu:
Fixes #2387.

`Configuration` gains `proxy_ssl_context: Optional[ssl.SSLContext] = None`, passed through to `urllib3.ProxyManager` in `RESTClientObject`. This lets the HTTPS proxy's TLS cert be validated independently of the destination (K8s API server) settings - the ["HTTPS Proxy + HTTPS Destination"](https://urllib3.readthedocs.io/en/stable/advanced-usage.html#https-proxy-https-destination) scenario urllib3 already supports natively.

`kubernetes/client` is generated from [kubernetes-client/gen](https://github.com/kubernetes-client/gen); companion PR kubernetes-client/gen#308 adds the same patch there. This PR applies it to the already-generated files so it's usable now.

**Verification:** `kubernetes/e2e_test/test_issue_2387_proxy_ssl_context.py` spins up a mock HTTPS proxy and destination server with independent self-signed CAs (no live cluster needed). Without `proxy_ssl_context`: fails with `CERTIFICATE_VERIFY_FAILED`. With it set to trust the proxy's CA: succeeds, destination still validated separately.

test report: both scenarios passing, with captured CERTIFICATE_VERIFY_FAILED log for the without-fix case

```release-note
Added `proxy_ssl_context` to `Configuration`, letting users provide a separate SSL context for HTTPS proxy TLS verification, independent of the destination TLS settings.
```


kubernetes/client/configuration.py
```diff
@@ -11,6 +11,7 @@
 
 
 import copy
+import ssl
```
roycaihw, line 14: This file is generated by upstream openapi-generator. Please send a PR to update the upstream generator to properly support this feature.

emmanuel-adu: Opened kubernetes-client/gen#308 to add `proxy_ssl_context` for the sync client, following the existing pattern, and this PR applies that generated change now.

Or are you asking for this to be implemented in OpenAPITools/openapi-generator itself instead?

yliaog: better to add to OpenAPITools/openapi-generator

modules/openapi-generator/src/main/resources/python-pydantic-v1/configuration.mustache

modules/openapi-generator/src/main/resources/python/configuration.mustache

emmanuel-adu: Okay we'll look into doing that

yliaog:
sounds good, thanks.

please add the corresponding e2e test for asyncio client.

kubernetes/client/configuration.py
emmanuel-adu: Opened and merged OpenAPITools/openapi-generator#24539 - adds `proxy_ssl_context` to both templates you flagged, gated the same way as `tls_server_name`/`disable_strict_ssl_verification`.

Given that, do we want to merge this PR now to ship the fix, or hold off and pick it up later via a full regen (bump the pinned generator commit in gen, rerun autoupdate.sh)? Either way works cleanly - if you go the regen route later, this PR's changes just get superseded by that regen with no conflict.

@yliaog @roycaihw

emmanuel-adu:
How would you like to proceed?

cc: @yliaog @roycaihw

> Opened and merged [OpenAPITools/openapi-generator#24539](https://github.com/OpenAPITools/openapi-generator/pull/24539) - adds `proxy_ssl_context` to both templates you flagged, gated the same way as `tls_server_name`/`disable_strict_ssl_verification`.
> 
> Given that, do we want to merge this PR now to ship the fix, or hold off and pick it up later via a full regen (bump the pinned generator commit in gen, rerun autoupdate.sh)? Either way works cleanly - if you go the regen route later, this PR's changes just get superseded by that regen with no conflict.
> 

>also comment from https://github.com/kubernetes-client/gen/pull/308#issuecomment-5169112057 @brendandburns

kubernetes/aio/client/configuration.py
```diff
@@ -238,6 +240,7 @@ def __init__(
         connection_pool_maxsize: Optional[int]=None,
         proxy: Optional[str]=None,
         proxy_headers: Optional[Any]=None,
+        proxy_ssl_context: Optional[ssl.SSLContext]=None,
```
yliaog: does it apply to aio client? the upstream openapi generator PR says it does not

yliaog:
@emmanuel-adu regen does not have the test, right? i think we want to merge this PR, please take a look at the question about aio, if it is needed.

kubernetes/aio/client/configuration.py
emmanuel-adu: Correct - OpenAPITools/openapi-generator#24539 only touches the `python` and `python-pydantic-v1` (urllib3/sync) templates, not the aiohttp templates, so a future regen wouldn't produce this for the aio client on its own.

The aio support in this PR (`configuration.py` field + the `TCPConnector` override in `rest.py`) is hand-written, not generator-templated, because aiohttp has no public equivalent to urllib3's `proxy_ssl_context` - it required overriding a private `TCPConnector` method to give the proxy leg its own SSL context.

So it's the one piece of this PR not backed by the generator - it'd need to be manually reapplied (or ported to an aiohttp template upstream) after any future regen. Happy to drop it and keep this PR scoped to the sync client if you'd rather not carry hand-maintained code here - your call.

emmanuel-adu:
@yliaog Correct - regen wouldn't add the e2e test on its own, e2e tests are hand-maintained in this repo, not templated by the generator (same as the rest of the test suite).

Replied inline on the aio question - the upstream generator PR doesn't touch the aiohttp templates, so the aio changes here are hand-written and not backed by regen. Left it to you whether to keep or drop that part.

yliaog:
@emmanuel-adu we can merge the first commit, could you please drop the other 3 commits in the PR? so we can focus on the fix for sync client.

emmanuel-adu:
@yliaog Done - dropped the asyncio commits, this PR is now just the sync client fix (`kubernetes/client/configuration.py`, `kubernetes/client/rest.py`, and the e2e test).

PR #2664, opened 2026-07-30
Align synchronous resource deletion responses

tamird:
/kind bug

Individual resource deletion may return either the deleted resource or a Status. Generate synchronous responses from the same corrected object schema already used by asyncio, preserving the full response without fabricating V1Status or stringifying structured resource status. Exercise both response shapes over real HTTP, update the Job example, and preserve background garbage collection in end-to-end coverage.

Compatibility: Relative to released v36, 58 synchronous individual-deletion methods and their HTTP-info variants now return dictionaries instead of V1Status. Update attribute access such as result.status to result.get("status"). Collection-delete methods and operations with an already resource-typed response are unchanged.

Generator: https://github.com/kubernetes-client/gen/pull/309
Maintainer discussion: https://github.com/kubernetes-client/gen/pull/306#discussion_r3686963670

```release-note
action required: Synchronous individual resource deletion methods that previously returned V1Status now return dictionaries containing the actual deleted resource or Status. Replace result.status with result.get("status").
```


yliaog:
thanks for the PR

/lgtm
/approve

PR #2665, opened 2026-07-31
Fix parameter name in ConfigMapLock.create docstring

LALITH0110:
#### What type of PR is this?

/kind documentation

`ConfigMapLock.create` documented `:param electionRecord:`, but the parameter is named `election_record`:

```python
def create(self, name, namespace, election_record):
    """
    :param electionRecord: Annotation string
```

So the docstring described a parameter that does not exist on the method, while the real one was left undocumented. This corrects the name to match the signature.

The change is in `kubernetes/base/leaderelection/resourcelock/configmaplock.py`. Note `kubernetes/leaderelection` is a symlink into `kubernetes/base/`, so there is a single real file and no second copy to update.

One line. Documentation only — no behaviour change.

None — found by comparing documented `:param` entries against actual function signatures.

The documented parameters on `create` now exactly match its signature (`name`, `namespace`, `election_record`). Nothing under `kubernetes/client/` was touched, so no generated code is affected, and no tests reference `leaderelection`.

```release-note
NONE
```

```docs

```

### DIFF
--- a/kubernetes/base/leaderelection/resourcelock/configmaplock.py
+++ b/kubernetes/base/leaderelection/resourcelock/configmaplock.py
@@ -73,7 +73,7 @@ class ConfigMapLock:
 
     def create(self, name, namespace, election_record):
         """
-        :param electionRecord: Annotation string
+        :param election_record: Annotation string
         :param name: Name of the configmap object to be created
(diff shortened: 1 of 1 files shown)


yliaog:
thanks for the PR
/lgtm
/approve

PR #2666, opened 2026-07-31
Fixed casing of PyYAML in requirements

### LINKED ISSUE #2663: change  requirements.txt to match case of PyYAML
When building OS level packages for python-kubernetes the lower case of pyyaml in the requirements.txt file results in a OS package dependency on lower case pyyaml, but when the python-pyyaml package is built, due to its mixed case, the provided package is a mixed case PyYAML. The automatic dependency checker of the OS package subsystem us unable to satisfy this dependency and installation fails.

If the requirements.txt file had the dependency
PyYAML>=6.0.3
the OS packaging would be seamless

p172913:
#### What type of PR is this?

/kind bug

Fixes casing of PyYAML in requirements

Fixes #2663

```release-note
```

```docs

```

### DIFF
--- a/requirements-asyncio.txt
+++ b/requirements-asyncio.txt
@@ -1,5 +1,5 @@
 python-dateutil>=2.8.2  # BSD
-pyyaml>=6.0.3  # MIT
+PyYAML>=6.0.3  # MIT
 urllib3>=2.7.0 # MIT
 aiohttp>=3.14.3,<4.0.0  # Apache-2.0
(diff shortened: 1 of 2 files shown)


yliaog:
please also update [requirements-asyncio.txt](https://github.com/kubernetes-client/python/blob/master/requirements-asyncio.txt)

yliaog:
thanks for the PR
/lgtm
/approve

PR #2668, opened 2026-08-05
Fix OIDC token refresh failing on Windows in the asyncio client

ealeonraz:
The asyncio client has the same Windows failure as #2644, in a different
form. `_refresh_oidc` writes through the open `NamedTemporaryFile` handle,
so it avoids the traceback in that PR, but it then passes `certfile.name`
to `OpenIDRequestor` while the handle is still open. On Windows a second
open of that path fails, so any consumer reading the CA cert by path hits
a PermissionError.

This applies the same fix. The aio module had no module-level
`_create_temp_file_with_content`, only a `FileOrData` method, so this adds
the module-level helper mirroring `kubernetes/base/config/kube_config.py`
and points the existing method at it, which also closes a descriptor that
`mkstemp` leaked.

Note the CA cert temp file now lives until process exit and is cleaned by
the existing `_cleanup_temp_files` atexit hook, matching the sync client,
rather than being removed when the `with` block ended.

Verified on Windows 11 / Python 3.14:

    new helper:   reopened by path OK, 59 bytes
    old pattern:  PermissionError [Errno 13] Permission denied

`kubernetes/aio/config/kube_config_test.py`: 61 passed.

Follow-up to #2644, requested by @yliaog.

```release-note
Fix OIDC token refresh failing on Windows in the asyncio client when idp-certificate-authority-data is set
```

### DIFF
--- a/kubernetes/aio/config/kube_config.py
+++ b/kubernetes/aio/config/kube_config.py
@@ -52,6 +52,26 @@ def _cleanup_temp_files():
     _temp_files = {}
 
 
+def _create_temp_file_with_content(content, temp_file_path=None, force_recreate=False):
+    if len(_temp_files) == 0:
+        atexit.register(_cleanup_temp_files)
+
+    # Because we may change context several times, try to remember files we
+    # created and reuse them at a small memory cost.
+    content_key = str(content)
+
+    if not force_recreate and content_key in _temp_files:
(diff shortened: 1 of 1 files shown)


yliaog:
thanks for the PR

/lgtm
/approve

PR #2670, opened 2026-08-05
Stop leading when onstarted_leading raises

### LINKED ISSUE #2075: leaderelection do not stop leading properly

LALITH0110:
#### What type of PR is this?

/kind bug

`onstarted_leading` is started on its own thread, so when it raises, the
exception is swallowed by that thread and the renew loop never learns about it.
The candidate keeps renewing its lease indefinitely even though the work the
lease is meant to protect has stopped, and `onstopped_leading` is never called.
Because the lease keeps being renewed, no other candidate can take over either.

This records the failure and checks it at the top of the renew loop, so the
candidate stops renewing and `run()` calls `onstopped_leading` — the same path
already taken when a renewal fails.

**Scope:** only the failure case reported in the issue changes. A callback that
returns normally behaves exactly as before, and none of the existing tests
needed to be modified.

Reproduced before fixing, with a lock that never refuses a renewal (so the only
thing that can end the leadership is the callback):

| | renewals after the callback died | `onstopped_leading` called |
|---|---|---|
| before | 8 (unbounded) | no |
| after | 1 | yes |

Fixes #2075

- The new test uses a `MockResourceLock` whose `renew_count_max` is
  `sys.maxsize`, so the renew loop can only exit because the callback failed —
  it fails on `master` (times out) and passes with this change. It runs the
  election on a daemon thread so a regression times out rather than hanging CI.
- `renew_loop` keeps its old zero-argument form via a default of `None`, so any
  existing caller is unaffected.
- Only exceptions are caught; `BaseException` is deliberately not caught, so
  `KeyboardInterrupt` and `SystemExit` keep their current behaviour.
- Verified locally on macOS (Python 3.12): full suite minus e2e is **1484
  passed, 42 skipped, 0 failed**; the leaderelection suite goes 4 -> 5 passed.
  `flake8 --select=E9,F63,F7,F82` is clean, and `pycodestyle` reports exactly
  the same pre-existing findings before and after this change.

```release-note
Leader election now stops leading and calls `onstopped_leading` when the `onstarted_leading` callback raises, instead of renewing the lease indefinitely. The exception is logged via the `leaderelection` logger.
```

```docs

```

### DIFF
--- a/kubernetes/base/leaderelection/leaderelection.py
+++ b/kubernetes/base/leaderelection/leaderelection.py
@@ -55,12 +55,23 @@ class LeaderElection:
             logger.info("{} successfully acquired lease".format(self.election_config.lock.identity))
 
             # Start leading and call OnStartedLeading()
-            threading.Thread(target=self.election_config.onstarted_leading, daemon=True).start()
+            leading_failed = threading.Event()
+            threading.Thread(target=self.run_onstarted_leading, args=(leading_failed,), daemon=True).start()
 
-            self.renew_loop()
+            self.renew_loop(leading_failed)
 
             # Failed to update lease, run OnStoppedLeading callback
             self.election_config.onstopped_leading()
+
+    def run_onstarted_leading(self, leading_failed):
+        # Run the callback in this thread, recording whether it raised. Without
+        # this the exception is swallowed by the worker thread and the lease keeps
+        # being renewed even though the work it protects is no longer running.
+        try:
+            self.election_config.onstarted_leading()
+        except Exception:
+            logger.exception("onstarted_leading raised an exception, stopping leading")
+            leading_failed.set()
 
     def acquire(self):
         # Follower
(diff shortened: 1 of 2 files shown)


yliaog:
please investigate the CI test failures

LALITH0110:
Sorry for the slow reply, and thanks for the retests.

Those failures were a GitHub Actions outage rather than anything in this PR — every failed job stopped at `Failed to resolve action download info: Service Unavailable / Bad Gateway` with only a ~35 line log, so pytest never actually started. The jobs that did get going passed, and everything is green now.

For what it's worth, locally the full suite minus e2e is 1484 passed / 42 skipped. The new test times out on master and passes with the fix, and none of the existing tests needed changing.

yliaog:
thanks for the PR

/lgtm
/approve

PR #2671, opened 2026-08-05
Add client-go-compatible Retry-After retries to generated clients

sttts:
#### What type of PR is this?

/kind feature

Wires the retry helpers from kubernetes-client/python#2634 into the generated sync and asyncio REST clients via repo-local post-generation patches, matching the Java-style approach instead of carrying Kubernetes-specific behavior in kubernetes-client/gen.

The generated clients gain opt-in `client_go_retries` support for `GET` and `HEAD` Retry-After responses. The retry ceiling follows `configuration.retries` when set and otherwise uses the client-go default. Sync uses urllib3 for normal retries and disables urllib3 status retries only for the client-go read wrapper; asyncio uses its async helper directly for that same read path.

None

```release-note
Added opt-in client-go-compatible Retry-After retries for generated Python and asyncio GET/HEAD requests.
```

```docs
NONE
```

#### Testing

- `git diff --check`
- `git apply --unidiff-zero --reverse --check scripts/client_go_retry_patch.diff`
- `git apply --unidiff-zero --reverse --check scripts/client_go_retry_asyncio_patch.diff`
- `git apply --unidiff-zero --check scripts/client_go_retry_patch.diff` against a fresh `origin/master` worktree
- `git apply --unidiff-zero --check scripts/client_go_retry_asyncio_patch.diff` against a fresh `origin/master` worktree
- `python3 -m py_compile ...` for the changed retry/rest/config/test modules
- `uv run --with-requirements requirements.txt --with-requirements requirements-asyncio.txt python -c "import kubernetes.client.rest; import kubernetes.aio.client.rest"`
- `uv run --with-requirements requirements.txt --with-requirements requirements-asyncio.txt python -m unittest -q ...` for the new targeted retry tests

### DIFF
--- a/kubernetes/aio/client/configuration.py
+++ b/kubernetes/aio/client/configuration.py
@@ -355,6 +355,22 @@ conf = client.Configuration(
         self.retries = retries
         """Retry configuration
         """
+        self.client_go_retries = False
+        """Enable Kubernetes client-go-compatible retry semantics.
+
+           When enabled, GET and HEAD requests retry Retry-After responses.
+           The retry ceiling is read from ``retries`` when set; otherwise it
+           follows the client-go default of at most 10 retries.
+        """
+        self.client_go_retry_backoff = None
+        """Backoff for Kubernetes client-go-compatible retries.
+
+           If unset, client-go-compatible GET and HEAD retries use the
+           client-go default retry ceiling with no additional client-side
+           delay beyond Retry-After. When set, ``retries`` still overrides
+           the retry ceiling if it is not None.
+        """
+
         self.trace_configs = trace_configs
         """aiohttp.TraceConfig list forwarded to ClientSession for tracing.
         """
--- a/kubernetes/aio/client/rest.py
+++ b/kubernetes/aio/client/rest.py
@@ -21,6 +21,11 @@ from typing import Any, Dict, Optional, Union
 import aiohttp
 import aiohttp_retry
 
+from kubernetes.aio.utils.retry import (
+    is_retry_after_response,
+    on_retry_after_error,
+    retry_after_backoff,
+)
 from kubernetes.aio.client.exceptions import ApiException, ApiValueError
 
 RESTResponseType = aiohttp.ClientResponse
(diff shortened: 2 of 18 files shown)


yliaog:
@sttts thanks for the PR, please take a look at the CI test failures

sttts:
@yliaog it's green now

yliaog:
please remove fixup! in the commit message

also investigate e2e test failure

kubernetes/aio/client/_retry.py
```diff
@@ -0,0 +1,61 @@
...
+    is_retry_after_response,
+    retry_after_backoff,
+    retry_after_seconds,
+)
+
+
+T = TypeVar("T")
+
+
+async def on_retry_after_error(
```
yliaog: why not use kubernetes/aio/utils/retry.py ?

kubernetes/client/_retry.py
```diff
@@ -0,0 +1,61 @@
...
+    is_retry_after_response,
+    retry_after_backoff,
+    retry_after_seconds,
+)
+
+
+T = TypeVar("T")
+
+
+def on_retry_after_error(
```
yliaog: why not use kubernetes/utils/retry.py ?

yliaog:
close to retrigger the CI tests
/close

sttts:
@yliaog have squashed to remove the fixup commit.

yliaog:
@sttts why do you need to add [kubernetes/aio/client/_retry.py](https://github.com/kubernetes-client/python/pull/2671/files#diff-ca05d0625984e25ddcb3f88b7721961d1f56e85d2bbeeeceb4defc912370cbf7), instead of using the util directly?

kubernetes/client/_retry.py
sttts: Good point. The generated sync REST client now imports `kubernetes.utils.retry` directly. I removed the `_retry.py` wrapper and fixed the import cycle with local imports at the actual use sites instead.

kubernetes/aio/client/_retry.py
sttts: Good point. The generated async REST client now imports `kubernetes.aio.utils.retry` directly. I removed the `_retry.py` wrapper and fixed the import cycle with local imports at the actual use sites instead.

sttts:
Updated this to use `kubernetes.utils.retry` and `kubernetes.aio.utils.retry` directly. The import cycle is handled with local imports at the actual use sites, so no lazy package exports are needed.

yliaog:
thanks for the PR
/lgtm
/approve

PR #2684, opened 2026-08-24
Regenerate both clients with openapi-generator 7.25.0

### LINKED ISSUE #2681: `HTTP_PROXY`/`HTTPS_PROXY` ignored by `load_kube_config()`: proxy scheme resolved before the host is known
Split out of #2679 at @yliaog's request — that issue covered the kubeconfig `proxy-url` field, fixed in #2680. This one is about the environment variables.

`HTTP_PROXY` and `HTTPS_PROXY` are ignored when the client configuration is built by `load_kube_config()`. `Configuration.proxy` stays `None` and requests go direct.

Only `ALL_PROXY` works. That makes the commonly suggested workaround for proxy setups ("just export `HTTPS_PROXY`") ineffective on the `load_kube_config()` path, which is the path almost every user takes.

`HTTPS_PROXY` (or `HTTP_PROXY`) is honoured for the cluster endpoint, as it is for `curl`, `kubectl`, and client-go.

```console
$ HTTPS_PROXY=http://127.0.0.1:8888 python -c "
print(repr(Configuration().proxy))
print(repr(Configuration(host='https://example:6443').proxy))"
None
'http://127.0.0.1:8888'
```

The same asymmetry end to end, against a cluster reachable only through a proxy (the setup from #2679):

| env var set | `Configuration.proxy` after `load_kube_config()` | API call |
|---|---|---|
| `HTTPS_PROXY` | `None` | fails |
| `HTTP_PROXY` | `None` | fails |
| `ALL_PROXY` | set | works |

**Root cause**

`Configuration.__init__` resolves the proxy from the environment by the scheme of `self.host`:

https://github.com/kubernetes-client/python/blob/9c3456a351594b59a19862d9caec1e78e7b4e287/kubernetes/client/configuration.py#L339-L347

```python
scheme = urlparse(self.host).scheme
proxy = proxies.get(scheme) or proxies.get("all")
```

But `host` is a property, not the constructor argument:

https://github.com/kubernetes-client/python/blob/9c3456a351594b59a19862d9caec1e78e7b4e287/kubernetes/client/configuration.py#L663-L666

```python
@property
def host(self) -> str:
    """Return generated host."""
    return self.get_host_from_settings(self.server_index, variables=self.server_variables)
```

and `server_index` is set to `0` whenever no `host` argument is passed:

https://github.com/kubernetes-client/python/blob/9c3456a351594b59a19862d9caec1e78e7b4e287/kubernetes/client/configuration.py#L247

```python
self.server_index = 0 if server_index is None and host is None else server_index
```

With `server_index = 0`, `host` returns `get_host_settings()[0]['url']`, which is `""` for this spec. So `urlparse("").scheme` is `''`, `proxies.get('')` misses, and only the `proxies.get("all")` fallback can match — which is why `ALL_PROXY` is the one variable that works.

`load_kube_config()` constructs the object with no `host`:

https://github.com/kubernetes-client/python/blob/9c3456a351594b59a19862d9caec1e78e7b4e287/kubernetes/base/config/kube_config.py#L781-L784

```python
if client_configuration is None:
    config = type.__call__(Configuration)
    loader.load_and_set(config)
```

so it always takes the broken path. The loader assigns the real server afterwards — and the `host` setter does reset `server_index` to `None` — but by then the proxy has already been resolved in `__init__`. Passing `host=` explicitly is what makes the working case above work.

**Possible fixes**

Two options, and I'd like guidance on which is preferred before writing a patch:

1. Re-evaluate the environment proxy in the loader once `host` is known — a change in `kubernetes/base/config/kube_config.py`, which is developer-written and patchable in this repo.
2. Fix the resolution order in `Configuration.__init__` so it uses the effective base path rather than the generated `host` property. This is in `kubernetes/client/configuration.py`, which is generated (`kubernetes/.openapi-generator/FILES`), so it would need whatever hotfix/regeneration flow you prefer.

Happy to implement either. Given #2147's history I'd rather not touch generated files without agreement first.

Related: #2679 / #2680 (kubeconfig `proxy-url`, fixed), #2520 (`no_proxy` env var handling).

**Environment**:
- Python client version: master (`9c3456a351594b59a19862d9caec1e78e7b4e287`), also reproduced on released 36.0.3
- Python version: 3.12.3
- OS: Linux

/kind bug

yurnov:
#### What type of PR is this?

/kind bug
/kind cleanup

Regenerates **both** clients with openapi-generator `v7.25.0` (released 2026-08-24), up from the `830e9d15` master snapshot. Sync and asyncio move together, as @yliaog asked on [#2681][c1].

The motivating fix is [#24716][p1]: `Configuration` resolved the environment proxy once in `__init__`, before `load_kube_config` assigns `host`. Because the scheme-specific proxy depends on the host, `HTTPS_PROXY` was silently ignored — at construction `host` is `""`, so `urlparse("").scheme` is `""` and neither `https` nor `http` matched. The proxy is now re-resolved when `host` is assigned, and an explicitly set `proxy` still wins.

Upstream changes that reach the generated code:

| PR | Effect | Sync | Asyncio |
|----|--------|:----:|:-------:|
| [#24716][p1] | Re-resolve the environment proxy when `host` is assigned — fixes #2681 | ✅ | — |
| [#24539][p2] | Native `proxy_ssl_context` support (requested in #2387) | ✅ | — |
| [#23992][p3] | Honour content-type parameters when preparing a request | ✅ | — |
| [#24357][p4] | Null handling in generated models | ✅ | ✅ |
| [#24359][p5] | Serialize booleans as lowercase in all parameter styles | ✅ | ✅ |
| [#24718][p6] | Deserialize `Optional[...]` instead of raising `AttributeError` | ✅ | ✅ |

[#24716][p1] and [#24539][p2] land only on the sync side — they change the urllib3-based templates, which the aiohttp-based asyncio client does not use.

[#24357][p4] accounts for the bulk of the diff and is a behaviour fix, not churn. The old template dropped falsy items entirely, in both list and map serialization:

```python
for _item in self.conditions:
    if _item:
        _items.append(_to_openapi_value(_item))
```

so a `0`, `""`, or `False` element silently vanished from the serialized output. It now emits `None` only for genuine `None`:

```python
_items.append(_to_openapi_value(_item) if _item is not None else None)
```

`scripts/swagger.json` is byte-identical, so every change here is generator-driven; no spec drift is mixed in.

[c1]: https://github.com/kubernetes-client/python/issues/2681#issuecomment-5345041942
[p1]: https://github.com/OpenAPITools/openapi-generator/pull/24716
[p2]: https://github.com/OpenAPITools/openapi-generator/pull/24539
[p3]: https://github.com/OpenAPITools/openapi-generator/pull/23992
[p4]: https://github.com/OpenAPITools/openapi-generator/pull/24357
[p5]: https://github.com/OpenAPITools/openapi-generator/pull/24359
[p6]: https://github.com/OpenAPITools/openapi-generator/pull/24718

Fixes #2681

Also delivers the `proxy_ssl_context` support requested in #2387 (already closed).

**Requires kubernetes-client/gen#310**, which pins `openapi/python.sh` and `openapi/python-aio.sh` to `v7.25.0`. That merged on 2026-08-25, so both trees here reproduce from a clean checkout using gen's current defaults.

**Where to look.** The diff is large (523 files over two commits) but almost entirely mechanical regeneration. The only hand-written changes are:

- `scripts/client_go_retry_patch.diff` and `scripts/client_go_retry_asyncio_patch.diff` — regenerated with context lines
- `scripts/rest_client_patch.diff` — regenerated with context lines
- `scripts/update-client.sh` and `scripts/update-client-asyncio.sh` — dropping `--unidiff-zero`

Everything else is generator output.

**The local patches needed refreshing, and one case is worth showing.** `rest_client_patch.diff` failed loudly, because [#23992][p3] rewrote the line it anchored on — that is the good outcome. `client_go_retry_patch.diff` did not. Its hunks were zero-context pure insertions (`@@ -359,0 +360,16 @@`), and with no context lines `git apply --unidiff-zero` has nothing to match — it inserts at the recorded offset unconditionally. This bump adds eight lines to `configuration.py` above that offset, so replaying the old patch on the new output gives:

```
$ git apply --unidiff-zero scripts/client_go_retry_patch.diff
$ echo $?
0
```

…while having inserted the block at line 360 instead of 368, in the middle of the `proxy_ssl_context` docstring:

```python
        self.proxy_ssl_context = proxy_ssl_context
        """SSL context used only for the TLS handshake with the proxy itself
        (e.g. an HTTPS CONNECT tunnel), independent of the destination TLS
        self.client_go_retries = False          # <-- landed here
```

Here the corruption happens to produce a `SyntaxError`, so it would be caught — but by `compileall` or the test run, pointing at generated code rather than at the patch that broke it. A smaller shift would just as easily produce valid code in the wrong scope.

Both patches are regenerated with normal context and `--unidiff-zero` is dropped from `update-client.sh` and `update-client-asyncio.sh`.

**On the asyncio patch specifically:** it still applies correctly on this bump, because the changes above it are confined to the urllib3 templates. Refreshing it is hardening rather than a fix — happy to drop that hunk of the change if you would rather keep it minimal.

**AI Assistance Disclosure**: This code was developed with the assistance of Claude Opus 5 (claude-opus-5). The author has reviewed, tested, and takes full responsibility for the logic and changes

#### Verification

All run in a container:

- `compileall` clean over `kubernetes/` (sync + asyncio)
- sync `kubernetes/test` + `kubernetes/base`: 246 passed, 14 skipped, 99 subtests
- asyncio `kubernetes/aio/test`: 951 passed, 6 subtests
- Both refreshed patches apply with no fuzz to pristine generated output and reproduce the committed files byte-for-byte
- Clean-room round-trip: fresh clone, both regeneration scripts run against gen master with `OPENAPI_GENERATOR_COMMIT` unset, gives zero diff outside `doc/` — so the tree is reproducible and no hand-maintained content is lost on regeneration
- #2681 confirmed end-to-end, with and without the environment set:

```
$ HTTPS_PROXY=... HTTP_PROXY=... NO_PROXY=.internal
host unset at construction : '' -> proxy None
after https host assigned  : 'http://secure-proxy.example:8888'
after http host assigned   : 'http://plain-proxy.example:8080'
no_proxy                   : '.internal'
explicit proxy survives    : 'http://explicit.example'
explicit None stays None   : None

$ (no proxy environment)
after https host assigned  : None
```

#### Follow-ups deliberately left out

- `rest.py` still compares the content type by exact string when coercing `application/json-patch+json` to strategic merge; with [#23992][p3] available that could move to `contenttype_matches()`, but it is a behaviour change and does not belong in a regeneration PR.

```release-note
Fixed `load_kube_config` ignoring the `HTTPS_PROXY`/`HTTP_PROXY` environment variables. The environment proxy is now re-resolved when `Configuration.host` is assigned, so it is selected against the actual cluster URL rather than an empty host. Both the synchronous and asyncio clients are regenerated with openapi-generator 7.25.0, which also adds `proxy_ssl_context` support, honours content-type parameters when preparing requests, serializes booleans in query parameters as lowercase, and no longer drops falsy items from serialized lists and maps.
```


yurnov:
This PR in status Draft until kubernetes-client/gen#310 is merged

yliaog:
thanks for the PR

https://github.com/kubernetes-client/gen/pull/310 has merged.

/lgtm
/approve

PR #2687, opened 2026-08-27
Use SPDX license identifier in package metadata

### LINKED ISSUE #2686: License metadata is not SPDX-compliant
The metadata of the Python package contain a license which is not SPDX-compliant, which is the 'new' standard for licensing Python packges (see [PEP 639](https://peps.python.org/pep-0639). It declares to be licensed as `Apache License Version 2.0`, whereas the SPDX-standard would require it to be `Apache-2.0` (https://spdx.org/licenses/).

Additionally, the license classifier `License :: OSI Approved :: Apache Software License` is deprecated and can be replaced as soon as `project.license` is SPDX-compliant (see [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-classifiers)).

Reproduced via:

```bash
❯ pip3 download --no-deps kubernetes
Collecting kubernetes
  Downloading kubernetes-36.0.3-py2.py3-none-any.whl.metadata (1.8 kB)
Downloading kubernetes-36.0.3-py2.py3-none-any.whl (4.6 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 4.6/4.6 MB 19.8 MB/s eta 0:00:00
Saved ./kubernetes-36.0.3-py2.py3-none-any.whl
Successfully downloaded kubernetes
❯ unzip -qo kubernetes-36.0.3-py2.py3-none-any.whl
❯ grep License kubernetes-36.0.3.dist-info/METADATA
License: Apache License Version 2.0 # <-- non-SPDX-compliant
Classifier: License :: OSI Approved :: Apache Software License # <-- deprecated
License-File: LICENSE
```

The deprecation of the license classifier is also visible in a warning when building the package, e.g. via `uv build .`:

```log
********************************************************************************
Please consider removing the following classifiers in favor of a SPDX license expression:

License :: OSI Approved :: Apache Software License

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.

********************************************************************************
```

The required diff to address both findings looks like:

```diff
diff --git a/setup.py b/setup.py
index 892dfa8ff..9afac786e 100644
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     description="Kubernetes python client",
     author_email="",
     author="Kubernetes",
-    license="Apache License Version 2.0",
+    license="Apache-2.0",
     url="https://github.com/kubernetes-client/python",
     keywords=["Swagger", "OpenAPI", "Kubernetes"],
     install_requires=REQUIRES,
@@ -79,7 +79,6 @@ setup(
         "Topic :: Utilities",
         "Intended Audience :: Developers",
         "Intended Audience :: Information Technology",
-        "License :: OSI Approved :: Apache Software License",
         "Operating System :: OS Independent",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
```

After applying the diff, the metadata looks like:

```bash
❯ grep License kubernetes-36.0.0+snapshot.dist-info/METADATA
License: Apache-2.0
License-File: LICENSE
```
I'm happy to provide a PR after approval!

chala2001:
#### What type of PR is this?

/kind cleanup

Sets the license metadata to the SPDX identifier `Apache-2.0` instead of the free
text `Apache License Version 2.0`, and drops the deprecated
`License :: OSI Approved :: Apache Software License` classifier, per PEP 639.

The diff in the issue covers `setup.py`. `setup-asyncio.py` and `setup-release.py`
carried the same two values, so I changed those too - otherwise the asyncio and
release builds would keep emitting the old metadata.

Checked the generated metadata with setuptools 84.0.0, before and after:

```
before:  License: Apache License Version 2.0
         Classifier: License :: OSI Approved :: Apache Software License

after:   License: Apache-2.0
```

One thing worth flagging: `setup(license="Apache-2.0")` still lands in the legacy
`License` field rather than PEP 639's `License-Expression`. The value is a valid
SPDX identifier now and the deprecated classifier is gone, which is what the issue
asks for, but a real `License-Expression` would need a `pyproject.toml` with
`[project] license`. Happy to do that as a follow-up if you want it.

Fixes #2686

Packaging metadata only, no code or behaviour changes.

```release-note
Package license metadata now uses the SPDX identifier `Apache-2.0`, and the deprecated Apache Software License classifier has been removed.
```


yliaog:
thanks for the PR
please follow up on pyproject.toml with [project] license

/lgtm
/approve

PR #2314, opened 2024-12-17
add support Lease-based leader election (rather than ConfigMaps) #1877

rvlane:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Enhance leaderelection module to support V1Lease based lease lock.

#### Which issue(s) this PR fixes:

Fixes #1877 

#### Special notes for your reviewer:

Have coded this such that the LeaderElectionRecord is unchanged to ensure backward compatibility with anyone using the existing structures.

#### Does this PR introduce a user-facing change?

```release-note
Support Lease-based leader election
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #2690, opened 2026-08-28
Add LeaseLock for leader election

### LINKED ISSUE #1877: add support Lease-based leader election (rather than ConfigMaps)

Currently, the K8S Python API only [supports leader election](https://github.com/kubernetes-client/python/tree/master/kubernetes/base/leaderelection) using [`ConfigMapLock`](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/leaderelection/resourcelock/configmaplock.py).

The K8S Go API supports using the proper [`Lease`](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/lease-v1/) resource type with its [`LeaseLock`](https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/resourcelock/leaselock.go), and actually recently [removed support for the legacy ConfigMap approach](https://github.com/kubernetes/client-go/commit/276ea3ed979947d7cdd4b3d708862245ddcd8883#diff-7c2833e4f05e0ea6d94d565760fccf36366514d09df4a1e8d3bbcbc8ef98babb).

We should update the Kubernetes Python API to provide a `Lease`-based leader election.

---

/cc @Invictus17 @tdihp

chala2001:
#### What type of PR is this?

/kind feature

Leader election can currently only use a `ConfigMap`. `client-go` moved to the `coordination.k8s.io` `Lease` resource and removed ConfigMap support, so this adds a `LeaseLock` next to `ConfigMapLock`. The example now uses it, with the ConfigMap line kept commented out.

Fixes #1877

This picks up @rvlane's #2314, which you reviewed with "LGTM overall. Could you add a test?" and which the triage robot later closed as rotten. The original work is credited with a `Co-authored-by` trailer. The tests you asked for are in `leaselock_test.py`.

Four things changed from that version while rebasing onto current master:

- **Logger.** The original called `logging.basicConfig(level=logging.INFO)` at import, which reconfigures the root logger of whatever application imports the library. This package moved to `logger = logging.getLogger("leaderelection")` in the meantime, so it uses that now.
- **Time formats.** `str(datetime)` leaves the microseconds out when they are exactly zero, and the original only accepted the `%f` spellings. Those times fell through and returned `None`, so the lease was written with a null `acquireTime`. Covered by `test_record_without_microseconds_survives_a_write_and_read`.
- **`lease_reference` is kept after `create`**, so a following `update` has something to work from. `update` also returns `False` rather than raising `AttributeError` when it is called before a `get`, which matches client-go's `lease not initialized` check.
- **UTC.** `leaderelection.py` builds its times with `datetime.fromtimestamp()`, which is naive local time. The original appended `Z` to it, which labels local wall clock as UTC and writes a timestamp that is wrong by the host's offset. Since the point of a Lease is that other clients read it, the times are now converted to the real UTC instant and converted back on read, so the election record still round trips unchanged. `test_times_are_written_as_the_real_utc_instant` pins that under `Asia/Kolkata`.

One thing I left out, happy to follow up if you would like it: `leaseTransitions` is neither read nor written, because `LeaderElectionRecord` has no field for it. client-go writes it on every update. Adding it would touch the shared record class and `ConfigMapLock` as well, so it seemed better as a separate change.

Verified: 17 tests pass under `kubernetes/base/leaderelection` (5 existing, 12 new), and reverting each of the four fixes above makes the matching test fail. `pycodestyle --max-line-length=100` is clean on both new files. I have not run the functional or e2e suites.

```release-note
Add LeaseLock for leader election, using the coordination.k8s.io Lease resource.
```

### DIFF
--- a/kubernetes/base/leaderelection/example.py
+++ b/kubernetes/base/leaderelection/example.py
@@ -16,6 +16,7 @@ import uuid
 from kubernetes import client, config
 from kubernetes.leaderelection import leaderelection
 from kubernetes.leaderelection.resourcelock.configmaplock import ConfigMapLock
+from kubernetes.leaderelection.resourcelock.leaselock import LeaseLock
 from kubernetes.leaderelection import electionconfig
 
 
@@ -42,8 +43,13 @@ def example_func():
 # A user can choose not to provide any callbacks for what to do when a candidate fails to lead - onStoppedLeading()
 # In that case, a default callback function will be used
 
+lock = LeaseLock(lock_name, lock_namespace, candidate_id)
+# Choose the lock to use. LeaseLock uses the coordination.k8s.io Lease
+# resource and is the recommended one; ConfigMapLock is kept for existing users.
+# lock = ConfigMapLock(lock_name, lock_namespace, candidate_id)
+
 # Create config
-config = electionconfig.Config(ConfigMapLock(lock_name, lock_namespace, candidate_id), lease_duration=17,
+config = electionconfig.Config(lock, lease_duration=17,
                                renew_deadline=15, retry_period=5, onstarted_leading=example_func,
                                onstopped_leading=None)
 
--- a/kubernetes/base/leaderelection/resourcelock/leaselock.py
+++ b/kubernetes/base/leaderelection/resourcelock/leaselock.py
@@ -1,0 +1,155 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from kubernetes.client.rest import ApiException
+from kubernetes import client
+from ..leaderelectionrecord import LeaderElectionRecord
+from datetime import datetime, timezone
+import logging
+logger = logging.getLogger("leaderelection")
+
+# Formats produced by str(datetime). The microsecond component is omitted
+# when it is exactly zero, so both spellings have to be accepted.
+TIME_FORMATS = (
+    "%Y-%m-%d %H:%M:%S.%f%z",
+    "%Y-%m-%d %H:%M:%S.%f",
+    "%Y-%m-%d %H:%M:%S%z",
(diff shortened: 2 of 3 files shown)


kubernetes/base/leaderelection/example.py
```diff
@@ -42,8 +43,13 @@ def example_func():
 # A user can choose not to provide any callbacks for what to do when a candidate fails to lead - onStoppedLeading()
 # In that case, a default callback function will be used
 
+# Choose the lock to use. LeaseLock uses the coordination.k8s.io Lease
+# resource and is the recommended one; ConfigMapLock is kept for existing users.
```
Invictus17, line 48: Please move this below line 48 just above # lock = ConfigMapLock(lock_name, lock_namespace, candidate_id).

Invictus17 (changes requested):

kubernetes/base/leaderelection/example.py
chala2001: Done, moved in 6073f7f. The comment block now sits directly above the commented-out ConfigMapLock line. Thanks for the review.

Invictus17 (approved):

chala2001:
@yliaog this one is ready when you get a chance. @Invictus17 reviewed it and approved, and I pushed the change he asked for in 6073f7f.

He isn't in OWNERS though, so Prow still shows it as not approved ,, it needs /lgtm and /approve from you or @fabianvf before it can go in..

kubernetes/base/leaderelection/resourcelock/leaselock.py
```diff
@@ -0,0 +1,155 @@
+# Copyright 2021 The Kubernetes Authors.
```
yliaog: s/s2021/2026/

kubernetes/base/leaderelection/resourcelock/leaselock_test.py
```diff
@@ -0,0 +1,208 @@
+# Copyright 2021 The Kubernetes Authors.
```
yliaog: s/2021/2026/

kubernetes/base/leaderelection/resourcelock/leaselock.py
```diff
@@ -0,0 +1,155 @@
...
+
+        return leader_election_record
+
+    def time_str_to_iso(self, str_time):
+        """Convert an election record time into the instant to store.
+
+        ``leaderelection.py`` builds its times with
+        ``datetime.fromtimestamp()``, which is local and naive. The Lease is
+        shared with other clients, so the value has to go on the wire as the
+        real UTC instant rather than as local wall clock labelled as UTC.
```
yliaog: s/labelled/labeled/

chala2001:
Fixed all three in 9b87f18 - both copyright years and the spelling. Thanks for the review.

yliaog:
thanks for the PR

/lgtm
/approve

Issue #1841, opened 2022-06-22
Unexpected WebSocketConnectionClosedException when streaming long running commands, missing pings?


hterik:
**What happened (please include outputs or screenshots)**:
After exactly 300sec of idle, exec commands streamed with `stream` loose their connection, even though 
* pod is still running,
* The process that was execd in container is still running (checked with `kubectl exec POD_NAME ps aux`)
* Network otherwise appears to be healthy.
* No warnings logged by kube-apiserver

Following stack trace is given
```
  File "kubernetes/stream/ws_client.py", line 141, in readline_stderr
    return self.readline_channel(STDERR_CHANNEL, timeout=timeout)
  File "kubernetes/stream/ws_client.py", line 104, in readline_channel
    self.update(timeout=(timeout - time.time() + start))
  File "kubernetes/stream/ws_client.py", line 192, in update
    op_code, frame = self.sock.recv_data_frame(True)
  File "websocket/_core.py", line 401, in recv_data_frame
    frame = self.recv_frame()
  File "websocket/_core.py", line 440, in recv_frame
    return self.frame_buffer.recv_frame()
  File "websocket/_abnf.py", line 338, in recv_frame
    self.recv_header()
  File "websocket/_abnf.py", line 294, in recv_header
    header = self.recv_strict(2)
  File "websocket/_abnf.py", line 373, in recv_strict
    bytes_ = self.recv(min(16384, shortage))
  File "websocket/_core.py", line 524, in _recv
    return recv(self.sock, bufsize)
  File "websocket/_socket.py", line 122, in recv
    raise WebSocketConnectionClosedException(
websocket._exceptions.WebSocketConnectionClosedException: Connection to remote host was lost. 
```

**What you expected to happen**:
* Connection is not lost
* If it is lost, connection is automatically reconnected (up to some configured timeout limit) 
   Reading https://github.com/websocket-client/websocket-client#long-lived-connection, there is a description on how to use a built in way from the websocket library to do this. It appears that the stream command is not using this method.

**How to reproduce it (as minimally and precisely as possible)**:
(Mostly inspired by https://github.com/kubernetes-client/python/blob/master/examples/pod_exec.py)
```py
from kubernetes.stream import stream
...
...
s = stream(
    _client.connect_get_namespaced_pod_exec,
    _pod.metadata.name,
    _pod.metadata.namespace,
    command=["sleep", "100000"],           
    async_req=False,
    stderr=True,
    stdin=False,
    stdout=True,
[...]

PR #2693, opened 2026-08-29
Add an opt-in TCP keepalive option to the sync client

### LINKED ISSUE #2067: Set tcp keepalive for http connections

Currently when a watch is opened to the kubernetes api we do not send regular keepalives on the connection.
That means that if the watch is not receiving any updates a reverse proxy in the path might terminate the tcp connection.

The go client is currently sending a tcp keepalive every 30 seconds: https://github.com/kubernetes/client-go/blob/release-1.27/transport/cache.go#L104-L107

On linux set a socket options to update the keepalive time, just like it is done in go here: https://go.dev/src/net/tcpsockopt_unix.go

I do not have particular feelings regarding other operating systems

chala2001:
#### What type of PR is this?

/kind feature

An idle watch sends nothing on its connection, so a proxy or load balancer in
the path is free to drop it, and the client only finds out on the next read.
client-go avoids this by dialling with a keepalive
([transport/cache.go](https://github.com/kubernetes/client-go/blob/master/transport/cache.go)),
while the Python client asks the kernel for nothing.

`Configuration.socket_options` already reaches urllib3, so the plumbing exists.
What is missing is a supported way to turn keepalive on without hand-writing
platform specific socket tuples.

This adds `Configuration.keep_alive`, off by default:

```python
configuration = client.Configuration()
configuration.keep_alive = True
```

The timings match what client-go actually asks the kernel for. Its
`Dialer.KeepAlive: 30 * time.Second` becomes the *idle* time, and Go leaves the
probe interval and count at its own defaults, so on Linux client-go ends up
with `TCP_KEEPIDLE=30`, `TCP_KEEPINTVL=15`, `TCP_KEEPCNT=9`. The helper uses
those three values.

Two details worth flagging:

- The list is built on top of urllib3's `default_socket_options` rather than
  replacing them. urllib3 swaps its defaults for whatever list it is handed
  instead of merging, so building on them keeps `TCP_NODELAY` in place.
- Each option is `hasattr` guarded. macOS spells the idle time `TCP_KEEPALIVE`
  and has no `TCP_KEEPIDLE`, so a bare reference would raise `AttributeError`
  at import time there.

`socket_options` still wins when set, so callers already passing their own list
are unaffected.

The layout follows #2671: the logic is hand-written in `kubernetes/utils/`, and
the generated client gets a small `scripts/keepalive_patch.diff` that
`update-client.sh` re-applies after regeneration.

Sync client only. The asyncio client stores `socket_options` but never reads it
- `_create_connector` builds an `aiohttp.TCPConnector`, which takes no socket
options - so parity there is a separate change.

Fixes #2067

`kubernetes/utils/keepalive_test.py` covers the timings, the preserved urllib3
defaults, the macOS fallback and the platform guard, and checks the pool
manager actually receives the options. Not covered: whether probes reach the
wire, which needs a real idle connection.

#2067 is cross-linked from #1841. That one is the websocket exec stream dying
after 300s idle, which needs ping frames in `ws_client.py` - a different
mechanism, not fixed here.

```release-note
Added an opt-in `keep_alive` option to `Configuration`, enabling TCP keepalive on the sync client's sockets with the same timings client-go uses.
```

### DIFF
--- a/kubernetes/client/configuration.py
+++ b/kubernetes/client/configuration.py
@@ -387,6 +387,15 @@ conf = client.Configuration(
         self.socket_options = socket_options
         """Options to pass down to the underlying urllib3 socket
         """
+        self.keep_alive = False
+        """Enable TCP keepalive on the underlying urllib3 sockets.
+
+           Long lived requests such as watches are otherwise dropped
+           silently by an idle proxy or load balancer. When enabled, the
+           client asks the kernel for the same keepalive timings client-go
+           uses. Ignored if ``socket_options`` is set, which takes
+           precedence.
+        """
 
         self.datetime_format = datetime_format
         """datetime format
--- a/kubernetes/client/rest.py
+++ b/kubernetes/client/rest.py
@@ -27,6 +27,7 @@ from kubernetes.utils.retry import (
     on_retry_after_error,
     retry_after_backoff,
 )
+from kubernetes.utils.keepalive import tcp_keepalive_socket_options
 from kubernetes.client.exceptions import ApiException, ApiValueError
 
 SUPPORTED_SOCKS_PROXIES = {"socks5", "socks5h", "socks4", "socks4a"}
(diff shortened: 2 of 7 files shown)


chala2001:
Hi @yliaog, when you get a chance could you take a look at this one? It's opt-in and off by default, so nothing changes for anyone who doesn't set it. Happy to adjust anything.

yliaog:
thanks for the PR

/lgtm
/approve

PR #2699, opened 2026-08-31
Add example for enabling TCP keepalive via socket_options

### LINKED ISSUE #2067: Set tcp keepalive for http connections

thev1ndu:
#### What type of PR is this?

/kind documentation

`Configuration.socket_options` is already passed through to the underlying urllib3 pool, but there is no example showing how to use it. Long-lived requests such as watches or `follow=True` log streams can hang when a connection is silently dropped by a load balancer or firewall. Enabling TCP keepalive lets the kernel detect the dead peer and close the socket. This adds `examples/tcp_keepalive.py` and lists it in the examples README. The helper only sets the platform-specific options (`TCP_KEEPIDLE` and friends) when they are available.

Fixes #2067

Example only, no library code changes.

```release-note
NONE
```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -9,6 +9,7 @@ comments throughout the code.
 
 - pod_logs.py — basic (blocking) pod log streaming example
 - pod_logs_non_blocking.py — non-blocking streaming of pod logs with graceful shutdown
+- tcp_keepalive.py — enable TCP keepalive via Configuration.socket_options so dead connections on long-lived watches/log streams are detected
 
 ## Setup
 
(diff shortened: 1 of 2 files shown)


aojea:
> Long-lived requests such as watches or `follow=True` log streams can hang when a connection is silently dropped by a load balancer or firewall

TCP_KEEPALIVE is not the right solution, it does not work as people thing it does, once the connection is established the keepalives are not sent periodically ... http2 ping frames is the solution to idle connections and middleware silent drops https://github.com/kubernetes/kubernetes/pull/95981

PR #2703, opened 2026-09-01
Support sendInitialEvents in the dynamic client watch

### LINKED ISSUE #2700: Support watch parameter sendInitialEvents
K8s added an additional parameter to watch API calls in 1.32.

```python
sendInitialEvents: bool = Field(default=False)
```

https://kubernetes.io/docs/reference/kubernetes-api/core/event-v1/#query-parameters-6

```
`sendInitialEvents=true` may be set together with `watch=true`. In that case, the watch stream will begin with synthetic events to produce the current state of objects in the collection. Once all such events have been sent, a synthetic "Bookmark" event will be sent. The bookmark will report the ResourceVersion (RV) corresponding to the set of objects, and be marked with `"k8s.io/initial-events-end": "true"` annotation. Afterwards, the watch stream will proceed as usual, sending watch events corresponding to changes (subsequent to the RV) to objects watched. When `sendInitialEvents` option is set, we require `resourceVersionMatch` option to also be set. The semantic of the watch request is as following:
- `resourceVersionMatch` = NotOlderThan is interpreted as "data at least as new as the provided `resourceVersion`" and the bookmark event is send when the state is synced to a `resourceVersion` at least as fresh as the one provided by the ListOptions. If `resourceVersion` is unset, this is interpreted as "consistent read" and the bookmark event is send when the state is synced at least to the moment when request started being processed.
- `resourceVersionMatch` set to any other value or unset Invalid error is returned. Defaults to true if `resourceVersion=""` or `resourceVersion="0"` (for backward compatibility reasons) and to false otherwise.
```

Which allows for better controls of watch resourceVersion return results from the starting list call rather than waiting for an initial Bookmark event.

Adding this parameter to the APIs and DyanmicClient watch function.

chala2001:
#### What type of PR is this?

/kind feature

`DynamicClient.watch()` had no way to set `sendInitialEvents`, so a dynamic watch could not ask the server to replay the current state before streaming changes. This adds `send_initial_events` and `resource_version_match` and forwards them to `Watch.stream()`.

The generated APIs already have both parameters from the OpenAPI spec, so the only gap was the dynamic client.

`request()` builds its query string by hand and had no entry for either name, so the kwargs would have been dropped on the way out. They are added there too. That also makes `resourceVersionMatch` work for plain dynamic `list()` calls, where it was silently ignored before.

Fixes #2700

Two tests in `client_test.py`: one that the parameters reach the watcher, one that they reach the query string. The second is the one that matters, since a passthrough-only test still passes while `request()` drops them.

I kept `send_initial_events` set when `Watch.stream()` re-issues the request after a 410, rather than stripping it, to match how the rest of this layer passes parameters through. There is no client-side check that `resource_version_match` is set, since the API server returns `Invalid` for that itself.

This touches the same `watch()` signature as #2698, so whichever lands second needs a one line rebase.

```release-note
Dynamic client `watch()` now accepts `send_initial_events` and `resource_version_match`.
```

### DIFF
--- a/kubernetes/dynamic/client.py
+++ b/kubernetes/dynamic/client.py
@@ -158,7 +158,7 @@ class DynamicClient:
 
         return self.request('patch', path, body=body, force_conflicts=force_conflicts, **kwargs)
 
-    def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None, watcher=None, allow_watch_bookmarks=None):
+    def watch(self, resource, namespace=None, name=None, label_selector=None, field_selector=None, resource_version=None, timeout=None, watcher=None, allow_watch_bookmarks=None, send_initial_events=None, resource_version_match=None):
         """
         Stream events for a resource from the Kubernetes API
 
@@ -172,6 +172,10 @@ class DynamicClient:
         :param timeout: The amount of time in seconds to wait before terminating the stream
         :param watcher: The Watcher object that will be used to stream the resource
         :param allow_watch_bookmarks: Ask the API server to send BOOKMARK events
+        :param send_initial_events: Ask the API server to begin the stream with synthetic events
+                                    for the current state, followed by a BOOKMARK event
+        :param resource_version_match: How resource_version is matched, e.g. "NotOlderThan".
+                                       Required by the API server when send_initial_events is set
 
         :return: Event object with these keys:
                    'type': The type of event such as "ADDED", "DELETED", etc.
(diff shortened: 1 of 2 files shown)


chala2001:
@yliaog could you approve the workflow run here? It's been sitting at "action_required" since I opened it, so there are no CI results yet.

I rebased onto master yesterday for the kubernetes/base cleanup ,,same change, just moved to kubernetes/dynamic/. Tests pass locally

yliaog:
thanks for the PR

/lgtm
/approve

PR #2707, opened 2026-09-03
fix(leaderelection): survive malformed API error bodies and lock annotations

nkbeast:
**What this fixes**

`try_acquire_or_renew()` parsed the raw API error body with `json.loads` and indexed straight into it:

```python
if json.loads(old_election_record.body)['code'] != HTTPStatus.NOT_FOUND:
```

That assumes whatever came back is a Kubernetes Status object. Anything sitting in front of the API server — an ingress, load balancer or proxy — answers with an HTML error page, an empty payload or some other non-JSON body on a bad day, and `ApiException.body` can also be `None`. Each of those raises out of the election loop (`JSONDecodeError`/`TypeError`), and in the aio elector an empty body dies even earlier on an `assert`.

Leader election dying like this is the one failure mode the code exists to prevent: the controller stops renewing its lease, never calls `onstopped_leading`, and the workload sits in limbo instead of failing over. Reproduced locally with a lock returning a 502 HTML body — `JSONDecodeError` before this change, clean `False` retry after.

**The fix**

- An unparsable or missing error body is treated as "not a 404": the elector logs and retries on the next period, exactly like any other API hiccup. A clean 404 still allows lock creation, unchanged.
- The same crash existed on the read path of the ConfigMap lock: a corrupted leader-election annotation raised out of `get()`. It is now treated like a missing annotation so the next update rewrites a clean record.
- Both the sync and the aio implementations are fixed.

**Tests**

Four regression tests added: HTML error body, `None` body, the clean-404 create path (to prove that behaviour is untouched), and the aio elector. Full suite green locally: `436 passed, 28 skipped` with the same command CI uses.

```release-note
Fixed leader election crashing when the API server or an intermediary returns a malformed or non-JSON error body; the elector now treats it like any other API error and retries on the next period instead of raising. A corrupted leader-election annotation on the ConfigMap lock no longer breaks reads either.
```

### DIFF
--- a/kubernetes/aio/leaderelection/leaderelection.py
+++ b/kubernetes/aio/leaderelection/leaderelection.py
@@ -145,11 +145,22 @@ class LeaderElection:
 
         # A lock is not created with that name, try to create one
         if not lock_status:
-            assert (
+            # The error body comes straight from the API server, but anything
+            # sitting in front of it (ingress, load balancer, proxy) can answer
+            # with an HTML page, an empty payload or some other non-JSON body.
+            # Only a clean 404 means the lock is absent and may be created;
+            # everything else is retried on the next period instead of taking
+            # the whole leader election down.
+            error_code = None
+            if (
                 isinstance(old_election_record, ApiException)
                 and old_election_record.body is not None
-            )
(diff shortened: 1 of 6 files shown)


yliaog:
thanks for the PR

/lgtm
/approve

